PR #7, opened 2021-04-15
Split specification into separate files.

MarkLodato:
This should make the spec easier to find and read. Now the specification lives entirely within the spec/ directory and other supplemental material lives outside.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,15 +281,15 @@ Chapter 14, page 328, "Ensure Unambiguous Provenance." Instead, we recommend
 keying primarily by resource name, in addition to content hash.
 
 [Binary Authorization]: https://cloud.google.com/binary-authorization
-[Envelope]: spec/envelope.md
+[Envelope]: spec/README.md#envelope
 [ITE-6]: https://github.com/in-toto/ITE/pull/15
 [Link]: spec/predicates/link.md
-[Predicate]: spec/predicate.md
+[Predicate]: spec/README.md#predicate
 [Provenance]: spec/predicates/provenance.md
 [SLSA Attestation Model]: https://github.com/slsa-framework/slsa-controls/blob/main/attestations.md
 [SLSA]: https://github.com/slsa-framework/slsa
 [SPDX]: spec/predicates/spdx.md
-[Statement]: spec/statement.md
+[Statement]: spec/README.md#statement
 [in-toto 0.9]: https://github.com/in-toto/docs/blob/v0.9/in-toto-spec.md
 [in-toto]: https://in-toto.io
 [processing model]: spec/processing_model.md
--- a/README.md
+++ b/README.md
@@ -53,169 +53,22 @@ None yet!
 
 ## Specification
 
-An attestation has three layers that are independent but designed to work
-together:
+See [spec/README.md](spec/README.md). Summary:
+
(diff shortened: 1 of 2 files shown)


MarkLodato:
@TomHennen @nenaddedic I'm trying to trim down the README so that it is more of a pointer to the spec rather than being an all-in-one document. Do you think this makes the doc easier to read?

TomHennen (commented):
I sort of feel like envelope and statement go well in the same file, but I don't feel strongly.

MarkLodato:
My motivation for splitting Statement into a separate file was so that you can link to it and have it just describe that, without digging into the other docs. Though if it hurts readability, I'd also be open to putting all three Envelope + Statement + Predicate in one file, such as spec/README.md.

spec/statement.md
```diff
@@ -0,0 +1,49 @@
...
+For example, a [Provenance](predicates/provenance.md) attestation might use the
+name to specify output filename, expecting the consumer to only considers
+entries with a particular name. Alternatively, a vulnerability scan attestation
+might use the name "\_" because the results apply regardless of what the
+artifact is named.
+
+IMPORTANT: Subject artifacts are matched purely by digest, regardless of content
+type. If this matters to you, please open a GitHub issue to discuss.
+
+The `predicateType` and `predicate` together form the [Predicate], describing
```
TomHennen, line 46: [Predicate] link is broken.

spec/envelope.md
```diff
@@ -0,0 +1,14 @@
+# Envelope
```
TomHennen, line 1: Could this also link to statement.md?  that might help people follow along.

TomHennen (commented):
Maybe it's fine to leave them separate as long as people can easily go from one to the other?

So envelope should link to statement, statement to envelope/predicate, predicate to statement?

spec/statement.md
MarkLodato: Oops, fixed.

spec/envelope.md
MarkLodato: this is now obsolete

MarkLodato:
You're right - it is clearer as a single file. Please take another look. Now this PR just splits out the Specification and Processing Model sections to spec/README.md, nothing more. I think the result is ultimately easier to read, but let me know. (I have a follow-on commit to improve the text after the split.)

README.md
```diff
@@ -428,17 +281,15 @@ Chapter 14, page 328, "Ensure Unambiguous Provenance." Instead, we recommend
...
-[Predicate]: #predicate
+[Predicate]: spec/predicate.md
 [Provenance]: spec/predicates/provenance.md
-[RFC 3339]: https://tools.ietf.org/html/rfc3339
-[RFC 3986]: https://tools.ietf.org/html/rfc3986
 [SLSA Attestation Model]: https://github.com/slsa-framework/slsa-controls/blob/main/attestations.md
 [SLSA]: https://github.com/slsa-framework/slsa
-[Statement]: #statement
+[SPDX]: spec/predicates/spdx.md
+[Statement]: spec/statement.md
```
TomHennen, line 292: Should this and 'Predicate' link to spec/README.md?

MarkLodato: Oops, yup! Fixed.

TomHennen (approved):

Issue #4, opened 2021-04-13
Document convention for versioning


MarkLodato:
It would be good to come up with some convention for version numbers, using that for official specs (Statement, Provenance, etc.) and recommending it for other predicates.

### Proposal

Strawman based on [Semantic Versioning 2.0.0](https://semver.org/): "vMAJOR.MINOR". A change MAY increment only the minor version if consumers can parse the version as an earlier minor revision and have it still be considered accurate according to that earlier revision. For example, parsing a 1.1 message as 1.0 results in something that has the same semantic meaning as if the producer produced 1.0 directly. (I don't see a need for PATCH number since this is just a spec, not code.)

Examples for Provenance, going from 1.0 to 1.1 vs 2.0:

- Adding a new `buildFinished` timestamp may increment MINOR because the field does not affect the meaning of any other field. The predicate still makes sense as a 1.0 message ignoring that field.
- Add a new `recipe.extraArgs` field requires incrementing MAJOR because the overall meaning of the `recipe` changes. If you ignore the field, it is NOT semantically a 1.0 message because the `recipe` assumed that the existing 1.0 fields fully defined the recipe.
- Modifying the meaning of `builder.id` requires incrementing MAJOR because it's changing an existing field. Interpreting it as 1.0 is not valid.

### Alternatives

Alternative would be to just have MAJOR versions (v1, v2, v3, ...) and increment it on every single change. The downside is that then any minor increases would require all consumers to update.

Any thoughts?

MarkLodato:
Another idea:

- The schema uses SemVer2, i.e. MAJOR.MINOR.PATCH. (PATCH isn't really needed, but we do it just so we're SemVer compliant.) This way implementations can state specifically which version they support.
- Any message MUST be semantically correct when parsed as any other version with the same major number. That is:
    - Parsing an older message as a newer minor/patch version MUST be semantically equivalent to parsing it as the original version. (All fields present in both versions have the same meaning, and the absence of any new fields has no semantic meaning.)
    - Parsing a newer message as an older minor/patch version MUST be semantically equivalent to parsing it as the newer version with all new fields removed. (All fields present in both version have the same meaning, regardless of the presence or absence of the new fields.)
- The type ID only includes the major version. This allows implementations to support any major version without having to parse the URI to pull out hte version. (Parsing is impossible since we don't mandate the format of the URI.) Implementations just use the URI as an opaque string.

Examples:

[...]

PR #12, opened 2021-04-20
Add `Statement.type` and use MediaType in envelope.

MarkLodato:
## Current description

Add a new Statement `type` field that indicates the schema version,
and use an unversioned MediaType in the Envelope `payloadType` field.

Reasoning:

-   This makes the Statement self-describing, which aids readability.
-   Not all envelopes authenticate the payload type. Since we want the
    layers to be independent, we should include the schema version
    inside to protect against this.
-   MediaType is the standard way of indicating encoding; URI is
    uncommon for that purpose.
-   Most formats include a version inside the payload (e.g. PDF or
    JsonSchema) and have an unversioned MediaType. We do the same.
-   Removing the encoding from the type URI (i.e. removing "-json")
    allows policy engines to be encoding-agnostic.
-   The existing in-toto format (link and layout) already use a `type`
    field, so we can continue using the same concise MediaType.

Note that the same reasoning does *not* apply to `predicateType`, which
continues to be a URI. We considered moving it inside `predicate.type`,
but then we need to worry about field name collisions.

## Original description (obsolete)

Make each layer define its own schema via a standard `type` field,
rather than relying on the outer layer to specify it. This way each
layer is not dependent on the outer layers. Now, for example, the
definition of Predicate is much simpler: it is a JSON object with a
`type` field, not a combination of `predicateType` + `predicate`.

Another minor advantage is that the encoding (e.g. JSON) is independent
of the schema (e.g. v1). This can simplify consumers' logic.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -27,7 +27,7 @@ See the [top-level README](../README.md) for background and examples.
 
 ```jsonc
 {
-  "payloadType": "https://in-toto.io/Statement/v1-json",
+  "payloadType": "application/vnd.in-toto+json",
   "payload": "<Base64(Statement)>",
   "signatures": [{"sig": "<Base64(Signature)>"}]
 }
@@ -39,22 +39,24 @@ adopted by in-toto in [ITE-5]. It is a [JSON] object with the following fields:
 
 `payloadType` *string, required*
 
->   Always `https://in-toto.io/Statement/v1-json` (for the [Statement] defined
->   below).
+> Identifier for the encoding of the payload. Always
+> `application/vnd.in-toto+json`, which indicates that it is a JSON object with
+> a `type` field indicating its schema.
 
 `payload` *string, required*
 
->   Base64-encoded JSON [Statement].
+> Base64-encoded JSON [Statement].
 
 `signatures` *array of objects, required*
 
->   One or more signatures over `payloadType` and `payload`, as defined in
->   [signing-spec].
+> One or more signatures over `payloadType` and `payload`, as defined in
+> [signing-spec].
 
 ## Statement
 
 ```jsonc
 {
+  "type": "https://in-toto.io/Statement/v1",
   "subject": [
     {
       "name": "<NAME>",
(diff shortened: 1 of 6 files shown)


MarkLodato:
@TomHennen @nenaddedic @SantiagoTorres Here is the change we discussed in our meeting. What do you think?

spec/README.md
```diff
@@ -39,30 +39,30 @@ adopted by in-toto in [ITE-5]. It is a [JSON] object with the following fields:
...
->   One or more signatures over `payloadType` and `payload`, as defined in
->   [signing-spec].
+> One or more signatures over `payloadType` and `payload`, as defined in
+> [signing-spec].
 
 ## Statement
 
 ```jsonc
 {
+  "type": "https://in-toto.io/Statement/v1",
```
TomHennen, line 58: optional: Should we use "_type" to help prevent collisions?

TomHennen (changes requested):

spec/README.md
MarkLodato: In-toto currently uses "_type", so that would also be a collision. Other ideas include "$type", "predicateType", "$predicateType", or variants involving "schema" instead of "type". I don't have much of an opinion here. I went with "type" because it as the simplest. 

Any other thoughts?

letmaik: I think it's quite common to define the format/media type in the outer layer if you want to treat the inner layer as an opaque thing. Whether the inner layer itself is self-describing using some standard or custom method (e.g. JSON-LD) is a separate concern. I would also be worried of injecting properties as some formats are strict about it and are not valid anymore (something that's possible to be expressed in JSON schema for validation for example).
If you want to inject/enforce a certain field then I think you'd have to make stronger statements about what types of formats are supported.

MarkLodato: Thanks for the feedback, Maik. A few scattered thoughts:

- It seems like a good idea to add `Statement.type` is in case the envelope doesn't authenticate `payloadType`, since not all envelopes do. This seems like a good idea to me.
- I'm less confident about renaming `payloadType` to `payload.type`. That isn't for security - it's just for comprehension. Note that you can always include an opaque value as `payload.value`, which we do for SPDX. But maybe this isn't a good idea.
- Either way, I'm leaning towards differentiating between "type" and "version". If we think about other formats like JPEG, PDF, or JSON Schema, they all have a single Media Type and have a version within the file. Maybe we should use that same concept here. See also #4.

MarkLodato: Reverted the `predicateType` -> `predicate.type` change, and switched to Media Type for the `payloadType`.

PR #16, opened 2021-04-22
Rename reproducibility to environment

### LINKED ISSUE #14: Provenance: rename `reproducibility`
The name `recipe.reproducibility` sounds like whether it's a reproducible build or not, rather than the set of builder-controlled values that affected the build.

Maybe `recipe.environment`?

MarkLodato:
The original name misleadingly sounded like whether it was reproducible, as opposed to the type.

The word `environment` matches in-toto's current usage and also better conveys that these are builder-controlled parameters.

Fixes #14.

@TomHennen FYI


MarkLodato:
Actually there is more discussion on #14, so I'll withdraw this for now.

PR #18, opened 2021-04-22
Add rekord type from sigstore/rekor

joyliu-q:
Re: in-toto community meeting, where we discussed adding the schema of the rekord type in ITE6 for more general link formats.

Not fully sure if putting it in predicate is the right place, but it looks like the rekord type contains metadata that we want to accept and pass along, and because that falls under the Predicate part of the attestation (from my understanding), I added the schema in that folder. 

Some questions/clarifications:
- Is there already a pre-existing URI for rekord, and where would I find it? I looked at the other links such as SPDX which seem to be not actual links and just something that made up to differentiate the resources. RFC 3986 also states "A Uniform Resource Identifier (URI) is a compact sequence of characters that identifies an abstract or physical resource." At the same time, what are the official naming conventions (aside from the upper/lowercase rules + the v1/v2 if we may want to update) and where are these URIs stored? I tried looking up some of the URIs we already have but couldn't find where or how we came across those.
- When is "v1" for predicate types used and when is it not used (just the predicate type name)?
- Since these are documentation in the spec, where are we enforcing these types (or WIP for now?)

### DIFF
--- a/spec/predicates/rekord.md
+++ b/spec/predicates/rekord.md
@@ -1,0 +1,26 @@
+# Predicate type: Rekord v1
+
+Type URI: https://sigstore.dev/Rekord
+
+## Purpose
+Support for the sigstore's [rekord type](https://github.com/sigstore/rekor). Rekord is a default type for the rekor, which enables software maintainers and build systems to record signed metadata to an immutable record.
+
+## Schema
+
+```jsonc
+{
+  "subject": [{ ... }],
+  "predicateType": "https://sigstore.dev/Rekord", // or https://rekord.sigstore.dev/Rekord - is there already a pre-defined URI?
+  "predicate": {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://rekor.sigstore.dev/types/rekord/rekord_schema.json",
+    "title": "Rekor Schema",
+    "description": "Schema for Rekord objects",
+    "type": "object",
+    "oneOf": [
+        {
+            "$ref": "v0.0.1/rekord_v0_0_1_schema.json"
+        }
+    ]
+  }
+}
--- a/spec/README.md
+++ b/spec/README.md
@@ -15,6 +15,7 @@ independent but designed to work together:
     *   [Provenance]: To describe the origins of a software artifact.
     *   [Link]: For migration from [in-toto 0.9].
     *   [SPDX]: A Software Package Data Exchange document.
+    *   [Rekord]: Recording signed metadata
 
 The [processing model] provides pseudocode showing how these layers fit
 together.
--- a/spec/README.md
+++ b/spec/README.md
@@ -15,7 +15,7 @@ independent but designed to work together:
     *   [Provenance]: To describe the origins of a software artifact.
     *   [Link]: For migration from [in-toto 0.9].
     *   [SPDX]: A Software Package Data Exchange document.
-    *   [Rekord]: Recording signed metadata
+    *   [Rekord](predicate/rekord.md): Recording signed metadata
 
 The [processing model] provides pseudocode showing how these layers fit
 together.
--- a/spec/README.md
+++ b/spec/README.md
@@ -138,6 +138,7 @@ This repo defines the following predicate types:
 *   [Provenance]: To describe the origins of a software artifact.
 *   [Link]: For migration from [in-toto 0.9].
 *   [SPDX]: A Software Package Data Exchange document.
+*   [Rekord](predicate/rekord.md): Recording signed metadata
 
 ### Predicate conventions
 


shibumi (commented):
I think you forgot this abstract here: https://github.com/in-toto/attestation/pull/18/files#diff-4a5c2f67987bb0937e194ebff0e4b7f8b63d6fcc9bd4116a08ecb457d0fc609eR134-R141

EDIT: the link may not work.. it's line 134  to 141 in the spec/Readme.md file

spec/predicates/rekord.md
```diff
@@ -0,0 +1,26 @@
...
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://rekor.sigstore.dev/types/rekord/rekord_schema.json",
+    "title": "Rekor Schema",
+    "description": "Schema for Rekord objects",
+    "type": "object",
+    "oneOf": [
+        {
+            "$ref": "v0.0.1/rekord_v0_0_1_schema.json"
+        }
+    ]
```
MarkLodato, line 24: This is not a JSON Schema but rather a listing of fields. If this is an externally defined schema, you can just say so in text below. Please see the other files in this directory for examples, particularly spdx.md.

spec/predicates/rekord.md
```diff
@@ -0,0 +1,26 @@
+# Predicate type: Rekord v1
+
+Type URI: https://sigstore.dev/Rekord
+
+## Purpose
+Support for the sigstore's [rekord type](https://github.com/sigstore/rekor). Rekord is a default type for the rekor, which enables software maintainers and build systems to record signed metadata to an immutable record.
```
MarkLodato, line 6: Please wrap to 80 columns.

MarkLodato (changes requested):
Sorry, I don't understand. What is a rekord and why is it a predicate? Neither this PR nor the rekor docs seem to explain this at all.

shibumi:
Hi @MarkLodato , thanks for reviewing this PR. Joy and me are both 'new' to the ITE6 concept, so maybe we are wrong here.
The whole purpose of this PR is to investigate if rekord is a potential custom type for the new ITE6 attestation or not.

Rekords are more briefly explained here on the sigstore website: https://sigstore.dev/what_is_sigstore/ (difficult to find.. no idea why they are not directly linking to it in their README).

As far as I understand this, rekords are signing manifests recorded in a transparency log system provided by the Rekord Server. They are additional metadata. The purpose seems to be similar to certificate transparency logs for x509 certificates.

What is the possible relationship to predicates / custom types in the in-toto attestation? According to the attestation README a predicate contains arbitrary metadata about the subject, with a type-specific schema. As example the README mentions a test result:
```
{
  "type": "https://in-toto.io/Statement/v0.1",
  "subject": [
    { "digest": { "sha1": "859b387b985ea0f414e4e8099c9f874acb217b94" } }
  ],
  "predicateType": "https://example.com/TestResult/v1",
  "predicate": {
    "passed": true
  }
}
```

A rekord would now be a transparency log or signing manifest. Imagine a supply chain, where you sign a specific artifact eg. 
a gzipped tarball. Instead of just signing the tarball and publishing the public key on a website, we would instead publish the public key and the signature in the transparency log of the rekord server. While doing so we would get a 'rekord' from the rekord server as anchor to the signing manifest  in the transparency log of the rekord server. 

@MarkLodato  Does this make sense at all or are we completely on a wrong path and misunderstood the whole concept of ITE6?

Nor Joi or me are affiliated to the sigstore project, maybe we should get someone from sigstore to the next in-toto meeting and discuss this. I remember that we talked about a possible integration of rekords in our ITE6 (@SantiagoTorres  mentioned it), but I might be wrong. I had a chill that day so it was a little bit difficult for me as non native english speaker to follow the discussion.

letmaik:
I don't think this PR makes sense. The stack is inverted. An in-toto attestation is signed by someone or some machine, and the signature can then optionally be logged in rekor to provide transparency. The output of rekor is a proof of inclusion, which is different than a document signed with a digital signature. So, rekor is further up the hierarchy and in-toto shouldn't be aware of it.

shibumi:
Sure, but wouldn't it make sense to include the output of rekor as proof of inclusion in the supply chain as message to someone validating the supply chain that the transparency of the signature can be validated via rekor?

in my opinion it is still some sort of metadata that gets produced in the supply chain process. The question is if it should be included or not. It has no effect on the input or outputs of the supply chain step, but it is additional metadata for the generated signature during the supply chain process. I might be wrong, here.

letmaik:
> Sure, but wouldn't it make sense to include the output of rekor as proof of inclusion in the supply chain as message to someone validating the supply chain that the transparency of the signature can be validated via rekor?

Yes, like SCTs in certificate transparency, the rekor proofs should be bundled with the signed in-toto statement (the exact mechanism will vary depending on the distribution mechanism), but they are fundamentally a different thing I think. I'm not 100% sure but if you require that all in-toto attestations are logged in rekor, then you would create an infinite loop if the rekor proofs are themselves in-toto attestations.

shibumi:
> Yes, like SCTs in certificate transparency, the rekor proofs should be bundled with the signed in-toto statement

Ah ok. I think this was my misunderstanding in this discussion. I thought submitting a new record to the Certificate Transparency log (Rekor) would be a dedicated step/link in the in-toto supply, but instead it's preferred to just submit the final signed in-toto statement to Rekor's certificate transparency log. I see.

MarkLodato:
My understanding aligns with Maik's. Rekor sits at the envelope (authn) layer. When deciding whether or not to accept an attestation, the consumer might require a rekor proof. The predicate is an entirely different layer.

MarkLodato:
Shall we close this PR in that case?

shibumi:
@MarkLodato  Yes, makes sense to close this PR. Thanks for your contribution @joyliu-q, I think the discussion was needed :)

joyliu-q:
Sounds good, thank you for clarifying and considering this!

PR #22, opened 2021-04-25
Add a Null predicate type.

### LINKED ISSUE #9: Make predicateType optional
For the case of plain code signing (e.g. via [cosign](https://github.com/sigstore/cosign)), there is no predicate type. (More accurately, there is some implicit predicate implied by the public key.) To support such use cases, we should make `predicateType` optional if `predicate` is not present.

MarkLodato:
This allows attestations to be used as a drop-in replacement for
existing code signing applications, where the predicate is not defined.

Fixes #9.

### DIFF
--- a/spec/predicates/null.md
+++ b/spec/predicates/null.md
@@ -1,6 +1,6 @@
 # Predicate type: Null
 
-Type URI: https://in-toto.io/Null
+Type URI: https://in-toto.io/Null/v1
 
 Version: 1.0.0
 
@@ -18,7 +18,7 @@ traditional code signing, where the notion of a predicate does not exist.
   "subject": [{ ... }],
 
   // Predicate:
-  "predicateType": "https://in-toto.io/Null",
+  "predicateType": "https://in-toto.io/Null/v1",
   "predicate": null  // or unset
 }
 ```
--- a/spec/predicates/null.md
+++ b/spec/predicates/null.md
@@ -1,0 +1,29 @@
+# Predicate type: Null
+
+Type URI: https://in-toto.io/Null
+
+Version: 1.0.0
+
+## Purpose
+
+Denotes the absense of a predicate. May be used as a drop-in replacement for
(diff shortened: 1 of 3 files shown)


spec/predicates/null.md
```diff
@@ -0,0 +1,29 @@
+# Predicate type: Null
+
+Type URI: https://in-toto.io/Null
+
+Version: 1.0.0
+
+## Purpose
+
+Denotes the absense of a predicate. May be used as a drop-in replacement for
```
joshuagl, line 9:
```suggestion
Denotes the absence of a predicate. May be used as a drop-in replacement for
```

spec/predicates/null.md
```diff
@@ -0,0 +1,29 @@
+# Predicate type: Null
+
+Type URI: https://in-toto.io/Null
```
joshuagl, line 3: I assume the idea here is that this Type URI does not need to be versioned because it will only ever be a field-less object?

MarkLodato: Yup, that's the thinking. Should I explain this, or is it sufficient as-is?

spec/predicates/null.md
MarkLodato: Oops, thanks!

spec/predicates/null.md
joshuagl: I think it's probably worth being explicit given that the parsing rules state
> "Each type has a SemVer2 version number and the TypeURI reflects the major version number."

MarkLodato: Sent #33. Think I need something here, too?

MarkLodato: As discussed in #33, I added the version number here.

joshuagl (approved):
Looks good, thanks!

MarkLodato:
Since there hasn't been a use case for this, I'm going to close this for now.

PR #33, opened 2021-05-03
Add footnote that TypeURI may omit version number.

MarkLodato:


MarkLodato:
@joshuagl What do you think?

joshuagl:
It might make implementation simpler if we just require all TypeURI to have a version, even if some types do not expect version to ever change? Same concern applies to the SPDX predicate, too.

MarkLodato:
Yeah, good point. SPDX is special since it has its own embedded version number, so a double would be confusing, but what I'll do is add a special explanation of that, rather than making it some sort of rule here. Thanks for the feedback!

joshuagl:
> SPDX is special since it has its own embedded version number, so a double would be confusing, but what I'll do is add a special explanation of that, rather than making it some sort of rule here.

My reading of the current draft predicate is that the version (of the SPDX spec) could be indicated by the Type URI, or a field in the predicate, but not both?

I'd like to suggest we have versions in the Type URI and not as a field in the predicate, but that does prevent us from changing the predicate unless there's a bump in spec version. I haven't figured out if this is really a problem, though.

PR #34, opened 2021-05-05
Switch Link to v0.1 (was v1).

MarkLodato:
It is not yet clear that the Link schema is finalized. Open questions:
- Should "_type" be optional?
- Should "products" be removed in favor of "subject"?

Since we are not yet decided, the 0.1 version makes this more clear.

NOTE: This goes against semantic versioning, but no one has depended on
this yet, and we do want to use the 0.x series to indicate "in
development", so it seems like a decent enough tradeoff to make at this
point.

### DIFF
--- a/spec/predicates/link.md
+++ b/spec/predicates/link.md
@@ -1,8 +1,8 @@
 # Predicate type: Link
 
-Type URI: https://in-toto.io/Link/v1
+Type URI: https://in-toto.io/Link/v0.1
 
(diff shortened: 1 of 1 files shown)


MarkLodato:
In particular, do you agree that it's OK to go backwards in version since no one is relying on this yet?

SantiagoTorres:
I think it's ok to go backwards :)

PR #35, opened 2021-05-06
Link v0.2: omit `_type` and `products`; add conversion.

MarkLodato:
Omit redundant fields in the predicate, add pseudocode for how to convert to an old-style link, and bump the version number to 0.2.

- _type: Can be inferred from predicateType.
- products: We now define that this is equivalent to subject, rather than saying subject is some ambigous combination of products and/or materials. Although this isn't always a perfect fit, doing so makes it much more clear how to interpret the predicate and fit it into the existing in-toto ecosystem.

@SantiagoTorres What do you think of this conversion rule? I'm updating ITE-6 now and can use this once it's merged.

### DIFF
--- a/spec/predicates/link.md
+++ b/spec/predicates/link.md
@@ -1,14 +1,14 @@
 # Predicate type: Link
 
-Type URI: https://in-toto.io/Link/v0.1
+Type URI: https://in-toto.io/Link/v0.2
 
-Version: 0.1.0
+Version: 0.2.0
 
 ## Purpose
 
-A generic attestation type with the same schema as [in-toto 0.9]. This allows
-existing in-toto users to make minimal changes to upgrade to the new attestation
-format.
+A generic attestation type with the a schema isomorphic to [in-toto 0.9]. This
+allows existing in-toto users to make minimal changes to upgrade to the new
+attestation format.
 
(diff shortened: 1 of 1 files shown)


spec/predicates/link.md
```diff
@@ -1,14 +1,14 @@
...
 
-Version: 0.1.0
+Version: 0.2.0
 
 ## Purpose
 
-A generic attestation type with the same schema as [in-toto 0.9]. This allows
-existing in-toto users to make minimal changes to upgrade to the new attestation
-format.
+A generic attestation type with the a schema isomorphic to [in-toto 0.9]. This
```
joshuagl, line 9:
```suggestion
A generic attestation type with a schema isomorphic to [in-toto 0.9]. This
```

joshuagl (approved):
This is a major improvement. The pseudocode is a nice touch, too.

spec/predicates/link.md
MarkLodato: Thanks for always fixing my typos. :)

joshuagl: Haha, no problem. I'm honestly not reviewing _only_ to fix typos 😬

PR #39, opened 2021-06-14
Provenance: clarify what the builder means.

MarkLodato:
Previously, the examples were unclear as to who generates the provenance.

### DIFF
--- a/spec/predicates/provenance.md
+++ b/spec/predicates/provenance.md
@@ -304,7 +304,8 @@ WARNING: This is only for demonstration purposes. The GitHub Actions team has
 not yet reviewed or approved this design, and it is not yet implemented. Details
 are subject to change!
 
-GitHub-Hosted runner:
+If GitHub is the one to generate provenance, and the runner is GitHub-hosted,
+then the builder would be as follows:
 
 ```json
 "builder": {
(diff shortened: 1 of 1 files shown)


spec/predicates/provenance.md
```diff
@@ -304,7 +304,8 @@ WARNING: This is only for demonstration purposes. The GitHub Actions team has
 not yet reviewed or approved this design, and it is not yet implemented. Details
 are subject to change!
 
-GitHub-Hosted runner:
+If GitHub is the one to generate provenance, and the runner is GitHub-hosted,
```
msuozzo, line 307: I think this just brings up a detail that wasn't obvious to me: "Builder" primarily refers to the (id)entity generating the provenance, not the "runner" that actually did the build. This feels like it could easily be broken into two pieces to make things more clear (esp. at the lower SLSA levels): 1) authority that attests to the result of the build / content of provenance 2) runner that executes the build steps.

But given that that'd be a bigger change, I do think this is on the right path towards improving the examples.

msuozzo (approved):

spec/predicates/provenance.md
MarkLodato: OK, I'll file an issue for that. It's a good point.

Issue #38, opened 2021-06-04
Define a "attestation bundle" data structure and naming convention


MarkLodato:
_Copied from https://github.com/secure-systems-lab/signing-spec/issues/20._

We need a **data structure** and **file naming convention** to associate multiple attestations to a single software artifact.

### Motivating use case

Suppose file `foo.out` is associated with two attestations, both of which are required by the layout of `foo.out`:

* Provenance: the build system generates a link saying that `foo.out` was produced from material `foo.c`.
* Vulnerability scan: a scanner generates a link saying that `foo.out` is free of known vulnerabilities.

The build system and the scanner need to know where to place the links on the filesystem, and the verifier needs to know how to find those links when evaluating `foo.out`.

### Data structure

Define a "bundle" data structure containing multiple attestations (i.e. multiple [DSSEs](https://github.com/secure-systems-lab/dsse)).

Ideas:

* [JSON Lines](https://jsonlines.org/)
* ZIP file with particular naming convention
* JSON object or array

### File naming convention

Define a convention for locating the attestation bundle related to a given file on disk.

Proposal: `<file>.<bundle_extension>`

Example: If we choose JSON Lines, perhaps the attestations for `foo.out` could be found in `foo.out.intoto.jsonl`?

TomHennen:
I'd like to propose we go with JSON Lines.

1. I think it's pretty easy.
2. It doesn't require any extra parsing of all the attestations (as a JSON object/array would)
3. It migrates easily to other metadata stores that can store text (where zip files might be tricky).

Ideally I think the extension should reference dsse, so `foo.out.dsse.jsonl`?  Otherwise `foo.out.intoto.jsonl` would be fine.

FWIW getting this defined might help with [how to handle cert chains](https://github.com/secure-systems-lab/dsse/issues/39) and help us clarify that really we expect there to be multiple attestations for different purposes and you can have specialized ones along with 'standardized' ones.

MarkLodato:
I'm OK with this, but it would be good to get input from others.

trishankatdatadog:
Are there good reasons for JSON Lines, and not just good old-fashioned JSON?

TomHennen:
[This site](https://hackernoon.com/json-lines-format-76353b4e588d) gives some good reasoning.

tl;dr: With regular JSON you'd have to read the entire file to parse _or add_ a single record.

trishankatdatadog:
[...]

PR #45, opened 2021-07-15
Proposal for 'Attestation Bundles'

TomHennen:
Propose using JSON Lines since it makes it easy parse and add individual records.

[This post](https://hackernoon.com/json-lines-format-76353b4e588d) explains some of the advantages well.


TomHennen:
Addresses #38

TomHennen:
Instead of saying each line _MUST_ contain a DSSE should we instead just say that consumers should ignore _any_ line they don't understand?  That would allow for other JSON objects to be embedded as well (not sure if that would be good or bad).

bundles.md
```diff
@@ -0,0 +1,131 @@
+# Attestation Bundle
```
MarkLodato, line 1: nit: add blank line after headings (here and below)

MarkLodato, line 1: Link to this page from README.md. Also, should this go within `spec/`?

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+might be able to _delete_ an attestation without being detected.  Predicates that
+follow [the monotonic principle](spec/README.md#parsing-rules) should not have any
+issues with this behavior.
+
+## Data structure
+
+Attestation Bundles use [JSON Lines](https://jsonlines.org/) to store multiple
+[DSSEs](https://github.com/secure-systems-lab/dsse).
+
+*  Each line MUST contain a _single_ DSSE.
```
MarkLodato, line 25: Is it OK for a bundle to contain old-style in-toto links and layouts? Since consumers ignore any unrecognized payload types, why not allow other JSON objects as well?

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+*  Each line MAY contain any DSSE `payloadType`
+*  in-toto attestations (`payloadType` == `application/vnd.in-toto+json`) MAY use
+   any ` _type`/`predicateType`
+*  [in-toto Statements](https://github.com/in-toto/attestation/blob/main/spec/README.md#statement)
+   in a Bundle MAY reference different Subjects
+*  Consumers MUST ignore any attestations whose `payloadType`, `_type_`, `predicateType`
```
MarkLodato, line 31: `_type` (no trailing underscore) and add "or" before `predicateType`.

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+## Data structure
+
+Attestation Bundles use [JSON Lines](https://jsonlines.org/) to store multiple
+[DSSEs](https://github.com/secure-systems-lab/dsse).
+
+*  Each line MUST contain a _single_ DSSE.
+*  Each line MAY contain any DSSE `payloadType`
+*  in-toto attestations (`payloadType` == `application/vnd.in-toto+json`) MAY use
+   any ` _type`/`predicateType`
+*  [in-toto Statements](https://github.com/in-toto/attestation/blob/main/spec/README.md#statement)
```
MarkLodato, line 29: nit: make a relative link

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+   in a Bundle MAY reference different Subjects
+*  Consumers MUST ignore any attestations whose `payloadType`, `_type_`, `predicateType`
+   they do not understand.
+
+Example:
+```
```
MarkLodato, line 35: nit: add blank line before and add a language (```jsonl)

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+{ "payloadType": "application/vnd.foo+xml", "payload": "...", "signatures": [...] }
+```
+
+## File naming convention
+
+* Attestation Bundles SHOULD use the `.intoto.jsonl` extension.
+* Bundles that concern a single artifact SHOULD name the bundle file
+  `<artifact filename>.intoto.jsonl`.
+* Bundles that concern multiple artifacts SHOULD name the bundle file
+  `build.intoto.jsonl`.
```
MarkLodato, line 47: nit: Why "build"? Should this maybe just be `intoto.jsonl` in case it's about non-build properties?

TomHennen:
Adding some more folks to get their feedback per @MarkLodato's suggestion.

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+### Testing
+
+The CI/CD system then requests that automated tests be run on `fooly.apk`.  The test
+system produces an attestation as an in-toto Statement with
+`predicateType=https://in-toto.io/TestResult/v1` which indicates `fooly.apk` with hash
+`aaa...` passed 23 tests.
+
+The TestResult is then _appended_ to the contents of `fooly.apk.intoto.jsonl`
```
MarkLodato, line 75: nit: period at end of sentence

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+
+## File naming convention
+
+* Attestation Bundles SHOULD use the `.intoto.jsonl` extension.
+* Bundles that concern a single artifact SHOULD name the bundle file
+  `<artifact filename>.intoto.jsonl`.
+* Bundles that concern multiple artifacts SHOULD name the bundle file
+  `build.intoto.jsonl`.
+
+## Example Use Case
```
MarkLodato, line 49: nit: I think this section is valuable, but it is pretty long relative to the rest of the doc. Should we make it significantly shorter?

bundles.md
```diff
@@ -0,0 +1,131 @@
...
+attestations would simply be added to the existing bundle.
+
+**NOTE**: Attestation Bundles themselves are **not authenticated** instead each
+individual attestation is authenticated
+([using DSSE](https://github.com/secure-systems-lab/dsse)). As such, an attacker
+might be able to _delete_ an attestation without being detected.  Predicates that
+follow [the monotonic principle](spec/README.md#parsing-rules) should not have any
+issues with this behavior.
+
+## Data structure
```
MarkLodato, line 20: Should there be something about the order of lines (or lack thereof)?

bundles.md
```diff
@@ -0,0 +1,131 @@
+# Attestation Bundle
+in-toto attestations support multiple different predicate types with the idea that
+users will want to propagage multiple assertions about a single software package
+(Provenance, Code Review, Test Result, vuln scan, ...).
+
+Attestation Bundles provides a canonical method of storing those attestations in
+a single place.
+
+The attestations stored in a bundle may all come from the same source, signed by
```
MarkLodato, line 9: Should this move to the "data structure" section where you say what MAY happen?

bundles.md
```diff
@@ -0,0 +1,131 @@
+# Attestation Bundle
+in-toto attestations support multiple different predicate types with the idea that
```
MarkLodato, line 2: I find this intro a bit confusing since you mention predicates, then assertions, then attestations. Maybe just start with "An attestation bundle is a collection of multiple [attestations](...) in a single file." and then explain why it's useful?

MarkLodato (changes requested):

bundles.md
TomHennen: You're suggesting it be moved to the spec folder?

bundles.md
TomHennen: Done WDYT?

bundles.md
TomHennen: Sure, I also trimmed it down some.

bundles.md
TomHennen: "Processing of a bundle MUST NOT depend on the order of the attestations."

bundles.md
TomHennen: I like this idea from a usability perspective, but I'm concerned from a security one.  Since JSON objects don't have standardized typing someone might mis-interpret one object as another.

bundles.md
TomHennen: I think the 'smart' editor did that to me.  Good catch.

bundles.md
TomHennen: Good idea.

bundles.md
TomHennen: Any thoughts on how?  I could just cut some sections.

bundles.md
```diff
@@ -0,0 +1,133 @@
...
+
+Attestation Bundles use [JSON Lines](https://jsonlines.org/) to store multiple
+[DSSEs](https://github.com/secure-systems-lab/dsse).
+
+*  Each line MUST contain a _single_ DSSE.
+*  Each line MAY contain any DSSE `payloadType`
+*  in-toto attestations (`payloadType` == `application/vnd.in-toto+json`) MAY use
+   any ` _type`/`predicateType`
+*  [in-toto Statements](spec/README.md#statement)
+   in a Bundle MAY reference different Subjects
```
trishankatdatadog, line 25: What are Subjects again?

bundles.md
```diff
@@ -0,0 +1,133 @@
...
+{ "payloadType": "application/vnd.foo+xml", "payload": "...", "signatures": [...] }
+```
+
+## File naming convention
+
+* Attestation Bundles SHOULD use the `.intoto.jsonl` extension.
+* Bundles that concern a single artifact SHOULD name the bundle file
+  `<artifact filename>.intoto.jsonl`.
+* Bundles that concern multiple artifacts SHOULD name the bundle file
+  `intoto.jsonl`.
```
trishankatdatadog, line 47: Why not just simplify and use one set of rules? Particular, the multiple artifacts case can cover single artifacts.

bundles.md
trishankatdatadog: It's also kinda repetitive. Maybe we can take out the redundant example above.

trishankatdatadog: Some examples here can definitely be cut, they look pretty much the same at first glance. We should insert different examples covering the different combinations of SHOULDs and MAYs in the data structure above.

trishankatdatadog (changes requested):

bundles.md
joshuagl: A subject is the "Set of software artifacts that the attestation applies to. Each element represents a single software artifact." [from here](https://github.com/in-toto/attestation/blob/main/spec/README.md#statement)

bundles.md
TomHennen: We've seen cases where some builders emit multiple artifacts but want to emit separate attestations for each artifact.

E.g. They emit foo.out, foo.out.intoto.jsonl, and bar.out, bar.out.intoto.jsonl.

bundles.md
TomHennen: I made these changes:

* Removed 'Testing' section since it was redundant with others.
* Updated the vuln scanning section to use a different payloadType to demonstrate that not everything needs to be an in-toto statement.
* Updated the deployment section to indicate that the policy engine skips attestations it doesn't understand (in this case SPDX, but it could easily be anything else).
* Removed example from the data structure section.

bundles.md
adityasaky: The file names seem a bit awkward to me as well, specifically with the multiple artifacts rule. Also, does the last rule contradict the first one?

TomHennen: Can you provide anything you're happier with?  FWIW .jsonl is the suggest JSON Lines suffix.

I can remove the first bullet.  I think it's redundant.

adityasaky: > I can remove the first bullet. I think it's redundant.

I agree.

>  FWIW .jsonl is the suggest JSON Lines suffix.

I'm okay with the .jsonl extension. I wonder if we can use `<artifact name>.intoto.jsonl` and something like `multiple.intoto.jsonl` to remain consistent? I think that also handles the first bullet. Further, I'm guessing we're not considering `<user defined name>.intoto.jsonl` to allow for finding bundles in tooling? That is, "look for <artifact>.intoto.jsonl or just intoto.jsonl"-esque behaviour?

Apologies if these are silly questions! :)

bundles.md
MarkLodato: LGTM

MarkLodato (changes requested):

bundles.md
TomHennen: Done

bundles.md
```diff
@@ -0,0 +1,109 @@
+# Attestation Bundle
+
+An attestation bundle is a collection of multiple attestations in a single file.
+This allows attestations from multiple different points in the software supply
+chain (e.g. Provenance, Code Review, Test Result, vuln scan, ...) to be grouped
+together, allowing users to make decisions based on all available information.
+
+**NOTE**: Attestation Bundles themselves are **not authenticated** instead each
```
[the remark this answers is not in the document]
TomHennen: Done

bundles.md
```diff
@@ -0,0 +1,109 @@
...
+individual attestation is authenticated
+([using DSSE](https://github.com/secure-systems-lab/dsse)). As such, an attacker
+might be able to _delete_ an attestation without being detected.  Predicates that
+follow [the monotonic principle](spec/README.md#parsing-rules) should not have any
+issues with this behavior.
+
+## Data structure
+
+Attestation Bundles use [JSON Lines](https://jsonlines.org/) to store multiple
+[DSSEs](https://github.com/secure-systems-lab/dsse).
```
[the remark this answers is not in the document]
TomHennen: Done

bundles.md
TomHennen: Ah yes, that makes sense.

I think this comment is obsolete since accepting your other suggestion.

bundles.md
TomHennen: Removed first bullet.

I do like this idea of just calling it 'multiple.intoto.jsonl', for one thing this easily lets people grab all the extensions by just searching for .endswith('.intoto.jsonl') and not having to worry about 'notintoto.jsonl'.

Changed to 'multiple.intoto.jsonl'.  

> Further, I'm guessing we're not considering <user defined name>.intoto.jsonl to allow for finding bundles in tooling?

Right!

Going to resolve this comment, please reopen if there are still issues.

README.md
```diff
@@ -67,6 +67,7 @@ See [spec/README.md](spec/README.md). Summary:
     *   [Provenance]: To describe the origins of a software artifact.
     *   [Link]: For migration from [in-toto 0.9].
     *   [SPDX]: A Software Package Data Exchange document.
+*   [Bundles]: Defines a method of grouping multiple attestations together.
```
MarkLodato, line 70: nit: use singular for consistency with the other entries in the list. Same in other README and for the filename (bundle.md).

MarkLodato (approved):

README.md
TomHennen: Done.

bundles.md
adityasaky: No issues, thanks!

trishankatdatadog:
Great job, everyone, thanks!

Going fwd, may I recommend having two reviewers, each from separate orgs?

TomHennen:
Oh yes, absolutely!

I'm assuming you mean two people from different orgs actually clicking 'Approve'.

It would be great if GitHub had a way to require this.

For this PR let me know if there's _anything_ you'd like me to change.

trishankatdatadog:
> It would be great if GitHub had a way to require this.

None, I'm afraid, but we could increase threshold of reviews to 2 🙂 

> For this PR let me know if there's _anything_ you'd like me to change.

I don't think so, I think it's in very good shape now. Thanks!

adityasaky:
image

Updated the repo settings. :)

PR #50, opened 2021-07-21
Where is the attestation provided?

SurferJeffAtGoogle:
Question.  Do Not Merge.


MarkLodato:
The artifact is published out of band in an as-yet-unspecified manner. If you'd like to ask further questions, please open a GitHub issue rather than a pull request. Thanks!

PR #53, opened 2021-07-23
README: Replace "master" branch names with "main" in examples

axelsimon:
"main" is now the default branch name, and it's also a bit friendlier than "master".

Leaving "master" in the curl example, as the branch hasn't currently been renamed in the curl/curl-docker repo and thus "master" is correct here, from a URL perspective.


MarkLodato (approved):
Thanks!

adityasaky (approved):

PR #57, opened 2021-08-03
Remove Provenance predicate and ci_survey.

### LINKED ISSUE #54: Move Provenance to SLSA repo
The current Provenance predicate is described as a generic way to express provenance, but it was designed expressly for [SLSA](https://slsa.dev/). It makes certain assumptions and trade-offs, such as carefully designing the fields to avoid mistakes when applying a SLSA policy. Other use cases of "provenance" may make different trade-offs, such as including the list of build steps that were performed to allow policies to detect `curl | bash`, which for SLSA is unnecessary and may lead to confusion.

To avoid these issues, it might be best to move all of the predicates out of this repo and instead maintain an index of links.
* Provenance -> SLSA
* Link -> in-toto (a different repo)
* SPDX -> something maintained by SPDX team

That would make it more clear that (a) other definitions of "provenance" are OK for different use cases, and (b) not all predicates need to be defined in this repo.

Any thoughts? cc @adityasaky @TomHennen @dlorenc

MarkLodato:
Fixes #54.


README.md
```diff
@@ -45,7 +44,7 @@ Goals:
     anyone, such as [in-toto] or [Binary Authorization].
 -   Make it possible to write automated policies that take advantage of
     structured information.
--   Fit within the [SLSA Framework][SLSA]. The [provenance] format defined here
+-   Fit within the [SLSA Framework][SLSA].
     is the official SLSA recommendation.
```
TomHennen, line 48: I think something got mangled here.  Now it just reads "Fit within the [SLSA Framework][SLSA].
    is the official SLSA recommendation."

TomHennen (changes requested):

README.md
MarkLodato: Oops, fixed, thanks for catching.

TomHennen (approved):

adityasaky (approved):

Issue #92, opened 2022-04-13
Proposal: support subjects that have no digests


wietse-gmail:
Currently, an in-toto [Attestation](https://github.com/in-toto/attestation/blob/main/spec/README.md#statement) subject can have  the form `{name, digest}` and where both fields must be present. A name field can have the value "`_`" when a subject has no meaningful name, but it is currently impossible to specify a subject that has no meaningful digest.

We would like to support subjects of the form `{name, uri}` for subjects that do not have a meaningful digest. Examples:

-   A subject URI that identifies a builder (example: `https://build.example.com/worker@v0.2`). This could be the subject of an attestation that the builder meets a certain SLSA level, and it could be referenced in the builder section of a provenance attestation.
-   A subject URI that identifies a specific revision of a source-code repository (for example, `svn+ssh://<host>/<repo-name>/<revision-number>`). This subject could be referenced in the materials section of a provenance attestation, and it could match the subject of an attestation that the repository meets a certain SLSA level.

### Why not use a content digest?

-   Computing a digest may not be feasible. For example, computing a digest over a (large) source repository at a specific revision number; or over the components that make up a build system stack, from the applications that manage the execution of compilers and other tools, to the operating systems and virtual machines that the software runs on.
-   A digest may not have useful semantics. For example, when a build system's content digest were hypothetically computed over only the builder-specific components of the build system stack, even a trivial software update would change that digest, without affecting any of the build system's SLSA security properties, i.e. things that we actually care about.

### Proposal

-   In a [Statement](https://github.com/in-toto/attestation/blob/main/spec/README.md#statement) subject:
    -   A subject can have the new form `{name, uri}`, as an alternative to the existing form `{name, digest}`.
    -   The subject `uri` field uses [resource URI syntax](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#ResourceURI). See "What makes a good subject URI?" below for desirable properties.
    -   When searching for an attestation, require an exact match with a subject URI.
    -   No change is needed in the subject `name` field spec. This field may contain "`_`", or additional information to evaluate the attestation (for example to select between "production" vs "testing").
-   No change is needed in the [Provenance](https://slsa.dev/provenance/) predicate spec. The fields of interest,  `builder` and `materials`, already support an URI without a digest.

### What makes a good subject URI?

[...]

Issue #5, opened 2021-04-13
Support subjects that are not digests


MarkLodato:
We have use cases where we need to refer to subjects that cannot be describe by a cryptographic hash. Example: Subversion repository revision, which is only identified by URI. We'd only want to support things that are semantically immutable.

MarkLodato:
Related: #27 - the subject (or materials) might be another attestation

MarkLodato:
Another use case: attestation on a builder, where the subject is the builder ID. An auditor can, e.g. describe [SLSA requirements](https://slsa.dev/requirements) that are met for all builders having that builder ID.

This information should [not](https://github.com/slsa-framework/slsa/issues/315) go in the provenance itself because it is metadata about how the provenance was generated. No one should ever trust that field. Either the consumer believes it to be true already (so they don't need the field) or they don't (so they ignore the field). But having an attestation on the builder would be great.

One potential concern is related to #81: how do we know that the meaning of the builder ID doesn't drift over time? My intuition is that we can't. We just have to trust it. For example, if an auditor reviews a build system and certifies that it generates provenance non-falsifiably, there may be a future change that breaks that guarantee. But I think we have to accept that and maybe re-certify over time. It's not really feasible to hash the entire trusted computing base on which the attestation rests, and even if it were, it's likely not meaningful. An auditor wouldn't review every single change to every single dependency, thus we have to just trust that implementations don't drift substantially over time.

MarkLodato:
In terms of implementing this, we could add support for a `uri` subject, which would support most use cases.

A tricky question is what the semantics are when both `uri` and `digest` are set. Must the consumer verify both? What if they only have one to verify against? A simple workaround is to disallow that case, at least for now. You can set either field but not both, which is unambiguous. Once we have some concrete use cases for having both set, then we can figure out what the semantics should be.

wietse-gmail:
I'm proposing a solution with [Proposal: support subjects that have no digests #92]
(https://github.com/in-toto/attestation/issues/92). The basic idea is to allow a subject.uri (a resourceURI) instead of a subject.digest.

marcelamelara:
[...]

PR #95, opened 2022-04-29
Support attestation subject URI as alternative for digest

wietse-gmail:
Hello, I am a member of the Google team, working with Mark Lodato, Tom Hennen and others.

Currently, an in-toto Attestation Statement subject has the form `{name, digest}` where both fields must be present. It is currently impossible to specify a subject that has no meaningful digest.

We would like to support subjects of the form `{name, uri}` for subjects that have no meaningful digest. Examples:

*    A subject URI that identifies a builder (example: https://host/builder@version). This could be the subject of an attestation that the builder meets a certain SLSA level, and it could be referenced in the builder section of a provenance attestation.
*    A subject URI that identifies a specific revision of a source-code repository (for example, svn+ssh://host/path/revision-number). This subject could be referenced in the materials section of a provenance attestation, and it could match the subject of an attestation that the repository meets a certain SLSA level.

More background is in #92.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -148,8 +148,8 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 > Identifies an artifact by an immutable resource URI.
 > The `uri` field MUST be specified only if a subject cannot have a `digest` field.
 > 
-> Examples of subjects that cannot be identified by a digest, but that can be
-> identiified by an URI are a source-code revision (URI = svn+ssh://host/path/revision-number)
+> Examples of subjects that cannot be identified by a `digest`, but that can be
+> identiified by an `uri` are a source-code revision (URI = svn+ssh://host/path/revision-number)
 > or a builder API (URI = https://host/builder@version). 
 >
 > The semantics of an `uri` value MUST be _immutable_; for example each source-code
--- a/spec/README.md
+++ b/spec/README.md
@@ -148,8 +148,9 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 > Identifies an artifact by an immutable resource URI.
 > The `uri` field MUST be specified only if a subject cannot have a `digest` field.
 > 
-> Examples of a subject `uri` are a source-code revision (svn+ssh://host/path/revision-number)
-> or a builder API (https://host/builder@version). 
+> Examples of subjects that cannot be identified by a digest, but that can be
+> identiified by an URI are a source-code revision (URI = svn+ssh://host/path/revision-number)
+> or a builder API (URI = https://host/builder@version). 
 >
 > The semantics of an `uri` value MUST be _immutable_; for example each source-code
 > commit has a distinct revision number, and each semantic builder change has a
--- a/spec/README.md
+++ b/spec/README.md
@@ -122,8 +122,8 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 `subject[*].digest` _object ([DigestSet]), optional_
 
 > Collection of cryptographic digests for the contents of this artifact.
-> This is the preferred form: only if a subject cannot have a `digest field,
-> the `uri` field MUST be specified.
+> This is the preferred form: the `uri` field MUST be specified only if a
+> subject cannot have a `digest field.
 >
 > Two DigestSets are considered matching if ANY of the fields match. The
 > producer and consumer must agree on acceptable algorithms. If there are no
@@ -146,7 +146,7 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 `subject[*].uri` _string ([ResourceURI]), optional_
 
 > Identifies an artifact by an immutable resource URI.
-> Use this form only if a subject cannot have a `digest` field.
+> The `uri` field MUST be specified only if a subject cannot have a `digest` field.
 >
 > The `uri` field SHOULD be specified for subjects that are not identified by 
 > a digest, such as a source-code revision (svn+ssh://host/path/revision-number)
--- a/spec/README.md
+++ b/spec/README.md
@@ -147,9 +147,8 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 
 > Identifies an artifact by an immutable resource URI.
 > The `uri` field MUST be specified only if a subject cannot have a `digest` field.
->
-> The `uri` field SHOULD be specified for subjects that are not identified by 
-> a digest, such as a source-code revision (svn+ssh://host/path/revision-number)
+> 
+> Examples of a subject `uri` are a source-code revision (svn+ssh://host/path/revision-number)
 > or a builder API (https://host/builder@version). 
 >
 > The semantics of an `uri` value MUST be _immutable_; for example each source-code
--- a/spec/README.md
+++ b/spec/README.md
@@ -270,7 +270,7 @@ Output (to be fed into policy engine):
 [JSON]: https://www.json.org
 [Link]: predicates/link.md
 [Predicate]: #predicate
-[ResourceURI]: field_types.md#TResourceURI
+[ResourceURI]: field_types.md#ResourceURI
 [RFC 3339]: https://tools.ietf.org/html/rfc3339
 [RFC 3986]: https://tools.ietf.org/html/rfc3986
 [SLSA Attestation Model]: https://github.com/slsa-framework/slsa/blob/main/controls/attestations.md
--- a/spec/README.md
+++ b/spec/README.md
@@ -123,7 +123,7 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 
 > Collection of cryptographic digests for the contents of this artifact.
 > This is the preferred form: the `uri` field MUST be specified only if a
-> subject cannot have a `digest field.
+> subject cannot have a `digest` field.
 >
 > Two DigestSets are considered matching if ANY of the fields match. The
 > producer and consumer must agree on acceptable algorithms. If there are no
--- a/spec/README.md
+++ b/spec/README.md
@@ -122,7 +122,8 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 `subject[*].digest` _object ([DigestSet]), optional_
 
 > Collection of cryptographic digests for the contents of this artifact.
-> A `subject` MUST have either a `digest` field or an `uri` field.
+> This is the preferred form: only if a subject cannot have a `digest field,
+> the `uri` field MUST be specified.
 >
 > Two DigestSets are considered matching if ANY of the fields match. The
 > producer and consumer must agree on acceptable algorithms. If there are no
@@ -145,7 +146,7 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 `subject[*].uri` _string ([ResourceURI]), optional_
 
 > Identifies an artifact by an immutable resource URI.
-> A `subject` MUST have either a `digest` field or an `uri` field.
+> Use this form only if a subject cannot have a `digest` field.
 >
 > The `uri` field SHOULD be specified for subjects that are not identified by 
 > a digest, such as a source-code revision (svn+ssh://host/path/revision-number)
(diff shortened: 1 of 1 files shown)


spec/README.md
```diff
@@ -119,9 +119,10 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 > content type. If this matters to you, please comment on
 > [GitHub Issue #28](https://github.com/in-toto/attestation/issues/28)
 
-`subject[*].digest` _object ([DigestSet]), required_
+`subject[*].digest` _object ([DigestSet]), optional_
 
 > Collection of cryptographic digests for the contents of this artifact.
+> A `subject` MUST have either a `digest` field or an `uri` field.
```
MarkLodato, line 125: I think we should more strongly state that `digest` is generally required if possible, and to only use `uri` if digest is not possible. Also that `digest` and `uri` are mutually exclusive.

(same on line 148)

spec/README.md
```diff
@@ -256,6 +270,7 @@ Output (to be fed into policy engine):
 [JSON]: https://www.json.org
 [Link]: predicates/link.md
 [Predicate]: #predicate
+[ResourceURI]: field_types.md#TResourceURI
```
MarkLodato, line 273: typo: extra "T"

MarkLodato (commented):
I'd like to hear others' opinions on this, particularly @SantiagoTorres.

spec/README.md
wietse-gmail: I deleted the spurious T

spec/README.md
wietse-gmail: Updated the preferred form and exclusivity of digest versus uri.

spec/README.md
```diff
@@ -141,6 +143,19 @@ subject and unambiguously identifying the types of the [predicate]. It is a
 >
 > MUST be non-empty and unique within `subject`.
 
+`subject[*].uri` _string ([ResourceURI]), optional_
+
+> Identifies an artifact by an immutable resource URI.
+> The `uri` field MUST be specified only if a subject cannot have a `digest` field.
+> 
+> Examples of subjects that cannot be identified by a `digest`, but that can be
+> identiified by an `uri` are a source-code revision (URI = svn+ssh://host/path/revision-number)
```
mlieberman85, line 152: nit: type should be "identified"

mlieberman85:
I'm curious a bit more about the reason for `uri` to be exclusive? Some folks have begun to express some desire to potentially have and optional `uri` in the subject itself. This would allow reverse lookups of attestation to what that attestation's subject is. There might be a better way, but it's a bit unclear.

spec/README.md
wietse-gmail: Should I replace the two instances of "cannot have a digest" with "cannot be identified by a digest"? The word"type" does not appear in the diff.

wietse-gmail:
The subject `name` field has no defined semantics, and repurposing that field could invite confusion. To avoid confusion we propose a new field for non-digest identifiers that can be specified instead of the (currently required) `digest` field. 

Would a different field name than `uri` be preferable? 

Or a different structure like `"non-digest": {"uri": "svn+ssh"//host/repo/revision"}` as an alternative to `"digest": {...}`?

SantiagoTorres:
Hi all, thank you for the feedback so far! I just wanted to tie in the [ITE-4 bits](https://github.com/in-toto/ITE/blob/master/ITE/4/README.adoc).

I could envision that a non-hash then could look like:

```json
"uri": { "uuid": "THEID"}
```
rather than a hash identifier. I think it could also allow for integrity semantics inside of the "hash object" field

marcelamelara:
I have an alternate suggestion. Both the `digest` and `uri` are ultimate used to identify a subject, so what if we moved both the `digest` and `uri` as fields under a single `id` field within `subject`? Then, I think it might make sense to make the `uri` field required and the `digest` field optional. This way, as @mlieberman85 pointed out, the subject can be looked up, and if a digest is available for it, it can provide that additional integrity semantics, as @SantiagoTorres mentioned. I also think that this scheme would allow for extensibility for other types of subject identifiers in the future.

```
"subject": {
      "name": "",
      "id": {
               "uri": "<URI>", // required
               "digest": {"<ALG>": "<VAL>" } //optional
       }
}
```

spec/README.md
wietse-gmail: I have replaced "identiified" with "identified".

MarkLodato:
My main concern is that the semantics are crystal clear. Right now, you match purely on `digest`, and then optionally use `name` to filter out specific artifacts, though even the latter is arguably too ambiguous. If there is both a `uri` and `digest`, how does matching work? Do both have to match? Either? Always `digest` and optionally `uri`?

I'd like to have a set of concrete use cases for these. Wietse has some for pure `uri` cases, where there is no `digest` and thus it's unambiguous. What are the use cases for having both `uri` and `digest`?

spec/README.md
lumjjb: @SantiagoTorres , @mlieberman85 and I were talking about this in a slightly different context of using the attestations to be able to follow them or include locator information in them. 

For our case, i think we are thinking that `uri` and `digest` aren't mutually exclusive. For example, we want to include the digest and have the URI point to a blob store, such as `s3://..../app.spdx`, or `oci://...` or `gcs://...`. This would also potentially create a cleaner separation between rekor and its reliance on a blob store and indexing.

lumjjb:
To @mlieberman85 's point, I think we were thinking about things in a bit of a different way of using URI - and maybe that's going to be a separate field. 

We are thinking that `uri` and `digest` aren't mutually exclusive. For example, we want to include the digest and have the URI point to a blob store, such as `s3://..../app.spdx`, or `oci://...` or `gcs://...`. This would also potentially create a cleaner separation between rekor and its reliance on a blob store and indexing.

wietse-gmail:
Thank you for all the feedback. I'l try to summarize the discussion below.

The status quo:
```
"subject": [
  {
    "name": "<NAME>",                         // required, application-defined
    "digest": { "<ALGORITHM>": "<HEX_VALUE>"  }  // required
  }
]
```
The original proposal:
```
"subject": [
  {
    "name": "<NAME>",                         // no change
    "digest": { "<ALGORITHM>": "<HEX_VALUE>"  } // optional 
    "uri": "<SCHEME://HIERARCHY>"             // optional, proposed
  }
]
```
with the understanding that `digest` or `uri` MUST be specified.

@mlieberman85 suggested that some use cases could benefit from having both.
Would allowing "at last one of `digest` or `uri`" address that?

This generalizes further, for example @SantiagoTorres suggested that
UUIDs could be used "rather than" (or in addition to) a digest.
```
"subject": [
  {
    "name": "<NAME>",                         // no change
    "digest": { "<ALGORITHM>": "<HEX_VALUE>"  } // optional 
    "uri": "<SCHEME://HIERARCHY>"             // optional, proposed
    "uuid": "<UUID-VALUE>"                    // optional, hypothetical
  }
]
```
That should work (assuming that at least one of digest/uri/uuid must be
specified), but I'll resist the temptation to expand the scope of
this proposal.

@marcelamelara suggested grouping the non-name fields under an `id` 
sub-object. I think that would be a breaking change for `digest`. I do 
agree that `name` is different with its application-defined semantics,
compared to what I consider the more semantically important `digest`,
`uri`, or even `uuid` fields.

wietse-gmail:
> To @mlieberman85 's point, I think we were thinking about things in a bit of a different way of using URI - and maybe that's going to be a separate field.
> 
> We are thinking that `uri` and `digest` aren't mutually exclusive. For example, we want to include the digest and have the URI point to a blob store, such as `s3://..../app.spdx`, or `oci://...` or `gcs://...`. This would also potentially create a cleaner separation between rekor and its reliance on a blob store and indexing.

If both digest set and uri need to be in the subject, is that because one needs to match both the digest set (for integrity) and uri (to retrieve the attested-to resource from the blob store)?That would avoid ambiguity.

marcelamelara:
I wanted to bump this PR, per my comment https://github.com/in-toto/attestation/issues/5#issuecomment-1277960460.

It seems like there was consensus around adding an optional URI field to the subject, but we needed to be clear about the semantics of the subject when digest *and* URI are both specified?

laurentsimon:
Would we also allow PURL?

feelepxyz:
> Would we also allow PURL?

@laurentsimon we're currently adding a PURL in the subject `name` for a published npm package as part of adding provenance to npm packages. 

Does `name` not already cover the purl case? I'm mainly concerned with settling on something that we can keep stable over the long term as we'd like matching to keep working from old npm clients against new attestation documents. So ideally we could update to in-toto v1, and update predicate types as we wish without breaking old clients that are trying to match against the attestation subject.name and subject.digest.

Reading between the lines, are we really talking about some kind of "one of" rule, where you must use either a `digest`, `uri`, or `uuid`, and maybe `purl`? Or should they all be optional and should it be possible to have all of them set?

Simplified one-of schema:
```
"subject": [
  {
    "name": "<NAME>", // optional
    "oneOf": [{    
      "digest": { "<ALGORITHM>": "<HEX_VALUE>"  }
    }, {
      "uri": "<SCHEME://HIERARCHY>"
    }, {
      "uuid": "<UUID-VALUE>"
    }, {
      "purl": "<PURL-VALUE>"
    }]
  }
]
```

feelepxyz:
Although, I just spotted this reference to PURLs: https://github.com/in-toto/attestation/blob/3b9f1a206317c015986b5ba9c96196303d24053c/spec/field_types.md?plain=1#L31-L44 - @MarkLodato what is this line referring to? Is it a `_ResourceURI` attribute in the in-toto attestation document?

marcelamelara:
@feelepxyz Thanks for the suggestion! I like the idea of a `oneOf` field for the IDs/URIs. But I will push back on including the `digest` in that subfield. 

For one, there are use cases where `digest` and `uri`/`uuid`/`purl` should not be mutually exclusive. I believe this was described to some extent above, but the type of use case I want to make sure we capture is when the attestation and the actual subject artifact are not directly bound (e.g., if the attestation is in a Rekor log while the artifact sits on some web/ftp server). In these cases, the `uri` points to the ftp server, while the `digest` enables matching/integrity checking of the artifact itself.

There's also the point @wietse-gmail made, which is that moving the `digest` as a sub-field of another field would break backward-compatibility of the `subject`. So I agree that keeping the `digest` at the same level as `name` would be helpful from this standpoint.

SantiagoTorres:
Wouldn't it be better to use `name` as an identifier (e.g., pURL) instead of using it as an integrity marker?

marcelamelara:
> Wouldn't it be better to use `name` as an identifier (e.g., pURL) instead of using it as an integrity marker?

I've always interpreted `name` to be an identifier, not an integrity marker. My only concern with using `name` for a variety of different identifier types is how a verifier would determine how to interpret `name`. Whereas having a union for different _resolvable_ identifiers, would reduce the ambiguity, IMO.

MarkLodato:
> @MarkLodato what is this line referring to? Is it a `_ResourceURI` attribute in the in-toto attestation document?

Oops, looks like it's no longer referenced anywhere. An earlier version had referenced it.

MarkLodato:
I'm still worried about semantics. A "URI" (or "UUID" etc.) is just a type, not a semantic meaning. What we need to encode is how the field should be interpreted. We have at least the following distinct purposes:

- Debugging information
- A hint to where the artifact can be downloaded
- A selector between different artifacts within the attestation, e.g. to differentiate an output binary from a log file within the same provenance attestation
- A globally unique identifier as an alternative to `digest`, e.g. [scorecards](https://github.com/ossf/scorecard) attests that a repo (not a particular commit) had whatever properties at a point in time
- A globally unique identifier in addition to `digest`, e.g. some authorization attests that a Docker repository label (URI) pointed to a particular digest at a point in time.

Perhaps there are others.

I'm worried that these uses will get mixed up and that it will not be possible to tell what the attester intended.

TomHennen:
@MarkLodato @wietse-gmail do you still want to pursue this PR?  Should it be closed/rejected?  @marcelamelara suggests it might be handled by #117?

Also @AdamZWu who might have context on situations where we don't have a subject/digest?

wietse-gmail:
I do not have a use case for this. Since I opened this PR, should I take
action to close it?

Issue #28, opened 2021-04-28
Consider adding a "digest kind" and/or "content type" to subject


MarkLodato:
Currently, the attestation subject is a pure content digest. There are two related pieces of information we **may** want to consider adding:

1. "digest kind": how to serialize the artifact into the hashing algorithm. Examples:
    - PE file: straight hash or the hash used for Authenticode?
    - JAR: straight hash or the hash used for JAR signing?
    - Git: commit ID or tag ID (if an annotated tag)?
2. "content type": how the content is intended to be interpreted. Examples:
    - Docker image
    - Git commit
    - JPEG file
    - ZIP file

The main question is: **Are there any compelling security or implementation reasons why we need this?** If not, we leave it out to simplify things. If there are, then we need to come up with a design that overcomes the challenges below.

### Background

Originally we had included something like this into the property name of [DigestSet](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#DigestSet), e.g. `"subject": {"gitCommit": "<sha1>"}`. If you used "sha256", it meant straight file hash (no particular content type), but specific content types would use a digest kind appropriate for that type. Ultimately we decided against that because it added complexity without an obvious benefit.

Prior art: Rekor has a [type registry](https://github.com/sigstore/rekor/tree/main/pkg/types) that appears to be the same as "digest kind".

### Potential benefits

1. "Digest Kind" could tell verifiers how to compute the hash.

    - Counterpoint: They can just compute it in some canonical way or in multiple ways, without having to record it.
    - Counterpoint: In many applications, the hashing is done prior to reading the attestation, so having this information wouldn't help.

2. There could be security vulnerabilities if a producer intended one kind/type and a consumer interpreted it as another. For example, the producer signed a PE Authenticode hash, but then the consumer interpreted it as a raw file hash.

    - Counterpoint: This is too abstract to warrant the cost. We need some sort of proof-of-concept to show that this would be exploitable in a somewhat realistic scenario.

### Challenges

1. Do you include the kind/type in the matching? For example, if one Provenance attestation lists a material `sha256: X, type: Y` and another attestation has `subject: X, type: Z`, does it match?

[...]

PR #108, opened 2022-09-23
Add `goModuleH1` to the list of predefined digests

MarkLodato:
This is the first digest algorithm that is not a straight cryptographic hash over a blob of data.

Decisions:
- Use the name `goModuleH1` instead of `goDirhash` or similar because (a) it makes it more clear that this is only for go modules and (b) to allow future H2, H3, etc. to be expressed in the same DigestSet.
- Use hex encoding instead of base64, since that is what we use elsewhere.

See #28 for background.

### DIFF
--- a/spec/field_types.md
+++ b/spec/field_types.md
@@ -11,10 +11,27 @@ _DigestSet (object)_
 > It is RECOMMENDED to use at least `sha256` for compatibility between producers
 > and consumers.
 >
-> Pre-defined algorithm names: `sha256`, `sha224`, `sha384`, `sha512`,
-> `sha512_224`, `sha512_256`, `sha3_224`, `sha3_256`, `sha3_384`, `sha3_512`,
-> `shake128`, `shake256`, `blake2b`, `blake2s`, `ripemd160`, `sm3`, `gost`,
-> `sha1`, `md5`.
+> Pre-defined algorithms:
+>
+> -   `sha256`, `sha224`, `sha384`, `sha512`, `sha512_224`, `sha512_256`,
+>     `sha3_224`, `sha3_256`, `sha3_384`, `sha3_512`, `shake128`, `shake256`,
+>     `blake2b`, `blake2s`, `ripemd160`, `sm3`, `gost`, `sha1`, `md5`
+>
+>     Standard cryptographic hash algorithms, for cases when when the method of
+>     serialization is obvious or well known.
+>
+> -   `goModuleH1`: The go module [directory Hash1], omitting the "h1:"
+>     prefix and output in hexadecimal instead of base64. Can be computed over a
+>     directory named `name@version`, or the contents of zip file containing
+>     such a directory:
+>
+>     ```bash
+>     find name@version -type f | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
+>     ```
+>
+>     For example, the module dirhash
+>     `h1:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=` would be encoded as
+>     `{"goModuleH1": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"}`.
 >
 > Consumers MUST only accept algorithms that they consider secure and MUST
 > ignore unrecognized or unaccepted algorithms. For example, most applications
@@ -27,6 +44,8 @@ _DigestSet (object)_
 > -   `{"sha256": "abcd", "sha512": "1234"}` matches `{"sha256": "abcd"}`
 > -   `{"sha256": "abcd"}` does not match `{"sha256": "fedb", "sha512": "abcd"}`
 > -   `{"somecoolhash": "abcd"}` uses a non-predefined algorithm
+
+[directory Hash1]: https://cs.opensource.google/go/x/mod/+/refs/tags/v0.5.0:sumdb/dirhash/hash.go
 
 <a id="ResourceURI"></a>
 _ResourceURI (string)_


spec/field_types.md
```diff
@@ -11,10 +11,27 @@ _DigestSet (object)_
...
+>     directory named `name@version`, or the contents of zip file containing
+>     such a directory:
+>
+>     ```bash
+>     find name@version -type f | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
+>     ```
+>
+>     For example, the module dirhash
+>     `h1:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=` would be encoded as
+>     `{"goModuleH1": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"}`.
```
TomHennen, line 34: To what extent are go developers familiar with the base64 version?

I'm somewhat concerned folks will be surprised that the go reported hash _string_ differs from the hash reported in the provenance.  Would it be acceptable to just state in the docs that this is base64 encoded?  Verifiers will need to use special code to compute this value anyways, so it seems like doing so wouldn't add any complexity.

TomHennen (changes requested):
Thanks Mark!

spec/field_types.md
MarkLodato: I don't see a good solution here. Line 7 says that the value is always a lowercase hex encoding of the digest. Therefore, our options seem to be:

- Encode it using hex for consistency with all the other algorithms, at the cost of divergence from what go users see. (My suggestion.)
- Change the spec so that encoding is per-algorithm. This is inconsistent with the other fields, and is technically a breaking change but probably OK since no one should be parsing unrecognized fields anyway. (Your suggestion.)
- Make this the hex encoding of the base64, which seems like a bad idea.

Does anyone else have opinions?

TomHennen: What if "Pre-defined algorithms" require that it be the lowercase hex encoding, but allow custom values (like this one) to do what makes the most sense in context?  I'm not sure this being a breaking change would be a big problem?  [This note](https://github.com/in-toto/attestation#important) even indicates the contents are subject to change until 1.0.  If needed I think we could bump to 0.2.0.

Perhaps the most important feedback we could get would be from some Go experts.

spec/field_types.md
```diff
@@ -11,10 +11,27 @@ _DigestSet (object)_
...
+>     Standard cryptographic hash algorithms, for cases when when the method of
+>     serialization is obvious or well known.
+>
+> -   `goModuleH1`: The go module [directory Hash1], omitting the "h1:"
+>     prefix and output in hexadecimal instead of base64. Can be computed over a
+>     directory named `name@version`, or the contents of zip file containing
+>     such a directory:
+>
+>     ```bash
+>     find name@version -type f | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
```
khalkie, line 29: Is this the actual command to use in order to generate the goModuleH1 value on a Go module directory? 

Sorry if this is a stupid question, I'm trying to add it into our provenance generation and as a sanity check I'm running it on different copies of the same source to validate that the value is the same (i.e. I would expect the hash to be the same if the file contents are the same). 

Actually it gives a different hash value depending on the folder path passed into the command - am I using it wrong?

i.e.

khalk@khalk:~/tempgit$ git clone https://github.com/golang/example
khalk@khalk:~/tempgit$ find ~/tempgit/example -type f | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
45bde95411f095cd1040f6e5ad0c3f25f3c7ec8b9b048a32ceef01bac110e010
khalk@khalk:~/tempgit$ cd example
khalk@khalk:~/tempgit/example$ find . -type f | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
51d525520df583ea4c543d00381eb197c4190e39d92ff37feee1ee048a92e2e6

MarkLodato: The problem is that you need to literally type `find name@version` because the "name@version" is part of the output. If you use "~/tempgit/example" or "." or "./name@version" you get a different result. I need to update the text to make this more clear.

Here's a script that is less error prone: https://gist.github.com/MarkLodato/c03659d242ea214ef3588f29b582be70

MarkLodato: The name of the directory itself does not matter. What matters is that the outpuf of the first `sha256sum` command looks like this:

```
abcd1234...90  example@v1.0/file.go
dcba4321...88  example@v1.0/another-file.go
...
```

That is, a relative path whose first component is `<name>@<version>`.

My script does a `sed` command to replace the actual directory name with the second parameter, which would be another way to achieve this.

Alternatively, you could write a small go program to call the official [HashDir](https://cs.opensource.google/go/x/mod/+/master:sumdb/dirhash/hash.go;l=70;drc=346a37af5599be02f125bd8bc0a5e1d33db21ddc) function.

MarkLodato:
Thanks, Tom. I think it probably would be valuable, just hasn't been at the top of my list. The main thing seems to be the encoding (hex vs base64). Do we want to live with different encodings, force this new one to use hex, or use a new version of the attestation spec that switches to base64?

TomHennen:
I agree that that's the only sticking point.  I'd probably argue that it's better to have something rather than nothing so I can go either way (even as far as suggesting that everyone be able to support _both_ hex and base64 when handling hashes).

MarkLodato:
My inclination is to:

- For now, use hex since that's what the spec says. Yes it's a bit awkward for someone used to base64, it's not a huge deal.
- If desired, do a new release of the spec that supports something else, perhaps base64 for everything.

My concern with doing base64 with v0.1 is that (1) it's a breaking change and (2) it's more confusing to have some things base64 and some thing hex.

TomHennen:
Ok, let's go with this.

TomHennen (approved):

MarkLodato:
Thanks. I need one more approver to submit.

marcelamelara:
@MarkLodato @TomHennen Let's open an issue to track the addition of base64-encoded digests to the spec, if we don't already have one.

marcelamelara (approved):

MarkLodato:
Thanks!

PR #111, opened 2022-10-05
Add runtime trace predicate type

adityasaky:
This is an early draft of a runtime trace predicate type. We've noted some subsections / fields that need more input.

cc @pxp928 @nadgowdas @jedsalazar

### DIFF
--- a/spec/predicates/runtime-trace.md
+++ b/spec/predicates/runtime-trace.md
@@ -1,0 +1,273 @@
+# Predicate type: Runtime Trace
+
+Type URI: https://in-toto.io/attestation/runtime-trace/v0.1
+
+Version: 0.1.0
+
+Authors: Parth Patel (@pxp928), Shripad Nadgowda (@nadgowdas),
+  Aditya Sirish A Yelgundhalli (@adityasaky)
+
+## Purpose
+
+This predicate is used to describe system events that were part of some software
+supply chain step, for example, the build process of an artifact. Supply chain
+operations today are complex and opaque with little to no insights into
+underlying system operations. However, there are emerging introspection
+techniques that now allow us to monitor and capture system event logs during the
+monitored operation.
+
+## Use Cases
+
+Generally, this predicate can be used to express a runtime trace of any
+operation. The predicate specification does not mandate any specific monitoring
+tool or technology. The schema can be used to express runtime traces of anything
+that can be traced or monitored, from an operation spawned by a user using a CLI
+command to tasks performed in CI systems. The primary use case driving the
+development of this predicate is its applicability to runtime traces of build
+processes.
+
+This attestation can be used in conjunction with provenance attestations to
+leverage observability to attest to various SLSA requirements. The runtime trace
+can prove the build was invoked via a script, that the build was executed in a
+hermetic environment with no network access, and so on.
+
+## Prerequisites
+
+Understanding of monitoring tools and the larger in-toto attestation framework.
+
+## Model
+
+`monitor` identifies the specific instance of a tool that’s observing the target
+process and details the configurations of the monitor, while `monitoredProcess`
+contains information that when put together uniquely identifies the exact
+instance of the job being observed. The `monitorLog` field contains the actual
+runtime trace information.
+
+## Schema
+
+```json
+{
+    "_type": "https://in-toto.io/Statement/v1",
+    "subject": [{ ... }],
+    "predicateType": "https://in-toto.io/attestation/runtime-trace/v0.1",
+    "predicate": {
+        "monitor": {
+            "type": "<TypeURI>",
+            "configSource": "<ResourceDescriptor>",
+            "tracePolicy": { /* object */ }
+        },
+        "monitoredProcess": {
+            "hostID": "<URI>",
+            "type": "<URI>",
+            "event": "<STRING>"
+        },
+        "monitorLog": {
+            "process": [
+                { /* object */ }
+            ],
+            "network": [
+                { /* object */ }
+            ],
+            "fileAccess": ["<ResourceDescriptor>", ...]
+        },
+        "metadata": {
+            "buildStartedOn": "<TIMESTAMP>",
+            "buildFinishedOn": "<TIMESTAMP>"
+        }
+    }
+}
+```
+
+### Parsing Rules
+
+This predicate follows the in-toto attestation parsing rules. Summary:
+
+-   Consumers MUST ignore unrecognized fields.
+-   The `predicateType` URI includes the major version number and will always
+    change whenever there is a backwards incompatible change.
+-   Minor version changes are always backwards compatible and “monotonic.” Such
+    changes do not update the `predicateType`.
+-   Producers MAY add extension fields using field names that are unlikely to
+    collide with names used by other producers. Field names SHOULD avoid using
+    characters like `.` and `$`.
+-   Fields marked _optional_ MAY be unset or null, and should be treated
+    equivalently. Both are equivalent to empty for object or array values.
+
+### Fields
+
+`monitor` _object_, _required_
+
+Identifies the specific monitor instance used to trace the runtime.
+
+`monitor.type` _TypeURI_, _required_
+
+URI indicating the monitor’s type.
+
+`monitor.configSource` ResourceDescriptor, _optional_
+
+Effectively a pointer to the monitor's configuration.
+
+`monitor.tracePolicy` _object_, _optional_
+
+Indicates the trace policy used for the monitoring event that generated the
+runtime trace. The trace policy's format is dependent on the monitor used, and
+so it must be parsed after identifying the monitor using `monitor.type`.
+
+FIXME: should this be optional?
+
+`monitoredProcess` _object_, _required_
+
+Identifies the process being monitored.
+
+`monitoredProcess.hostID` _string (URI)_, _required_
+
+URI indicating the process host’s identity.
+
+`monitoredProcess.type` _string (TypeURI)_, _required_
+
+URI indicating the type of process performed. Ex: when monitoring a build, this
+field records the build type.
+
+`monitoredProcess.event` _string_, _required_
+
+String identifying the specific job or task associated with the attestation.
+
+`monitorLog` _object_, _required_
+
+Record of events that were traced by the monitor. At least one of `process`,
+`network`, and `fileAccess` must be present.
+
+`monitorLog.process` _list of objects_, _optional_
+
+Record of processes observed by monitor. The exact format of this field is
+currently dependent on the monitor, and can be determined using `monitor.type`.
+In future, after consulting different monitors, this field may have a consistent
+schema.
+
+`monitorLog.network` _list of objects_, _optional_
+
+Record of network activity observed by monitor. The exact format of this field
+is currently dependent on the monitor, and can be determined using
+`monitor.type`. In future, after consulting different monitors, this field may
+have a consistent schema.
+
+`monitorLog.fileAccess` _list of ResourceDescriptor objects_, _optional_
+
+Record of files accessed during the monitored process. A complete list of
+_materials_ can be derived from this information. Each entry in this list is
+expected to record the path of the file and one or more digests of the file, but
+as each entry is an instance of _ResourceDescriptor_, other supported
+information can also be captured. This field is a list rather than a key-value
+map because a single file may be used multiple times during the build process.
+Further, some files that are accessed may _change_ during the build process, and
+so, different entries for the same file may have different digests.
+
+Note: While this predicate can be used to log file accesses, the actual
+technique used to capture the file access event has some implications. If a
+synchronous monitor, for example one that uses `ptrace` to trace the file access
+system calls, is used, then the build process can be paused while the file's
+digest is calculated and stored. However, asynchronous monitors such as those
+using eBPF cannot pause the build process before the file is actually used.
+Therefore, they cannot make as strong guarantees about the digests of the files
+accessed. Verifiers using runtime trace attestations for file accesses must be
+careful about what guarantees they are actually getting based on how the build
+process was monitored.
+
+`metadata` _object_, _optional_
+
+Other properties of the monitoring event.
+
+`metadata.buildStartedOn` _Timestamp_, _optional_
+
+`metadata.buildFinishedOn` _Timestamp_, _optional_
+
+## Example
+
+```json
+{
+  "_type": "https://in-toto.io/Statement/v1",
+  "predicateType": "https://in-toto.io/attestation/runtime-trace/v0.1",
+  "subject": [
+    {
+      "name": "ttl.sh/testin123",
+      "digest": {
+        "sha256": "def2bdf8cee687d5889d51923d7907c441f1a61958f1e5dfb07f53041c83745f"
+      }
+    }
+  ],
+  "predicate": {
+    "monitor": {
+      "type": "https://github.com/cilium/tetragon/v0.8.4",
+      "configSource": {},
+      "tracePolicy": {
+        "policies": [
+          {
+            "Name": "connect",
+            "Config": ""
+          },
+          {
+            "Name": "sys-read-follow-prefix",
+            "Config": ""
+          }
+        ]
+      }
+    },
+    "monitoredProcess": {
+      "hostID": "https://tekton.dev/chains/v2",
+      "type": "https://tekton.dev/attestations/chains@v2",
+      "event": "run-image-pipelinerun-build-trusted"
+    },
+    "monitorLog": {
+      "process": [
+        {
+          "eventType": "process",
+          "processBinary": "/ko-app/entrypoint",
+          "arguments": [
+            "init /ko-app/entrypoint /tekton/bin/entrypoint step-prepare step-create step-results"
+          ],
+          "privileged": null
+        },
+        {
+          "eventType": "process",
+          "processBinary": "/bin/sh",
+          "arguments": [
+            "-c \"scriptfile=\"/tekton/scripts/script-0-vjklb\"\ntouch ${scriptfile} && chmod +x ${scriptfile}\ncat > ${s\" \"riptfile} << '_EOF_'\nIyEvdXNyL2Jpbi9lbnYgYmFzaApzZXQgLWUKCmlmIFtbICJ0cnVlIiA9PSAidHJ1ZSIgXV07IHRoZW\" KICBlY2hvICI+IFNldHRpbmcgcGVybWlzc2lvbnMgb24gJy93b3Jrc3BhY2UvY2FjaGUnLi4uIgogIGNob3duIC1SICIxMDAwOj wMDAiICIvd29ya3NwYWNlL2NhY2hlIgpmaQoKZm9yIHBhdGggaW4gIi90ZWt0b24vaG9tZSIgIi9sYXllcnMiICIvd29ya3NwYW lL3NvdXJjZSI7IGRvCiAgZWNobyAiPiBTZXR0aW5nIHBlcm1pc3Npb25zIG9uICckcGF0aCcuLi4iCiAgY2hvd24gLVIgIjEwMD 6MTAwMCIgIiRwYXRoIgoKICBpZiBbWyAiJHBhdGgiID09ICIvd29ya3NwYWNlL3NvdXJjZSIgXV07IHRoZW4KICAgICAgY2htb2 gNzc1ICIvd29ya3NwYWNlL3NvdXJjZSIKICBmaQpkb25lCgplY2hvICI+IFBhcnNpbmcgYWRkaXRpb25hbCBjb25maWd1cmF0aW uLi4uIgpwYXJzaW5nX2ZsYWc9IiIKZW52cz0oKQpmb3IgYXJnIGluICIkQCI7IGRvCiAgICBpZiBbWyAiJGFyZyIgPT0gIi0tZW 2LXZhcnMiIF1dOyB0aGVuCiAgICAgICAgZWNobyAiLT4gUGFyc2luZyBlbnYgdmFyaWFibGVzLi4uIgogICAgICAgIHBhcnNpbm fZmxhZz0iZW52LXZhcnMiCiAgICBlbGlmIFtbICIkcGFyc2luZ19mbGFnIiA9PSAiZW52LXZhcnMiIF1dOyB0aGVuCiAgICAgIC"
+          ],
+          "privileged": null
+        },
+        {
+          "eventType": "process",
+          "processBinary": "/bin/touch",
+          "arguments": [
+            "/tekton/scripts/script-0-vjklb"
+          ],
+          "privileged": null
+        },
+        ...,
+        {
+          "eventType": "exit",
+          "processBinary": "/bin/sh",
+          "arguments": [
+            "",
+            "-c \"scriptfile=\"/tekton/scripts/script-0-vjklb\"\ntouch ${scriptfile} && chmod +x ${scriptfile}\ncat > ${s\" \"riptfile} << '_EOF_'\nIyEvdXNyL2Jpbi9lbnYgYmFzaApzZXQgLWUKCmlmIFtbICJ0cnVlIiA9PSAidHJ1ZSIgXV07IHRoZW\" KICBlY2hvICI+IFNldHRpbmcgcGVybWlzc2lvbnMgb24gJy93b3Jrc3BhY2UvY2FjaGUnLi4uIgogIGNob3duIC1SICIxMDAwOj wMDAiICIvd29ya3NwYWNlL2NhY2hlIgpmaQoKZm9yIHBhdGggaW4gIi90ZWt0b24vaG9tZSIgIi9sYXllcnMiICIvd29ya3NwYW lL3NvdXJjZSI7IGRvCiAgZWNobyAiPiBTZXR0aW5nIHBlcm1pc3Npb25zIG9uICckcGF0aCcuLi4iCiAgY2hvd24gLVIgIjEwMD 6MTAwMCIgIiRwYXRoIgoKICBpZiBbWyAiJHBhdGgiID09ICIvd29ya3NwYWNlL3NvdXJjZSIgXV07IHRoZW4KICAgICAgY2htb2 gNzc1ICIvd29ya3NwYWNlL3NvdXJjZSIKICBmaQpkb25lCgplY2hvICI+IFBhcnNpbmcgYWRkaXRpb25hbCBjb25maWd1cmF0aW uLi4uIgpwYXJzaW5nX2ZsYWc9IiIKZW52cz0oKQpmb3IgYXJnIGluICIkQCI7IGRvCiAgICBpZiBbWyAiJGFyZyIgPT0gIi0tZW 2LXZhcnMiIF1dOyB0aGVuCiAgICAgICAgZWNobyAiLT4gUGFyc2luZyBlbnYgdmFyaWFibGVzLi4uIgogICAgICAgIHBhcnNpbm fZmxhZz0iZW52LXZhcnMiCiAgICBlbGlmIFtbICIkcGFyc2luZ19mbGFnIiA9PSAiZW52LXZhcnMiIF1dOyB0aGVuCiAgICAgIC"
+          ],
+          "privileged": null
+        },
+        {
+          "eventType": "exit",
+          "processBinary": "/tekton/bin/entrypoint",
+          "arguments": [
+            "",
+            "-wait_file /tekton/downward/ready -wait_file_content -post_file /tekton/run/0/out -termination_path /tekton/termination -step_metadata_dir /tekton/run/0/status -results APP_IMAGE_DIGEST,APP_IMAGE_URL -entrypoint /tekton/scripts/script-0-vjklb -- --env-vars"
+          ],
+          "privileged": null
+        }
+      ]
+    },
+    "metadata": {
+      "buildStartedOn": "2022-10-13T22:55:50Z",
+      "buildFinishedOn": "2022-10-13T22:56:17Z"
+    }
+  }
+}
+```
--- a/spec/predicates/README.md
+++ b/spec/predicates/README.md
@@ -16,6 +16,8 @@ our [vetting process], and may be of general interest:
 -   [Link]: For migration from [in-toto 0.9].
 -   [SCAI Report]: Evidence-based assertions about software artifact and
     supply chain attributes or behavior.
+-   [Runtime Traces]: To capture runtime traces of software supply chain
+    operations.
 
 [Link]: link.md
 [New Predicate Guidelines]: ../../docs/new_predicate_guidelines.md
@@ -23,3 +25,4 @@ our [vetting process], and may be of general interest:
 [SLSA Provenance]: https://slsa.dev/provenance
 [in-toto 0.9]: https://github.com/in-toto/docs/blob/master/in-toto-spec.md#44-file-formats-namekeyid-prefixlink
 [vetting process]: ../../docs/new_predicate_guidelines.md#vetting-process
+[Runtime Traces]: runtime-trace.md


spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,126 @@
...
+     "configSource": {
+        "uri": "<URI>",
+        "digest": { /* DigestSet */ },
+        "entryPoint": "<STRING>"
+      },
+      "tracePolicy": {}
+    },
+    "monitorLog": {
+      "process": [
+         { /* object */ }
```
colek42, line 49: Could we define "object" better here?  Normalizing output from various trace tools will be helpful in discoverability.

I have some rough ideas of that could look like in this branch.  https://github.com/testifysec/go-witness/blob/tetragon/attestation/commandrun/commandrun.go

pxp928: Hey Cole,

yes we have defied it in the demo example as the following. Would be great to get feedback :
``` {
          "eventType": "kprobe",
          "functionName": “__x64_sys_write/fd_install”,
          “processBinary” : “/bin/cat”
          “exitCode”: int
          "arguments": [
               "-wait_file-post_file ",
               "/tekton/run/2/out",
               "/tekton/run/3/out",
               "-termination_path",
               "/tekton/termination",
               "-step_metadata_dir",
               "/tekton/run/3/status",
               "-entrypoint",
               "/tekton/scripts/script-3-g5chp",
               "--",
          ],
          "privileged": [
                 “CAP_SYS_ADMIN”,
                 ],
          },
          {
          "eventType": "kprobe",
          "functionName": “tcp_connect”,
          “processBinary” : “”
          "arguments": [
               "tcp",
               "10.244.0.6:34965",
               "104.198.14.52:8",
          ],
          "privileged": [],
    },
```
we do want to add hash of files and such as that might be useful. Looking at your branch we are very much on the same page and would love to work together on it. I think there is information that you include that would be valuable to add into the predicate.

colek42: We will probably adopt whatever schema is defined here.  The only issue we have is that the data needs to be normalized.  For our use case, we would like the user to be able to define the same policy no matter what trace tool was used.  

It looks like the schema here is from Tetragon.  I think this needs to be mutated into a comment format, and we should add another field(s) to define the trace provider.

pxp928: 100% agree. We do want to make this generic that other tools can be used. Adding in fields that are missing. As for trace provider we do have:
```
  "predicate": {
    "monitor": {
      "id": "<URI>"
    },
```
which would define which tool is being used.

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2666 @@
...
+    "monitorType": "https://tetragon",
+    "build": {
+      "builderID": "https://tekton.dev/chains/v2",
+      "type": "https://tekton.dev/attestations/chains@v2",
+      "event": "run-image-pipelinerun-build-trusted"
+    },
+    "invocation": {
+      "configSource": {},
+      "tracePolicy": {
+        "policies": [
```
pxp928, line 151: Trace policies would also need to define how they are configured. Below is a snippet from an output from tetragon. This still needs to be re-worked and formatted to be much more generalized. Not a representation of the final format, but wanted to put an example here for discussion.
```
"policies": [
          {
            "Name": "connect",
            "Config": "Name: /var/lib/tetragon/bpf_generic_kprobe_v53.o, Attach: tcp_connect, Label: kprobe/generic_kprobe, PinPath: gkp-sensor-1-gkp-0-tcp_connect_prog, RetProbe: false, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:0};Name: /var/lib/tetragon/bpf_generic_kprobe_v53.o, Attach: tcp_close, Label: kprobe/generic_kprobe, PinPath: gkp-sensor-1-gkp-1-tcp_close_prog, RetProbe: false, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:1};Name: /var/lib/tetragon/bpf_generic_kprobe_v53.o, Attach: tcp_sendmsg, Label: kprobe/generic_kprobe, PinPath: gkp-sensor-1-gkp-2-tcp_sendmsg_prog, RetProbe: false, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:2}"
          },
          {
            "Name": "sys-read-follow-prefix",
            "Config": "Name: /var/lib/tetragon/bpf_generic_kprobe_v53.o, Attach: fd_install, Label: kprobe/generic_kprobe, PinPath: gkp-sensor-4-gkp-3-fd_install_prog, RetProbe: false, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:3};Name: /var/lib/tetragon/bpf_generic_kprobe_v53.o, Attach: __x64_sys_close, Label: kprobe/generic_kprobe, PinPath: gkp-sensor-4-gkp-4-__x64_sys_close_prog, RetProbe: false, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:4};Name: /var/lib/tetragon/bpf_generic_kprobe_v53.o, Attach: __x64_sys_read, Label: kprobe/generic_kprobe, PinPath: gkp-sensor-4-gkp-5-__x64_sys_read_prog, RetProbe: false, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:5};Name: /var/lib/tetragon/bpf_generic_retkprobe_v53.o, Attach: __x64_sys_read, Label: kprobe/generic_retkprobe, PinPath: gkp-sensor-4-gkp-5-__x64_sys_read_ret_prog, RetProbe: true, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:5};Name: /var/lib/tetragon/bpf_generic_kprobe_v53.o, Attach: __x64_sys_write, Label: kprobe/generic_kprobe, PinPath: gkp-sensor-4-gkp-6-__x64_sys_write_prog, RetProbe: false, ErrorFatal: true, Override: false, Type: generic_kprobe, LoaderDate: {ID:6}"
          }
        ]
      }
````

**Comment by colek42** [comment]
I uploaded a sample of our schema here: https://gist.github.com/colek42/a9c82d3314f2cb109bfbb0b67c5e0f99

This is the data that was generated during the SecurityCon demo.

**Comment by adityasaky** [comment]
@itaysk I'm attending the conference! I'll message you on Slack, let's find some time to chat.

File: spec/predicates/runtime-trace.md [hunk 3]
```diff
@@ -0,0 +1,2677 @@
...
+      "digest": {
+        "sha256": "def2bdf8cee687d5889d51923d7907c441f1a61958f1e5dfb07f53041c83745f"
+      }
+    }
+  ],
+  "predicate": {
+    "monitor": {
+      "id": "https://tetragon"
+    },
+    "monitorType": "https://tetragon",
```
itaysk, line 153: I didn't see `monitorType` defined earlier

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+Identifies the specific monitor instance used to trace the runtime.
+
+`monitor.id` _string (URI)_, _required_
+
+URI indicating the monitor’s identity.
+
+`build` _object_, _required_
+
+Feedback required: Should this predicate be scoped to runtime traces of builds only or generalize to runtime traces of any activity?
```
itaysk, line 85: I think we need to either rename the attestation to mention build, or try to generalize the all the build-specific fields (most obvious example is the field called `build`). currently it doesn't make sense

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+`monitorLog` _object_, _required_
+
+Feedback required: Should different types of monitors (process vs network for example) emit different attestation subtypes? If they are all combined in one, should there still be some differentiation within `monitorLog`?
+
+`monitorLog.process` _list_, _optional_
+
+`monitorLog.network` _list_, _optional_
+
+`monitorLog.fileAccess` _list_, _optional_
```
itaysk, line 119: Just a note, that for us including reads was way too verbose. instead we capture only writes

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+Record of files accessed during the build process. A complete list of _materials_ can be derived from this information. Each entry in this list is expected to record the path of the file and one or more digests of the file. This field is a list rather than a key value map because a single file may be used multiple times during the build process. Further, some files that are accessed may _change_ during the build process, and so, different entries may have different digests.
```
itaysk, line 121: every other monitorLog (process, network) was abstract, this is the only one that specifies the contents of the trace. I think at this stage we should keep all of them abstract.

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+While this predicate can be used to log file accesses, the actual technique used to capture the file access event has some implications. If a synchronous monitor, for example one that uses `ptrace` to trace the file access system calls, is used, then the build process can be paused while the file's digest is calculated and stored. However, asynchronous monitors such as those using eBPF cannot pause the build process before the file is actually used. Therefore, they cannot make as strong guarantees about the digests of the files accessed. Verifiers using runtime trace attestations for file accesses must careful about what guarantees they are actually getting based on how the build process was monitored.
+
+Feedback required: There were discussions about including a field for "materials" derived from file accesses. Should it be within monitorLog?
```
itaysk, line 125: I didn't see how materials is related to "runtime trace", so I think no, but I might have missed something

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+`invocation` _object_, _optional_
+
+Identifies the event that initiated the monitor. Similar to `invocation` in SLSA Provenance.
+
+`invocation.configSource` _object_, _optional_
+
+Effectively a pointer to the monitor's configuration.
+
+`invocation.tracePolicy` _object_, _optional_
```
itaysk, line 107: IMO `tracePolicy` should move under `monitor`. Different tracers have different features that might not be available across the board. also the syntax will definitely be different between tracers and it will make most sense to keep the trace policy in it's original form. so if the tracePolicy is monitor-specific it's clearer to move it there.

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+Feedback required: Should tracingPolicy point outwards to a separate attestation with the same subject that wraps around the actual policy? The policy can be more securely defined and distributed (signed for by developer keys rather than online keys for example).
+
+`monitorLog` _object_, _required_
+
+Feedback required: Should different types of monitors (process vs network for example) emit different attestation subtypes? If they are all combined in one, should there still be some differentiation within `monitorLog`?
```
itaysk, line 113: could you please clarify this question?

spec/predicates/runtime-trace.md
itaysk, line 107: Should we define what is the scope of the trace or it is up to the implementation to decide? For example, in GH Actions, each run gets a runner vm which executes the workflow. a tracer will likely be installed at the runner (vm) level, so it could trace the entire runner or be scoped to events just from the workflow.
I'm putting this note here because it could also be just part of the trace policy.

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+`monitorLog.process` _list_, _optional_
```
itaysk, line 115: process args/envs might include secrets which the user would not want to logged. since we're not specifying the schema of the process monitor, I guess it's an issue for the implementor and not for this attestation. I thought I'd just make the note here anyway

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+`invocation.configSource` _object_, _optional_
+
+Effectively a pointer to the monitor's configuration.
+
+`invocation.tracePolicy` _object_, _optional_
+
+Feedback required: Should tracingPolicy point outwards to a separate attestation with the same subject that wraps around the actual policy? The policy can be more securely defined and distributed (signed for by developer keys rather than online keys for example).
+
+`monitorLog` _object_, _required_
```
itaysk, line 111: probably very subjective but trace might be more descriptive in this case than monitor. the attestation is also called runtime *trace*. and it's definition is a *trace* policy. so calling this monitor here is no standing out for me

itaysk (commented):
Good stuff! added some comments. 
BTW when I say "us","we","our projects" I mean https://github.com/aquasecurity/tracee and https://github.com/aquasecurity/tracee-action

spec/predicates/runtime-trace.md
colek42: We want to capture reads to generate a BOM.

spec/predicates/runtime-trace.md
colek42: In order to create policy over the attestation we need to have a defined schema for the monitor log.

spec/predicates/runtime-trace.md
colek42: +1

spec/predicates/runtime-trace.md
colek42: Should the spec require the implementer to specify black-listed env vars?  We could also hash the variables.

spec/predicates/runtime-trace.md
itaysk: oh I see. policy use case was not mentioned in this spec so I presumed the use case was to enrich provenance data. 
I have a feeling that discussing the detailed logs schema will be much longer then the outline schema. If that's the intention, let's see the schemas and then we can discuss.

spec/predicates/runtime-trace.md
itaysk: depending on whether this statement of mine is still true of not "since we're not specifying the schema of the process log" after https://github.com/in-toto/attestation/pull/111#discussion_r1095485398

marcelamelara (changes requested):
Overall, I think exposing runtime trace events for builders will be really useful. That said, I have two concerns with this predicate as it's proposed right now. 

(1) It's decently closely tied to the SLSA Provenance format; Will this predicate have to be updated in lockstep with every SLSA Provenance update? Maybe this won't be too bad if this predicate only needs updates with every major SLSA update or something like that, but it's still a consideration. So my recommendation would be to ensure this predicate is sufficiently decoupled from SLSA to minimize this.

(2) Given the example, I expect full runtime traces to be quite large in practice, generating a lot of data. and large attestations. Especially if the intent is to generate a runtime trace attestation at _every_ build. Is there a way this predicate could be more of a summary of a runtime trace that points to the full event log instead of including the full event log in the attestation? I'm not too familiar with different tracing tools, but I imagine that might make the predicate more interoperable between different tracing tools, too. I'm also thinking making heavier use of Reference- and/or SCAI-type fields might be helpful for this.

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+
+This predicate is used to describe system events that were part of the artifact build process. Artifact build processes today are complex and opaque with little to no insights into build actions. However, there are emerging introspection techniques that now allow us to monitor and capture system event logs during the build.
+
+## Prerequisites
+
+Understanding of monitoring tools and the larger in-toto attestation framework.
+
+## Model
+
+The structure of this predicate closely mirrors the structure of the SLSA Provenance attestation as they are complementary. `monitor` identifies the specific instance of a tool that’s observing the build, while `build` contains information that when put together uniquely identifies the exact instance of the job being observed. `invocation` details the configurations of `monitor` and the `monitorLog` contains the actual runtime trace information.
```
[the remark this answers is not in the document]
adityasaky: I'll go a step further and say that it should probably not refer to SLSA Provenance at all, and instead define its fields in a standalone manner. This may be something that goes into ITE-9 as well so we don't have issues with specs going out of sync with each other over time. It's hard to ensure references are updated otherwise. Thoughts?

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2677 @@
...
+      "fileAccess": [
+        { /* object */ }
+      ],
+    },
+    "metadata": {
+      "buildStartedOn": "<TIMESTAMP>",
+      "buildFinishedOn": "<TIMESTAMP>",
+    }
+  }
+}
```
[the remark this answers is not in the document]
adityasaky: My editor says they align :thinking:

spec/predicates/runtime-trace.md
adityasaky: I've updated this to use `process` instead of `build`. WDYT?

cc @pxp928 @nadgowdas

spec/predicates/runtime-trace.md
adityasaky: @pxp928

adityasaky: I think it's up to the implementation to decide. This spec should be about how the final trace is communicated and verified. The verification policy could check, I think, the scope and if the trace is usable for its context.

spec/predicates/runtime-trace.md
adityasaky: Fair! I think we've got a couple of different field rename suggestions to consider and implement here.

spec/predicates/runtime-trace.md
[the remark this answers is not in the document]
adityasaky: I took a pass at removing those mentions. I think it's fine to be inspired by another attestation type but we should refrain from referring to other specs unless updates to both are closely linked to each other.

spec/predicates/runtime-trace.md
```diff
@@ -80,49 +105,68 @@ Identifies the specific monitor instance used to trace the runtime.
...
-
-`build.event` _string_, _required_
+`process.event` _string_, _required_
 
 String identifying the specific job or task associated with the attestation.
 
 `invocation` _object_, _optional_
 
-Identifies the event that initiated the monitor. Similar to `invocation` in SLSA Provenance.
+Identifies the event that initiated the monitor.
```
marcelamelara, line 124: I almost wonder if `invocation` could be taken out and `invocation.configSource` moved under `process`? This may make even more sense if the `tracePolicy` is moved as suggested below.

spec/predicates/runtime-trace.md
```diff
@@ -4,37 +4,59 @@ Type URI: https://in-toto.io/runtime-trace/v0.1.0
...
+chain processes today are complex and opaque with little to no insights into
+underlying system operations. However, there are emerging introspection
+techniques that now allow us to monitor and capture system event logs during the
+target process.
+
+## Use Cases
+
+Generally, this predicate can be used with a runtime trace of any process.
+However, the primary use case driving the development of this predicate is its
+applicability to runtime traces of build processes.
```
marcelamelara, line 23: Several of the fields in the predicate seem to imply that the "process" your tracing is more specifically a CI/CD or some other scripted process, versus a discreet manually entered CLI command? I think it would be helpful to just add this extra small detail in the use cases (or even purpose) description to be crystal clear.

spec/predicates/runtime-trace.md
```diff
@@ -80,49 +105,68 @@ Identifies the specific monitor instance used to trace the runtime.
...
 
 Effectively a pointer to the monitor's configuration.
 
 `invocation.tracePolicy` _object_, _optional_
 
-Feedback required: Should tracingPolicy point outwards to a separate attestation with the same subject that wraps around the actual policy? The policy can be more securely defined and distributed (signed for by developer keys rather than online keys for example).
+Feedback required: Should tracingPolicy point outwards to a separate attestation
+with the same subject that wraps around the actual policy? The policy can be
+more securely defined and distributed (signed for by developer keys rather than
+online keys for example).
```
marcelamelara, line 135: It strikes me that the `tracePolicy` wouldn't actually change as often as the runtime traces are generated (correct me if I'm wrong), so I think it would definitely make sense to include a reference to a separate policy attestation that can be reused for any trace it applies to.

marcelamelara (commented):
Thanks for the updates! Left a few comments.

spec/predicates/runtime-trace.md
adityasaky: I think we can move all of invocation's fields into the monitor object. WDYT @pxp928?

spec/predicates/runtime-trace.md
adityasaky: Done!

spec/predicates/runtime-trace.md
adityasaky: I think I was in favour of having it be a separate document but I think either @pxp928 or @nadgowdas noted that we need to unambiguously map the trace policy to the trace so that we can reason about it when writing policies for `monitorLog`. For a separate document, I guess the monitor would have to understand that a configured trace policy matches the corresponding attestation? Perhaps a TODO past the v0.1 spec?

spec/predicates/runtime-trace.md
```diff
@@ -105,40 +106,37 @@ Identifies the specific monitor instance used to trace the runtime.
 
 URI indicating the monitor’s identity.
 
-`process` _object_, _required_
-
-`process.HostID` _string (URI)_, _required_
+`monitor.configSource` _object_, _optional_
 
-URI indicating the process host’s identity. Ex: GitLab Runner ID
+Effectively a pointer to the monitor's configuration.
```
marcelamelara, line 111: Should this be a ResourceDescriptor type? The `entryPoint` could be specified as an `annotation`.

spec/predicates/runtime-trace.md
```diff
@@ -18,9 +18,12 @@ target process.
 
 ## Use Cases
 
-Generally, this predicate can be used with a runtime trace of any process.
-However, the primary use case driving the development of this predicate is its
-applicability to runtime traces of build processes.
+Generally, this predicate can be used with a runtime trace of any process. The
+attestation specification does not mandate any specific monitoring tool or
```
marcelamelara, line 22: Do you mean the "predicate specification" doesn't mandate specific monitoring tools? I would also disambiguate the use of the term "process" wherever possible (in some places you probably mean "operation" rather an actual OS process).

spec/predicates/runtime-trace.md
```diff
@@ -105,40 +106,37 @@ Identifies the specific monitor instance used to trace the runtime.
...
-URI indicating the process host’s identity. Ex: GitLab Runner ID
+Effectively a pointer to the monitor's configuration.
 
-`process.type` _string (TypeURI)_, _required_
+`monitor.tracePolicy` _object_, _optional_
 
-URI indicating the type of process performed.
+Indicates the trace policy used for the monitoring event that generated the
+runtime trace.
+FIXME: should this be optional?
```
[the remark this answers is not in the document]
adityasaky: The example uses an object with one field which is a list. I'll update the description to say it's monitor specific and parsing it is based on the monitor ID.

Also, I added the fixme in the latest iteration because I think we need to discuss making the field required.

spec/predicates/runtime-trace.md
adityasaky: Should `monitoredProcess` be renamed?

spec/predicates/runtime-trace.md
adityasaky: Done! I've removed mention of `entryPoint` explicitly, I think that was a holdover from when this was based off SLSA Provenance.

spec/predicates/runtime-trace.md
adityasaky: Updated the text generally to use "operation" / "step" in several places instead of "process"

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2712 @@
...
+have different digests.
+
+While this predicate can be used to log file accesses, the actual technique used
+to capture the file access event has some implications. If a synchronous
+monitor, for example one that uses `ptrace` to trace the file access system
+calls, is used, then the build process can be paused while the file's digest is
+calculated and stored. However, asynchronous monitors such as those using eBPF
+cannot pause the build process before the file is actually used. Therefore, they
+cannot make as strong guarantees about the digests of the files accessed.
+Verifiers using runtime trace attestations for file accesses must careful about
```
itaysk, line 161: nit: must *be* careful

adityasaky: fixed, thanks!

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2712 @@
...
+While this predicate can be used to log file accesses, the actual technique used
+to capture the file access event has some implications. If a synchronous
+monitor, for example one that uses `ptrace` to trace the file access system
+calls, is used, then the build process can be paused while the file's digest is
+calculated and stored. However, asynchronous monitors such as those using eBPF
+cannot pause the build process before the file is actually used. Therefore, they
+cannot make as strong guarantees about the digests of the files accessed.
+Verifiers using runtime trace attestations for file accesses must be careful
+about what guarantees they are actually getting based on how the build process
+was monitored.
```
itaysk, line 163: IMO this is overly detailed and tailored for an assumed use case, while other elements of the monitorLog are left undefined.

marcelamelara (approved):
Thanks for all the updates! I think this is ready for v0.1.

spec/predicates/runtime-trace.md
adityasaky: Correct, we were worried about using file hashes recorded here to validate certain guarantees and wanted to be explicit that in some cases, the files may have changed between being accessed by the process and their hashes being recorded. Certainly erring on the side of caution, I'm open to hearing alternatives to this. Perhaps this could be a general specification-wide disclaimer as well about how attestations are captured and used. Until then, I'd prefer to leave this in here.

adityasaky:
Updated after @marcelamelara's approval to update predicate type to `https://in-toto.io/attestation/runtime-trace/v0.1` from `https://in-toto.io/attestation/runtime-trace/v0.1.0`.

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2712 @@
...
+
+`metadata.buildStartedOn` _object_, _optional_
+
+`metadata.buildFinishedOn` _object_, _optional_
+
+## Example
+
+```json
+{
+  "_type": "https://in-toto.io/Statement/v0.1",
```
TomHennen, line 177: nit: update to v1 ?

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2712 @@
...
+        ]
+      }
+    },
+    "monitoredProcess": {
+      "hostID": "https://tekton.dev/chains/v2",
+      "type": "https://tekton.dev/attestations/chains@v2",
+      "event": "run-image-pipelinerun-build-trusted"
+    },
+    "monitorLog": {
+      "process": [
```
TomHennen, line 210: optional: Would it make any sense to truncate this list?

TomHennen (changes requested):

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2712 @@
...
+-   Fields marked _optional_ MAY be unset or null, and should be treated
+    equivalently. Both are equivalent to empty for object or array values.
+
+## Fields
+
+`monitor` _object_, _required_
+
+Identifies the specific monitor instance used to trace the runtime.
+
+`monitor.id` _string (URI)_, _required_
```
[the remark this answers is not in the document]
itaysk: the parts that may vary by tool are the "monitorLog" contents. For file access, the authors of this spec defined very clearly how the data should look like so I don't see much wiggle room there. For process and network, you commented that perhaps we should leave them out untilI we can define them. I think we need to clarify the goals and non-goals of this spec better, do we want to dictate how the monitor should look like (in which case we normalize all differences between tools), or we leave it for the tool to decide (in which case we should revert the file access log specification). I also made this comment several other times

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2712 @@
...
+
+FIXME: should this be optional?
+
+`monitoredProcess` _object_, _required_
+
+Identifies the process being monitored.
+
+`monitoredProcess.hostID` _string (URI)_, _required_
+
+URI indicating the process host’s identity. Ex: GitLab Runner ID
```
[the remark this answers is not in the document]
adityasaky: I took the example out for now, I think we need to nail this down some more (and it looks like SLSA Provenance has a similar TODO for self hosted github actions runners).

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,2712 @@
...
+`monitoredProcess.event` _string_, _required_
+
+String identifying the specific job or task associated with the attestation.
+
+`monitorLog` _object_, _required_
+
+Record of events that were traced by the monitor. One of `process`, `network`,
+and `fileAccess` must be present.
+
+`monitorLog.process` _list_, _optional_
```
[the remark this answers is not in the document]
adityasaky: I think we should include them and indicate that their exact format is still being worked upon by cross referencing different monitor tools? Until then, it may be determined using the monitor ID? cc @pxp928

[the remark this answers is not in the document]
adityasaky: I've updated the text to reflect this, happy to change it further depending on what everyone thinks.

spec/predicates/runtime-trace.md
adityasaky: Done.

spec/predicates/runtime-trace.md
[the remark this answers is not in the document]
adityasaky: @itaysk I agree that the `process` / `network` and `fileAccess` are recording at slightly different levels of specificity but I also suspect that file accesses are easier than process and network objects to have a consistent, tool agnostic format for. For now, I've added text to `process` and `network` that it's monitor dependent and that a future version may settle on a consistent format all tools can emit to.

@TomHennen Do you mean making `monitor.id` a TypeURI? Also looping in @pxp928 and @nadgowdas here, should `monitor.id` identify the type of monitor or also a specific instance of Tetragon / tracee? Does the latter have real use cases?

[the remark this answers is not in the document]
TomHennen: > Do you mean making monitor.id a TypeURI? 

No, I think we could use a _separate_ field that identifies the type, though as @itaysk suggests, perhaps that should go in the monitorLog field?  Alternatively if the goal is that it _is_ well specified for all of them then it's not necessary.

Basically, if we're telling people they need to tweak how they process this data based on `monitor.id` then we probably need an explicit type somewhere.

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,274 @@
...
+`monitorLog.network` _list of objects_, _optional_
+
+Record of network activity observed by monitor. The exact format of this field
+is currently dependent on the monitor, and can be determined using `monitor.id`.
+In future, after consulting different monitors, this field may have a consistent
+schema.
+
+`monitorLog.fileAccess` _list of objects_, _optional_
+
+Record of files accessed during the monitored process. A complete list of
```
TomHennen, line 158: Is the idea that fileAccess is sufficiently well defined that all monitors can use it?  If so the various fields should probably be defined here?  Can also be TBD for v0.1

spec/predicates/runtime-trace.md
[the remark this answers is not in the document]
adityasaky: I've updated the `monitor.id` field to a `monitor.type` field. It's now a `TypeURI` and can be used to identify the type of monitor used. I've updated the example as well from `http://tetragon` to pointing to its repository + version. I think we can revisit multi-monitor scenarios (and if they're in scope) later.

spec/predicates/runtime-trace.md
adityasaky: Updated to be list of `ResourceDescriptor` objects.

adityasaky: I primarily foresee `name` and `digest` being used so I think this should suffice until a use case indicates otherwise.

TomHennen: That sounds good.  How would you feel about adding some text to that effect?

spec/predicates/runtime-trace.md
```diff
@@ -0,0 +1,274 @@
...
+
+Record of network activity observed by monitor. The exact format of this field
+is currently dependent on the monitor, and can be determined using
+`monitor.type`. In future, after consulting different monitors, this field may
+have a consistent schema.
+
+`monitorLog.fileAccess` _list of ResourceDescriptor objects_, _optional_
+
+Record of files accessed during the monitored process. A complete list of
+_materials_ can be derived from this information. Each entry in this list is
```
TomHennen, line 159: optional: Does materials here refer to SLSA provenance materials?  If so they're changing the terminology to resolvedDependencies

TomHennen:
LGTM for v0.1 (just a couple minor comments outstanding).

Thanks!

TomHennen (approved):

spec/predicates/runtime-trace.md
adityasaky: I was referring to both `materials` as used in https://github.com/in-toto/docs/blob/master/in-toto-spec.md#44-file-formats-namekeyid-prefixlink and as in the context of "SBOM". There's a use case here for deriving file level bills of materials for a build process and such.

spec/predicates/runtime-trace.md
adityasaky: I tweaked the text to explicitly call out ResourceDescriptor and the data that may be captured in an instance.

spec/predicates/runtime-trace.md
pxp928: yes that looks good. Thanks @adityasaky

spec/predicates/runtime-trace.md
pxp928: Using a separate document for the policy is a better solution in retrospect. I agree that it would not change that often but it would have to be published alongside the runtime attestation each time to ensure which policy was set when the trace was captured. If the policy for example is set to not capture any network logs, we cannot attest that during the build time, the pipeline did not reach out to the internet.

spec/predicates/runtime-trace.md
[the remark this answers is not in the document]
pxp928: Our goal was to be a URI that could identify the monitor being used. It might have gotten lost in translation. A TypeURI does the job here.

pxp928:
I had some outstanding comments that I never submitted. Thanks, @adityasaky for updating this :)

Issue #76, opened 2021-12-04
Add support for SCAI predicate


marcelamelara:
Per my discussions with @SantiagoTorres and @MarkLodato, I'm opening an issue to get this discussion going on a broader forum. 

CDI (or Code Deployment Integrity) is a framework for high-integrity provenance for software artifacts. CDI enables verifiers to make trust decisions about artifacts based on attested code _properties_, and enables verifiers to additionally establish trust in the attesters generating provenance metadata.  (See our [position paper](https://arxiv.org/abs/2106.09843))

In more detail, CDI can enhance in-toto in three key areas: 
1. Capture metadata about code security properties/behavior. 
CDI attestations contain not only information about a step in the supply chain, but also authenticated claims about specific security properties (e.g. claim: "this binary enforces strict memory bounds checks", evidence: builder is a verified Wasm AoT compiler). CDI also supports obtaining attestations from static/binary analysis tools. 

2. Integrity for the attesters: Tool endorsements 
In order to be able to substantiate claims about code properties inserted by a specific supply chain step, CDI obtains additional attestations or endorsements (signed claims) about attesters. Endorsements about the tools may come from a variety of endorsers such as CAs or third-party auditors, or from running tools on trusted execution environments (e.g., Intel SGX). These endorsements provide additional integrity metadata about the tools, reducing trust in the tools (i.e., attesters) and allowing verifiers to determine the confidence they have in attestations coming from certain attesters.  

3. Contextual deployment policies 
Attesting to code properties throughout the supply chain enables developers to specify fine-grained contextual policies about artifacts that can be enforced at deployment time. For instance, I may wish for a container running my binary to be deployed with co-tenants so long as they are side channel-free. This last feature requires post facto auditing, but I envision these policies being attached as part of the metadata for an artifact, and enforced by a verifier (e.g., a container orchestrator) at deployment time. I also envision eventually being able to specify such policies for supply chain steps (e.g., "build my artifact with an AES library that is attested to be side channel free"). 

A couple extra notes. Some of these features are more well fleshed out than others. I'm also currently working on a prototype to demonstrate running compilation inside of a TEE like SGX, but supporting other TEEs will be important as well.  

[...]

Issue #2, opened 2021-04-13
Compositional Notion On Attestation Types


SantiagoTorres:
It'd be nice to describe a strategy for different types of attestations within the same action.

Should we have these two types of attestations live under the same object? or would they be separate attestations attached to the same action. I.e., you could have 1 subject with multiple predicates. 

F.e., Having a provenance type that includes a measured boot record:

```json
{
  "subject": [
    { "name": "curl-7.72.0.tar.bz2",
      "digest": { "sha256": "ad91970864102a59765e20ce16216efc9d6ad381471f7accceceab7d905703ef" }},
  "predicateType": "https://in-toto.io/Provenance/v1",
  "predicate": {
    "builder": { "id": "https://github.com/Attestations/GitHubHostedActions@v1" },
    "recipe": {
      "type": "https://github.com/Attestations/GitHubActionsWorkflow@v1",
      "definedInMaterial": 0,
      "entryPoint": "build.yaml:maketgz"
    },
    "metadata": {
      "buildStartedOn": "2020-08-19T08:38:00Z"
    },
    "materials": [
      {
        "uri": "git+https://github.com/curl/curl-docker@master",
        "digest": { "sha1": "d6525c840a62b398424a78d792f457477135d0cf" },
        "mediaType": "application/vnd.git.commit",
        "tags": ["source"]
      }
    ]
    "tpm-measured-boot: {
       "PCR0": "xxxx",
       "PCR1": "yyyy",
       ...
    }
  }
}
```

This way, we would be able to know provenance information of the build + information about the host's integrity.

MarkLodato:
There are two simple, existing solutions:
- Issue two independent attestations.
- Use a single predicate with all the information you want.

This feature request is about cases where the above solutions won't work.

Theoretical use cases (if there are more, please add them):

- One attestation depends on another.

    - Example: PolicyDecision attestation depends on a prior Provenance attestation.
    - Possible solution: consider the dependent attestation an artifact like any other. (Challenge: requires stable serialization of the attestation, which is broken if the envelope is signed multiple times. Could alternatively refer to the Statement instead of the Envelope, but then that loses the attester, which is often critical.)

- Multiple predicates must be tied together but also must be able to be authenticated independently.

[...]

PR #112, opened 2022-10-13
Add support for the SCAI predicate

marcelamelara:
The Software Supply Chain Attribute Integrity, or SCAI (pronounced "sky") data format is designed to capture functional attribute and integrity information about software artifacts and their supply chain. SCAI data can be associated with executable binaries, statically- or dynamically-linked libraries, software packages, container images, software toolchains, and compute environments.

Why do we need this? Existing supply chain data formats do not capture any information about the security functionality or
*behavior* of the resulting software artifact, nor do they provide sufficient evidence to support any claims of integrity of the supply chain processes they describe. The SCAI data format is designed to bridge this gap.

This PR addresses #76 . An example use case for SCAI is described at: https://github.com/in-toto/attestation/issues/2#issuecomment-1104582285

### DIFF
--- a/spec/predicates/scai.md
+++ b/spec/predicates/scai.md
@@ -1,6 +1,6 @@
 # Predicate type: Software Supply Chain Attribute Integrity (SCAI)
 
-Type URI: scai/attribute-assertion/v0.1
+Type URI: scai/attribute-report/v0.1
 
 Version: 0.1
 
@@ -27,11 +27,12 @@ capturing other types of software supply chain metadata, which can be extended t
 
 ## Model
 
-SCAI metadata, referred to as **Attribute Assertions**, comprises a set of statements that describe functional
-attributes of a software artifact and its supply chain, capable of covering the full software stack of the
+SCAI metadata, referred to as an Attribute Assertion, describes functional attributes of a software
+artifact and its supply chain, capable of covering the full software stack of the
 toolchain that produced the artifact down to the hardware platform. SCAI Attribute Assertions include
 information about the conditions under which certain functional attributes arise, as well as (authenticated)
-evidence for the asserted attributes.
+evidence for the asserted attributes. The set of Assertions about a subject artifact and its producer
+is referred to as the **Attribute Report**.
 
 SCAI is intended to be implemented as part of an existing software supply chain attestation
 framework by software development tools or services (e.g., builders, CI/CD pipelines, software analysis tools)
@@ -41,11 +42,14 @@ objects.
 
 ## Schema
 
+The core metadata in SCAI is the Attribute Assertion. A collection of Attribute Assertions
+for a specific supply chain step or operation are issued together in a SCAI Attribute Report predicate.
+
 ```
 {
-    "predicateType": "scai/attribute-assertion/v0.1",
+    "predicateType": "scai/attribute-report/v0.1",
     "predicate": {
-        "attributes": [{
+        "subjectAttributes": [{
             "attribute": "<ATTRIBUTE>",
             "target": { // optional
                 "name": "<NAME>",
@@ -53,9 +57,21 @@ objects.
                 "locationURI": "<URI>",
                 "objectType": "<OBJECT TYPE>" // optional
             },
-            "conditions": { /*object */ }, // optional
+            "conditions": { /* object */ }, // optional
+            "evidence": { /* object */ } // optional
+        }],
+        "producerAttributes": [{
+            "attribute": "<ATTRIBUTE>",
+            "target": { // optional
+                "name": "<NAME>",
+                "digest": { "<ALGORITHM>": "<HEX VALUE>", ... },
+                "locationURI": "<URI>",
+                "objectType": "<OBJECT TYPE>" // optional
+            },
+            "conditions": { /* object */ }, // optional
             "evidence": { /* object */ } // optional
         }]
+            
     }
 }
 ```
@@ -64,7 +80,7 @@ This predicate has been adapted from the [SCAI v0.1 specification](https://arxiv
 
 ### Parsing Rules
 
-At a high-level, Attribute Assertions MUST allow humans and programs to easily parse the asserted
+At a high-level, Attribute Reports MUST allow humans and programs to easily parse the asserted
 attributes. Additional fields MUST enable program-based consumers to automatically parse and evaluate
 the given information.
 
@@ -78,30 +94,34 @@ The following parsing rules apply in addition:
 
 `predicateType` _string ([TypeURI](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#TypeURI)), required_
 
-> Identifier for the schema of the Assertion. Always
-> `scai/attribute-assertion/v0.1` for this version of the spec.
+> Identifier for the schema of the Report. Always
+> `scai/attribute-report/v0.1` for this version of the spec.
 
-`predicate.attributes` _array of objects, required_
+`predicate.subjectAttributes` _array of objects, required_
 
 > An array of one or more SCAI Attribute Assertions about the subject.
 
-`predicate.attributes[*].attribute` _string, required_
+`predicate.producerAttributes` _array of objects, optional_
 
-> A string describing a specific functional feature of the Attestation subject.
+> An array of one or more SCAI Attribute Assertions about the subject's producer.
+
+`predicate.*Attributes[*].attribute` _string, required_
+
+> A string describing a specific functional feature of the Attestation subject or producer.
 >
 > Attributes are expected to be domain- or application-specific.
 
-`predicate.attributes[*].target` _object ([SCAI Object Reference](https://arxiv.org/pdf/2210.05813.pdf), optional_
+`predicate.*Attributes[*].target` _object ([SCAI Object Reference](https://arxiv.org/pdf/2210.05813.pdf), optional_
 
 > An object reference to a specific artifact or metadata object to which the `attribute` field applies. 
 >
 > The semantics of the optional `objectType` field are up to the producer and consumer.
 
-`predicate.attributes[*].conditions` _object, optional_
+`predicate.*Attributes[*].conditions` _object, optional_
 
 > An object representing specific conditions under which the associated attribute arises.
 
-`predicate.attributes[*].evidence` _object, optional_
+`predicate.*Attributes[*].evidence` _object, optional_
 
 > An object representing (authenticated) evidence for the asserted `attribute`.
 >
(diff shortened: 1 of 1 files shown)


spec/predicates/scai.md
```diff
@@ -0,0 +1,271 @@
...
+            "attribute": "<ATTRIBUTE>",
+            "target": { // optional
+                "name": "<NAME>",
+                "digest": { "<ALGORITHM>": "<HEX VALUE>", ... },
+                "locationURI": "<URI>",
+                "objectType": "<OBJECT TYPE>" // optional
+            },
+            "conditions": { /*object */ }, // optional
+            "evidence": { /* object */ } // optional
+        }]
```
pxp928, line 58: Thinking through this, we would want to include timestamps if there is an attribute that is only valid for a period of time.

marcelamelara: So far, I've been thinking of attributes being quite static, but this is a really interesting point. Do you have a specific example in mind where an attribute might expire?

pxp928: I was thinking through this for GUAC and wanted to use SCAI to certify an artifact has been scanned (contains vulns or not). The attestation being valid for a period of time before it expires and needs to be re-ceritified (re-checked for vulnerabilities).

```
Certifier: The certifier creates attestations on the knowledge graph that helps translate queries about negative statements into queries on positive statements by certification. For example, instead of asking “Does artifact A have vulnerabilities?” to “Has artifact A been certified by a vulnerability scanner?”, the certifier takes in documents like OSV, VEX, bad actors and periodically (or via trigger), queries the graph to determine answers to these questions and does a certification in the knowledge graph.
```

pxp928: An example that I just made up and testing with:
```
{
  "_type": "https://in-toto.io/Statement/v0.1",
  "subject": [
      {
        "name": "git://github.com/kubernetes/kubernetes", 
        "digest": {"sha1": "5835544ca568b757a8ecae5c153f317e5736700e"}
      }
  ],
  "predicateType": "http://in-toto.io/attestation/scai/attribute-assertion/v0.1",
  "predicate": {
      "producer": {
        "type": "guac",
        "id": "guecsec/guac"
      },
      "attributes": [{
        "attribute": "scanned",
        "evidence": {
            "scanner": {
                "type": "OSV",
                "id": "osv.dev"
            },
            "results": [
              {
                "OSVID": "GHSA-jfh8-c2jp-5v3q"
              }]
        },
        "date": "2022-10-03T12:00:00Z"
      }]
  }
}
```

marcelamelara: I see, thanks for this example. This kind of certification process is definitely something we want to be able to support. So if I understood correctly, it's not necessarily that a certifier's attestation expires after X amount of time. Rather the problem is, how do we indicate that one attestation is no longer valid because it is replaced or superseded by a more recent one?

pxp928: Well the attestation does not need to indicate that its valid or not. That is up to the system/user to decide. Having a timestamp for when it was generated, allows GUAC or a user to determine when it was last certified as being vulnerable or not.

marcelamelara: Got it! Is the object in the `evidence field` something that is auto-generated by the scanner? I do wonder, since it's scan specific, if the `date` field couldn't be part of the evidence itself. But, I can also imagine there are other use cases where having a timestamp at the predicate-level could be useful, so it's definitely a field we could make optional in the SCAI predicate.

pxp928: Agreed. Updated the above example to move the date into evidence.

marcelamelara: Looks good, thanks! I'd definitely like to hear other's thoughts or use cases for having an optional date or timestamp field at the predicate level, too.

marcelamelara:
The latest commit incorporates feedback from everyone so far:
* Need for producer information (this is part of the SCAI spec, but was previously omitted for simplicity)
* Clearer examples showcasing the use of the `target` and `evidence` fields.

spec/predicates/scai.md
```diff
@@ -0,0 +1,334 @@
...
+            "target": { // optional
+                "name": "<NAME>",
+                "digest": { "<ALGORITHM>": "<HEX VALUE>", ... },
+                "locationURI": "<URI>",
+                "objectType": "<OBJECT TYPE>" // optional
+            },
+            "conditions": { /* object */ }, // optional
+            "evidence": { /* object */ } // optional
+        }],
+        "producerAttributes": [{
```
pxp928, line 63: Should we not also capture the type/URI or ID of the producer to determine the schema of the evidence object? Based on our last call, we decided to tie the schema of the evidence object to the producer so that it does not remain arbitrary.

spec/predicates/scai.md
```diff
@@ -0,0 +1,334 @@
...
+            "attribute": "<ATTRIBUTE>",
```
pxp928, line 64: The attribute field for the producer should be optional as it might be either the subject or producer or both.

spec/predicates/scai.md
TomHennen: Adding a timestamp makes sense to me.  The one tweak I'd suggest is being more specific about the name.  `date` could mean a number things, instead should it be named something like `scannedOn`, `collectedOn`, `generatedOn`.  This is [discussed a bit here](https://github.com/in-toto/attestation/tree/main/spec#predicate-conventions:~:text=Timestamps%20SHOULD%20use%20RFC%203339%20syntax%20with%20timezone%20%22Z%22%20and%20SHOULD%20clarify%20the%20meaning%20of%20the%20timestamp.%20For%20example%2C%20a%20field%20named%20timestamp%20is%20too%20ambiguous%3B%20a%20better%20name%20would%20be%20builtAt%20or%20allowedAt%20or%20scannedAt.).  Thoughts?

pxp928: Agreed. `scannedOn` makes more sense in this case and follows the convention.

marcelamelara: For a predicate-level field, I think something like `generatedOn` makes sense to keep it more generally applicable to non-scanning use cases.

spec/predicates/scai.md
marcelamelara: I think we discussed this a little on today's call. The evidence isn't actually necessarily tied to the producer of the SCAI predicate, as it may come from a third-party auditor/scanner etc. 

That said, for the producer, I do think it would be useful to have some sort of `identifier` field for the producer. The main reason I haven't added it in explicitly, is because I think different use cases will identify the producer differently. For some, the producer's public key may be enough, others may want a URI, as you said. I wasn't sure how to encode these different use cases in a general manner. Maybe it could be dependent on the producer's type?

spec/predicates/scai.md
marcelamelara: I'm not sure I understand what you mean. I had split the fields in the predicate by subject and producer already, but is there another scenario you have in mind?

spec/predicates/scai.md
marcelamelara: @pxp928 I wanted to ping you about this again as I try to come up with a way to capture a type and ID for the producer. The producer of the subject is not necessarily the producer of the evidence, so we shouldn't encode this expectation into the predicate. But are there other use cases for having a type or ID for a producer? And what would the format of these fields be? 

I have a few different ways to think of a producer of a SCAI predicate:
* The producer is a build system that is making assertions about the attributes of subject artifacts. In this case, the type might be the URI of the build service (e.g. GHA), and the ID the specific build job.
* The producer is the actual compiler that built the subject. In case, the ID might describe the gcc binary itself. But I might also argue that we could describe the compiler in the `target` field of the `producer.attributes` object instead. In any case, what does the type/URI point to for a binary?

These are only two cases, but I would want to be able to capture these and other cases. And, I'm still trying to understand what a consumer of the predicate would do with this information.

marcelamelara: Here are a few examples. I also saw your example above with GUAC as the producer. Is GUAC a tool, a service or something else in this case?

Producer is GHA:
```
{
     "_type": "https://in-toto.io/Statement/v0.1",
     "subject": [{
        "name": "my-app",
        "digest": { "sha256": "78ab6a8..." }
    }],
    "predicateType": "scai/attribute-report/v0.1",
    "predicate": {
        "subjectAttributes": [{
             "attribute": "WITH_STACK_PROTECTION",
             "conditions": { "build-flags": "-fstack-protector" }
        }],
        "producer": {
            "type": "https://url-to-gha",
            "id": "https://example-builder.com/user/repo/actions/runs/runID"
            "attributes": [{ 
                "attribute": "ATTESTED_BUILD",
                "evidence": {
                    "name": "my-app-slsa-provenance",
                    "digest": { "sha256": "4567890..." },
                    "locationURI": "http://example.com/rekor-instance",
                    "objectType": "application/vnd.in-toto+json"
                }
            }]
        }    
    }
}
```

Producer is tool:
```
{
     "_type": "https://in-toto.io/Statement/v0.1",
     "subject": [{
        "name": "my-app",
        "digest": { "sha256": "78ab6a8..." }
    }],
    "predicateType": "scai/attribute-report/v0.1",
    "predicate": {
        "subjectAttributes": [{
             "attribute": "WITH_STACK_PROTECTION",
             "conditions": { "build-flags": "-fstack-protector" }
        }],
        "producer": {
            "type": "ELF binary",
            "id": {
                "name": "gcc9.3.0",
                "digest": { "sha256": "78ab6a8..." },
                "locationURI": "http://us.archive.ubuntu.com/ubuntu/pool/main/g/gcc-defaults/gcc_9.3.0 1ubuntu2_amd64.deb"
            }
            "attributes": [{ 
                "attribute": "ATTESTED_BUILD",
                "evidence": {
                    "name": "my-app-slsa-provenance",
                    "digest": { "sha256": "4567890..." },
                    "locationURI": "http://example.com/rekor-instance",
                    "objectType": "application/vnd.in-toto+json"
                }
            }]
        }    
    }
}
```

pxp928: So my comment was tied to when we spoke that the evidence field schema would be tied to a producer. In the latter meeting, we discussed this would not be the case. I agree the evidence filed would be different based on the type of producer (tool, service...etc.). Maybe we hold off for v0.1 but what you have for now should be sufficient:
```
        "producer": {
            "type": "<TYPE URI>",
```

spec/predicates/scai.md
pxp928: Disregard. I was thinking about it wrong.

spec/predicates/scai.md
pxp928: We should add a `generatedOn` for this release and change it as necessary for later releases? What do you think?

pxp928 (approved):
LGTM for v0.1! Thanks @marcelamelara for the hard work on this.

mikhailswift (approved):

PR #115, opened 2022-10-19
Clarify bundle naming convention

### LINKED ISSUE #107: Need better clarity on bundle filename for chains of attestations
The current bundle [file naming convention](https://github.com/in-toto/attestation/blob/main/spec/bundle.md#file-naming-convention) is misleading for the case when the bundle is a set of dependent attestations that are all "about" a single artifact but have different subjects. Current wording:

> -   Bundles that concern a single artifact SHOULD name the bundle file `<artifact filename>.intoto.jsonl`.
> -   Bundles that concern multiple artifacts SHOULD name the bundle file `multiple.intoto.jsonl`.

The phrase "concern a single artifact" needs more specificity. It could be read two ways:

- (unintended) If any of the attestations in the bundle have a subject other than `<artifact>`, name it `multiple.intoto.jsonl`.
- (intended) If the set of attestations are intended to verify a single artifact, regardless of the attestations' subject, name it `<artifact filename>.intoto.jsonl`.

cc: @shaunmlowry @TomHennen @mlieberman85

MarkLodato:
Clarify the naming convention to be unambiguous when a single bundle "concerns" a single artifact but has attestations for other dependent artifacts.

Remove the recommendation for `multiple.intoto.jsonl`. In its absence, implementations can choose a name that makes sense for that use case.

Fixes #107.

### DIFF
--- a/spec/bundle.md
+++ b/spec/bundle.md
@@ -26,10 +26,14 @@ attestations.
 
 ## File naming convention
 
--   Bundles that concern a single artifact SHOULD name the bundle file
-    `<artifact filename>.intoto.jsonl`.
--   Bundles that concern multiple artifacts SHOULD name the bundle file
-    `multiple.intoto.jsonl`.
+Bundles SHOULD use the suffix `.intoto.jsonl`.
+
+A bundle of attestations relevant for `<filename>` SHOULD be named
+`<filename>.intoto.jsonl`. Attestations in the bundle MAY have different
+subjects, but they SHOULD all be relevant to that file or its dependencies. For
+example, a package named `foo-1.2.3.tar.gz` with hash `abcd` that was built from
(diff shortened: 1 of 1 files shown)


TomHennen (approved):
This looks good to me.  Might be good to get a second reviewer before merging.

[edit]: I guess I don't have write access (surprising!) so you need another reviewer anyways.

laurentsimon (approved):
LGTM

marcelamelara (approved):
LGTM, will merge.

marcelamelara:
@TomHennen Regarding write access, you don't seem to be part of the maintainers team yet (though you should be).

cc @adityasaky @SantiagoTorres

adityasaky:
@TomHennen, can you accept the invitation to the attestation maintainers team?

PR #125, opened 2023-01-25
Clarify policy engines in README.md

danbev:
This commit suggests updating the introduction section in README.md and replaces `in-toto` with `OPA` as an example of a policy engine.

The motivation for this change is that, based on my (limited) understanding of in-toto, I would not classify it as a policy engine.

Signed-off-by: Daniel Bevenius <daniel.bevenius@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This repository defines the in-toto **attestation** format, which represents
 authenticated metadata about a set of software artifacts. Attestations are
-intended for consumption by automated policy engines, such as [in-toto] and
+intended for consumption by automated policy engines, such as [OPA] and
 [Binary Authorization].
 
 ## IMPORTANT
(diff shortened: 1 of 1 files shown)


MarkLodato (commented):
LGTM but let's get @adityasaky's approval as a core in-toto maintainer.

marcelamelara:
Hello! Thanks for this PR, OPA definitely needs to be added to the list of policy engines that support in-toto. That said, in-toto does also provide a policy engine/policy verification via layouts and their validation; this should be clarified in our documentation, so thank you for pointing this out!

danbev:
> That said, in-toto does also provide a policy engine/policy verification via layouts and their validation; this should be clarified in our documentation, so thank you for pointing this out!

Ah, this is what I was missing. Thanks for clarifying this!

Issue #82, opened 2022-01-31
Add support for Cyclonedx as a predicate


samj1912:
intoto attestations currently document the [SPDX predicate type](https://github.com/in-toto/attestation/blob/main/spec/predicates/spdx.md). SBOMs and BOMs in general are a diverse space as of now and [CycloneDX](https://cyclonedx.org/) is the other leading industry alternative to SPDX for SBOMs.

CycloneDX supports other capabilities apart from just SBOMs. A particularly interesting one is the [VEX capability](https://cyclonedx.org/capabilities/vex/) which introduce a standard format to attach vulnerability information.

intoto should document and introduce well defined predicate types for the various CycloneDX BOM formats (not just SBOM).

cc: @stevespringett @coderpatros

samj1912:
> 
> # Predicate type: CycloneDX
> 
> Type URI: (tentative) https://cyclonedx.org/BOM/v1.4
> 
> Version: 1.0.0
> 
> TODO: Ask CycloneDX project to choose a URI and to review this spec. Ideally the URI would resolve to this file. Also, decide whether we want the version number to reflect the version (e.g. 1.4) or have them be independent (no version number in URI).
> 
> ## Purpose
> 
> A Bill of Materials type following the [CycloneDX standard](https://cyclonedx.org/specification/overview/).
> 
> This allows to represent an "exportable" or "published" software artifact, services, vulnerability information and more. For a complete list of capabilities see [CycloneDX Capabilities](https://cyclonedx.org/capabilities/). It can also be used as an entry point for other types of in-toto attestation when performing policy decisions.
> 
> ## Schema
> 
> ```jsonc
> {
>   // Standard attestation fields:
>   "_type": "https://in-toto.io/Statement/v0.1",
>   "subject": [{ ... }],
> 
>   // Predicate:
>   "predicateType": "https://cyclonedx.org/BOM/v1.4",
>   "predicate": {
>     "bomFormat": "CycloneDX",
>     "specVersion": "1.4",
>     "serialNumber": "urn:uuid:3e671687-395b-41f5-a30f-a58921a69b79",
>     "version": 1,
>     "components": [
>         {
>         "type": "library",
>         "name": "acme-library",
>         "version": "1.0.0"
>         }
>     ]
>     ...
>   }
> }
> ```
> 
[...]

PR #129, opened 2023-02-02
Add predicate specification for CycloneDX

danbev:
This commit adds a predicate specification for CycloneDX using the example specification from issue #82.

The motivation for doing this is only to offer help with getting this into in-toto, in case the original author currently does not have time.

Co-authored-by: samj1912 <sambhavs.email@gmail.com>
Resolves: https://github.com/in-toto/attestation/issues/82

Signed-off-by: Daniel Bevenius <daniel.bevenius@gmail.com>


marcelamelara:
Thanks for this PR @danbev! Could you please make sure to write up your proposed predicate in [ITE-9](https://github.com/in-toto/ITE/tree/master/ITE/9#document-format) formatting guidelines? 

A separate question, how large do you expect CycloneDX predicates to be in practice? We do want to consider scalability of attestations as well.

danbev:
> Could you please make sure to write up your proposed predicate in [ITE-9](https://github.com/in-toto/ITE/tree/master/ITE/9#document-format) formatting guidelines?

Sorry about that, for some reason I though it was in that format. I'll take a look at this next week and update :+1:

danbev:
> A separate question, how large do you expect CycloneDX predicates to be in practice? We do want to consider scalability of attestations as well.

I'm afraid I don't know the answer to this question as I have very little practical experience with CycloneDX.  Hopefully others can chime in that do have experience.

spec/predicates/cyclonedx.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Example
+```jsonc
+{
+  // Standard attestation fields:
+  "_type": "https://in-toto.io/Statement/v0.1",
+  "subject": [{ ... }],
+
+  // Predicate:
+  "predicateType": "https://cyclonedx.org/BOM",
```
pxp928, line 41: CDX schema version should be part of the `predicateType`

pxp928 (commented):
This should be good to go. It does reference the most up-to-date [CDX schema](https://github.com/CycloneDX/specification/tree/1.4/schema) as the predicate

spec/predicates/cyclonedx.md
```diff
@@ -0,0 +1,71 @@
...
+The schema of this predicate type is documented in the
+[CycloneDX Specification].
+
+### Parsing Rules
+The parsing rules for this predicate type are documented in the
+[CycloneDX Specification].
+
+### Fields
+The fields that make up this predicate type are documented in the
+[CycloneDX Specification].
```
marcelamelara, line 31: I think it would be very helpful for implementers to not have to rely only on the CycloneDX specification document. Could you please provide a list of minimum required or expected fields, and a short description of their purpose? You may then still point to the overall CycloneDX spec for more details.

spec/predicates/cyclonedx.md
```diff
@@ -0,0 +1,71 @@
+# Predicate type: CycloneDX
+
+Type URI: (tentative) https://cyclonedx.org/bom
+
+Version: 1.0.0
```
marcelamelara, line 5: If this predicate is meant to comply with the latest CycloneDX spec version, this should probably be 1.4

spec/predicates/cyclonedx.md
```diff
@@ -0,0 +1,71 @@
...
+## Purpose
+
+A Software Bill of Materials type following the [CycloneDX standard].
+
+This allows representing "exportable", or "published" software artifacts,
+services, vulnerability information, and more. For a complete list of
+capabilities see [CycloneDX Capabilities].
+
+## Prerequisites
+The in-toto [attestation] framework.
```
marcelamelara, line 16: This predicate also needs an SBOM generation tool.

spec/predicates/cyclonedx.md
marcelamelara, line 41: This should match https://cyclonedx.org/specification/overview/#recognized-predicate-type

```suggestion
  "predicateType": "https://cyclonedx.org/bom/v1.4",
```

marcelamelara: +1

TomHennen: Is it supposed to _just_ be `https://cyclonedx.org/bom`?  The page doesn't include the version (which I assume is embedded in the `predicate` field as it is for SPDX?

marcelamelara: As stated below, the `specVersion` field is required by the CDX spec, which is why the version is not in the `predicateType`. See: https://cyclonedx.org/docs/1.4/json/#bomFormat

spec/predicates/cyclonedx.md
```diff
@@ -0,0 +1,71 @@
...
+    ...
+  }
+}
+```
+The `predicate` contains a JSON-encoded CycloneDX BOM. The CycloneDX format has
+a mandatory `specVersion` field, so we may choose to omit the version number
+from the predicateType URI to avoid confusion.
+
+The `subject` contains whatever software artifacts are to be associated with
+this CycloneDX BOM document.
```
marcelamelara, line 63: This should go under the Fields section.

spec/predicates/cyclonedx.md
marcelamelara, line 16: This predicate also needs a CycloneDX BOM generation tool.

spec/predicates/cyclonedx.md
marcelamelara, line 63: These descriptions should go under the Fields section.

spec/predicates/cyclonedx.md
pxp928: I still think it should be part of the type such that we don't have to go down to the predicate to understand the `specVersion`. I am thinking about this in terms of tooling that might process these in the future.

marcelamelara: That's totally fair. Probably shouldn't assume that all tools will know how to process CDX predicates.

danbev: Apologies for my late reply here!

So would something like this work:
```
"predicateType": "https://cyclonedx.org/BOM/v1.4"
```

spec/predicates/cyclonedx.md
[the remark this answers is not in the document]
danbev: Yes, I'll update this. Thanks~

spec/predicates/cyclonedx.md
danbev: I'll add this. Thanks!

spec/predicates/cyclonedx.md
danbev: I'll move these. Thanks

marcelamelara (approved):
Thank you for the changes, @danbev ! 

One more thing: Could you please add a link to the CycloneDX spec in the [predicates list](https://github.com/in-toto/attestation/blob/main/spec/predicates/README.md#vetted-predicates)?

spec/predicates/README.md
```diff
@@ -21,6 +21,7 @@ our [vetting process], and may be of general interest:
     operations.
 -   [SLSA Verification Summary]: SLSA verification decision about a software
     artifact.
+-   [CycloneDX]: CycloneDX BOM for software artifacts.
```
TomHennen, line 24: You'll need to add a link to [CycloneDX] below or else this will just show up as `[CycloneDX]` instead of as a real link.

spec/predicates/cyclonedx.md
```diff
@@ -0,0 +1,77 @@
+# Predicate type: CycloneDX
+
+Type URI: (tentative) https://cyclonedx.org/bom
```
TomHennen, line 3: nit: remove '(tentative)' ?

TomHennen (changes requested):
Thanks, generally looks good, just a couple small comments.

spec/predicates/README.md
danbev: Sorry about that, I'll add it now. Thanks!

spec/predicates/cyclonedx.md
danbev: Thanks, I'll remove this.

TomHennen (approved):

pxp928 (approved):
LGTM! Thanks for the changes

PR #132, opened 2023-02-06
Set up v1 draft folder

TomHennen:
Sets up a v1.0-draft folder to make changes related to v1.

These are copies of the corresponding files in spec/ with the only change made being the version number bump.


marcelamelara:
@TomHennen Yes! So I wanted to figure out what the best order of operations is between this merging PR and #132. I think we should merge yours first, and I'll rebase on that then. Thoughts?

marcelamelara (approved):
SGTM. Should we create a corresponding v0.1 folder too? Or not until we release v1.0?

TomHennen:
> SGTM. Should we create a corresponding v0.1 folder too? Or not until we release v1.0?

Yeah I figured we'd do that when we publish 1.0 (basically moving v1.0-draft/ to v1.0/)  Make sense?

marcelamelara:
> > SGTM. Should we create a corresponding v0.1 folder too? Or not until we release v1.0?
> 
> Yeah I figured we'd do that when we publish 1.0 (basically moving v1.0-draft/ to v1.0/) Make sense?

Yes, and 0.1.0 should probably also get its own folder then.

pxp928 (approved):
So these are just copies for now?

Edit: it says so above so LGTM!

Issue #138, opened 2023-02-22
Ignoring unrecognised fields?


joshuagl:
SLSA's summary of the in-toto attestations parsing tools is being update to suggest that unrecognised fields are ignored "unless otherwise noted" https://github.com/slsa-framework/slsa/pull/628

Should we update the [parsing rules](https://github.com/in-toto/attestation/tree/main/spec#parsing-rules) to match?

trishankatdatadog:
Yes

PR #139, opened 2023-02-23
Update parsing rules for unrecognized fields

joshuagl:
There are cases where unrecognised fields SHOULD reject unrecognised fields, rather than ignoring them. This is true for the externaleParameters field in SLSA provenance v1 and the summary of the parsing rules included in that specification were recently updated to clarify. https://github.com/slsa-framework/slsa/pull/628

This change updates the guidance here to match.

Fixes: #138


MarkLodato (approved):

TomHennen (approved):

pxp928 (approved):
LGTM

Issue #130, opened 2023-02-03
Create Statement v1


TomHennen:
SLSA is working on publishing SLSA 1.0.  Since the SLSA provenance spec uses in-toto statements it might be nice if they had a v1 of statements to use in their provenance spec.

The statement definition has been pretty stable.  Should we rev it to v1?

The one change that we might consider is adding an 'attestor id' field to the statement layer.  Rationale: [SLSA provenance](https://slsa.dev/provenance/v0.2#builder.id), [VSAs](https://slsa.dev/verification_summary/v0.2#verifier.id), [proposed runtime trace](https://github.com/in-toto/attestation/pull/111/files#diff-c1a4f7ac32ae7837cb9a3fb9e656a84d214432d3913a78453b21adb8fce526d9R81) [proposed reference attestations](https://github.com/in-toto/attestation/issues/117#issuecomment-1333815831) all have some sort of ID in the predicate field.  It might be helpful to up-level this to the statement so it can be standardized.  If folks don't like this idea we can drop it.

[EDIT] Blockers:
- [X] Get resolution on #133 @marcelamelara 
~~- [ ] Decide adding an 'attestor id' field to the statement layer (per above).~~
- [x] Resolve #135
- [x] Clarify DigestSet terminology [per this comment](https://github.com/in-toto/attestation/issues/130#issuecomment-1416022258) and [this comment](https://github.com/in-toto/attestation/pull/135#discussion_r1114675451) (probably blocked on #133?) @TomHennen 
- [x] Address #127 @joshuagl 
- [x] Resolve #131 
- [x] Determine inclusion (and resolution?) of #137 
- [x] Remove .0 from version number per [this comment](https://github.com/in-toto/attestation/issues/130#issuecomment-1435317334) @pxp928 
- [x] Resolve #143 @marcelamelara 
- [x] Remove "-draft" from all files
- [x] Resolve #157 @TomHennen & @pxp928

MarkLodato:
Another topic that has come up is the encoding of digests, e.g. #108. Currently the spec says:

> Set of alternative cryptographic digests [...] expressed as a JSON map from algorithm name to **lowercase hex-encoded value**.

There is some desire to either:

- Allow alternative encodings, such as base64 for `goModuleH1` since base64 is more natural for go (it's what other systems use).
- Switch them all to base64.
  - Advantages:
    - Smaller encoding
[...]

PR #135, opened 2023-02-07
spec v1.0-draft documentation updates

marcelamelara:
This PR makes a number of (simplifying) updates to documentation for the v1.0-draft spec release.

Open questions:
* Do we want to do a separate v1.0-draft release before v1.0?
* Should we place the definitions for envelope, statement and predicate in separate files? Currently they're all in one place.
* Should the documentation cleanup (#127) be blocking for v1.0-draft release?

PR #140, opened 2023-02-27
Specify lowercase-hex encoding for standard algorithms only.

TomHennen:
There are cases where custom digests might be better expressed using different encodings.

refs #130

### DIFF
--- a/spec/v1.0-draft/digest_set.md
+++ b/spec/v1.0-draft/digest_set.md
@@ -53,7 +53,7 @@ Supported algorithms:
 > Two DigestSets SHOULD be considered matching if ANY acceptable field
 > matches.
 >
-> New algorithms must document how the value is encoded. E.g. base64,
+> New algorithms MUST document how the value is encoded, e.g. URL-safe base64,
 > lowercase hex, etc...
 
 ## Examples
--- a/spec/v1.0-draft/digest_set.md
+++ b/spec/v1.0-draft/digest_set.md
@@ -26,7 +26,7 @@ agility.
 Supported algorithms:
 
 -   Standard cryptographic hash algorithms using [the NIST names][]
-    (converting to lowercase and replacing `-` replaced with `_`) as keys
+    (converting to lowercase and replacing `-` with `_`) as keys
     and lowercase hex-encoded values, for cases when the method of
     serialization is obvious or well known:
     `sha256`, `sha224`, `sha384`, `sha512`, `sha512_224`, `sha512_256`,
--- a/spec/v1.0-draft/digest_set.md
+++ b/spec/v1.0-draft/digest_set.md
@@ -25,14 +25,15 @@ agility.
 
 Supported algorithms:
 
--   Standard cryptographic hash algorithms with lowercase hex-encoded
-    values, for cases when the method of serialization is obvious or
-    well known:
+-   Standard cryptographic hash algorithms using [the NIST names][nist-hashes]
+    (converting to lowercase and replacing `-` replaced with `_`) as keys
+    and lowercase hex-encoded values, for cases when the method of
+    serialization is obvious or well known:
     `sha256`, `sha224`, `sha384`, `sha512`, `sha512_224`, `sha512_256`,
     `sha3_224`, `sha3_256`, `sha3_384`, `sha3_512`, `shake128`, `shake256`,
     `blake2b`, `blake2s`, `ripemd160`, `sm3`, `gost`, `sha1`, `md5`
 
--   `goModuleH1`: The go module [directory Hash1][], omitting the "h1:"
+-   `goModuleH1`: The go module [directory Hash1][Hash1], omitting the "h1:"
     prefix and output in lowercase hexadecimal instead of base64. Can
     be computed over a directory named `name@version`, or the contents
     of zip file containing such a directory:
@@ -64,4 +65,5 @@ Supported algorithms:
 -   `{"sha256": "abcd"}` does not match `{"sha256": "fedb", "sha512": "abcd"}`
 -   `{"somecoolhash": "abcd"}` uses a non-predefined algorithm
 
-[directory Hash1]: https://cs.opensource.google/go/x/mod/+/refs/tags/v0.5.0:sumdb/dirhash/hash.go
+[nist-hashes]: https://csrc.nist.gov/projects/hash-functions
+[Hash1]: https://cs.opensource.google/go/x/mod/+/refs/tags/v0.5.0:sumdb/dirhash/hash.go
--- a/spec/v1.0-draft/digest_set.md
+++ b/spec/v1.0-draft/digest_set.md
@@ -18,8 +18,10 @@ metadata object.
 ## Fields
 
 A DigestSet is represented as a _JSON object_ mapping algorithm name to
-a string value. Usually there is just a single key/value pair, but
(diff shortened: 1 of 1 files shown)


TomHennen:
@MarkLodato I'd be interested in your thoughts here.

spec/v1.0-draft/digest_set.md
```diff
@@ -53,6 +52,9 @@ Supported algorithms:
 >
 > Two DigestSets SHOULD be considered matching if ANY acceptable field
 > matches.
+>
+> New algorithms must document how the value is encoded. E.g. base64,
```
MarkLodato, line 56:
```suggestion
> New algorithms MUST document how the value is encoded, e.g. URL-safe base64,
```

spec/v1.0-draft/digest_set.md
```diff
@@ -18,23 +18,22 @@ metadata object.
 ## Fields
 
 A DigestSet is represented as a _JSON object_ mapping algorithm name to
-lowercase hex-encoded value. Usually there is just a single key/value pair,
-but multiple entries MAY be used for algorithm agility.
-
-> **TODO**: Add language for supporting base64-encoded digest values (re: goModuleH1).
+a string value. Usually there is just a single key/value pair, but
```
MarkLodato, line 21: Might help to be more explicit about encoding, e.g.

```suggestion
a string encoding of the digest using that algorithm. The named standard algorithms below use lowercase hex encoding. Usually there is just a single key/value pair, but
```

(I didn't wrap this properly)

MarkLodato (commented):
Sounds good to me. A few suggestions

spec/v1.0-draft/digest_set.md
```diff
@@ -18,23 +18,22 @@ metadata object.
...
+    well known:
     `sha256`, `sha224`, `sha384`, `sha512`, `sha512_224`, `sha512_256`,
     `sha3_224`, `sha3_256`, `sha3_384`, `sha3_512`, `shake128`, `shake256`,
     `blake2b`, `blake2s`, `ripemd160`, `sm3`, `gost`, `sha1`, `md5`
 
 -   `goModuleH1`: The go module [directory Hash1][], omitting the "h1:"
-    prefix and output in hexadecimal instead of base64. Can be computed
-    over a directory named `name@version`, or the contents of zip file
-    containing such a directory:
+    prefix and output in lowercase hexadecimal instead of base64. Can
```
marcelamelara, line 34: Since the `Hash1` function outputs a base64-encoded string, should we allow base64 encoding for this algorithm?

spec/v1.0-draft/digest_set.md
marcelamelara, line 56: Just an idea (and maybe this has been discussed already): Do we see a downside to adding an optional `encoding` field to the DigestSet type?

spec/v1.0-draft/digest_set.md
TomHennen: That would make sense, I wonder if it would have any backwards compatibility issues?  E.g. we'd allow EITHER lowercase-hex or base64 and users would need to support both.

@MarkLodato WDYT?

spec/v1.0-draft/digest_set.md
TomHennen: It's a good thing to discuss.  My first concern is that it would probably be a breaking change?  This would go from map<string, string> to map<string, ValueEntry> where ValueEntry has (encoding, value).  or is there a more clever way to do it?

spec/v1.0-draft/digest_set.md
TomHennen: Done

spec/v1.0-draft/digest_set.md
MarkLodato: That seems like it makes the producer's job ever-so-slightly easier but the consumer's job much harder, especially if most of the time they receive one encoding but then they suddenly get a different one. I'm inclined to pick one.

Between base64 and hex, I don't feel too strongly.

TomHennen: I think the best approach might be

1. Leave goModuleH1 as lowercase hex so that existing use cases can stay the same and we can unblock 1.0 with minimal risk.
2. In the future, if needed, add goModuleH1+base64 (per @marcelamelara's suggestion) so that people can specify _both_ goModuleH1 (lowercase hex) and goModuleH1+base64.

spec/v1.0-draft/digest_set.md
TomHennen: Right, I think people will generally be using the key field as a opaque string, so adding +base64 here just turns it into a new string type (which is still useful!).  Folks will be able to use both goModuleH1 and goModuleH1+base64 if they want.  Maybe in the future we could add fancy support for specifying the encoding in here, but then people would need to handle them as non-opaque strings, which would be a big change.

spec/v1.0-draft/digest_set.md
marcelamelara: Might be good to add a comment somewhere in the docs, or open an issue for this, just so we don't forget.

trishankatdatadog: Agree with Mark: generally, for security reasons (e.g., for the same reasons DSSE was formulated), best to decode using one expected format

joshuagl (approved):

spec/v1.0-draft/digest_set.md
```diff
@@ -18,23 +18,24 @@ metadata object.
...
+single key/value pair, but multiple entries MAY be used for algorithm
+agility.
 
 Supported algorithms:
 
--   Standard cryptographic hash algorithms, for cases when the method
-    of serialization is obvious or well known:
+-   Standard cryptographic hash algorithms with lowercase hex-encoded
+    values, for cases when the method of serialization is obvious or
+    well known:
```
marcelamelara, line 30: One more thing. Might be good to clarify the naming of the algorithms per https://github.com/in-toto/attestation/pull/135#discussion_r1114675451

spec/v1.0-draft/digest_set.md
TomHennen: Ok, I'll leave it as is, without a note until we actually need something else.

Thanks everyone!

spec/v1.0-draft/digest_set.md
TomHennen: Done!

TomHennen:
I think we're good now, but I'll let this sit for a day before merging.

spec/v1.0-draft/digest_set.md
```diff
@@ -18,23 +18,25 @@ metadata object.
...
+algorithms below use lowercase hex encoding. Usually there is just a
+single key/value pair, but multiple entries MAY be used for algorithm
+agility.
 
 Supported algorithms:
 
--   Standard cryptographic hash algorithms, for cases when the method
-    of serialization is obvious or well known:
+-   Standard cryptographic hash algorithms using [the NIST names][]
+    (converting to lowercase and replacing `-` replaced with `_`) as keys
```
MarkLodato, line 29:
```suggestion
    (converting to lowercase and replacing `-` with `_`) as keys
```

MarkLodato (approved):

trishankatdatadog (approved):

TomHennen:
Thanks everyone!

PR #141, opened 2023-03-02
Rename `goModuleH1` to `dirHash1` and add better explanation

MarkLodato:
Previously the spec was written specifically for go modules, which always have a single top-level directory entry of the form `name@version`. However, this function is not specific to go and does not need to have a restriction that there is a single `name@version` top-level directory.

Now the spec explains that it is designed but not specifically for go modules, and walks through a detailed example of how to compute. It also better explains how to run the "equivalent" command.

### Question for reviewers

Is this clear? Is it a good idea? I was surprised re-reading the spec that it requires `name@version` since that likely won't apply to other package ecosystems.

### DIFF
--- a/spec/v1.0-draft/digest_set.md
+++ b/spec/v1.0-draft/digest_set.md
@@ -33,18 +33,68 @@ Supported algorithms:
     `sha3_224`, `sha3_256`, `sha3_384`, `sha3_512`, `shake128`, `shake256`,
     `blake2b`, `blake2s`, `ripemd160`, `sm3`, `gost`, `sha1`, `md5`
 
--   `goModuleH1`: The go module [directory Hash1][], omitting the "h1:"
-    prefix and output in lowercase hexadecimal instead of base64. Can
-    be computed over a directory named `name@version`, or the contents
-    of zip file containing such a directory:
+-   `goModuleH1`: The [directory Hash1][] function, omitting the "h1:" prefix
+    and output in lowercase hexadecimal instead of base64. This algorithm was
+    designed for go modules but can be used to digest the _contents_ of an
+    arbitrary archive or file tree. Equivalent to extracting the archive to an
+    empty directory and running the following command in that directory:
 
     ```bash
-    find name@version -type f | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
+    find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
     ```
 
     For example, the module dirhash
-    `h1:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=` would be encoded as
-    `{"goModuleH1": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"}`.
+    `h1:Khu2En+0gcYPZ2kuIihfswbzxv/mIHXgzPZ018Oty48=` would be encoded as
+    `{"goModuleH1": "2a1bb6127fb481c60f67692e22285fb306f3c6ffe62075e0ccf674d7c3adcb8f"}`.
+
+    <details>
+    <summary>Detailed example</summary>
+
+    The go module `github.com/marklodato/go-hello-world@v0.0.1` has module
+    dirhash `h1:Khu2En+0gcYPZ2kuIihfswbzxv/mIHXgzPZ018Oty48=`:
+
+    ```bash
+    $ curl https://sum.golang.org/lookup/github.com/marklodato/go-hello-world@v0.0.1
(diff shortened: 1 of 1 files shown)


TomHennen:
I guess I didn't read the spec as requiring name@version, instead I read it as just an example of how you _could_ compute it.  Presumably folks would be allowed to compute this value in a library and not in a command line invocation?

MarkLodato:
> I guess I didn't read the spec as requiring name@version, instead I read it as just an example of how you _could_ compute it. Presumably folks would be allowed to compute this value in a library and not in a command line invocation?

Yes, it was a bit ambiguous before. It wasn't clear to me, upon re-reading it, whether it was allowed to be something other than `name@version`, or if you're allowed to have more than one top-level entry.

Agreed that you can compute it with any implementation. The new wording says "Equivalent to" to hopefully make this more clear?

I'm not sure how to interpret your comment. Are you suggesting that the previous wording was better?

TomHennen:
Oh yeah, I think the previous wording probably was better (because it was simpler), but it can probably be _improved_ by adding 'equivalent to' or something like that to make it clear folks don't have to use this method?  That way you don't have to go out of your way to show how it can be done generally?

(but I don't feel that strongly about it)

MarkLodato:
Before I try rewriting, I want to validate some assumptions first.

Overall my goal was to make it more clear that this is not specific to go modules. Here are the concerns I had with the old text:

- The first words are "The go module directory Hash1", implying it's just for go modules.
- The text and command say that it "be computed over a directory named `name@version`", which is overly restrictive.
- Similarly "contents of zip file" rather than any arbitrary archive.
- (minor) It doesn't explain that it is computing the tree hash over the names and contents of a set of files, but I'm OK omitting that.

Do you think these are worth clarifying? If so, do you have an idea on how to make it more clear? The reason I originally wrote `name@version` is because it removes all the difficulty around specifying that you need to be in the parent directory of that file and give a slash-less path to the top-level files. Alternatively I can reference https://gist.github.com/MarkLodato/c03659d242ea214ef3588f29b582be70 (or maybe make it a module, or check it in to this repo) but that's even more complex.

spec/v1.0-draft/digest_set.md
```diff
@@ -33,19 +33,26 @@ Supported algorithms:
     `sha3_224`, `sha3_256`, `sha3_384`, `sha3_512`, `shake128`, `shake256`,
     `blake2b`, `blake2s`, `ripemd160`, `sm3`, `gost`, `sha1`, `md5`
 
--   `goModuleH1`: The go module [directory Hash1][], omitting the "h1:"
-    prefix and output in lowercase hexadecimal instead of base64. Can
-    be computed over a directory named `name@version`, or the contents
-    of zip file containing such a directory:
+-   `goModuleH1`: The [directory Hash1][] function, omitting the "h1:" prefix
+    and output in lowercase hexadecimal instead of base64. This compute the
```
trishankatdatadog, line 37:
```suggestion
    and output in lowercase hexadecimal instead of base64. This computes the
```

trishankatdatadog:
Makes sense to me, but would be best to explicitly try it on packages from other languages (ideally, one each from Python, Ruby, NPM, Crates, and so on).

TomHennen:
I think I see.  There's a desire to make this hash type work for _explicitly_ for things beyond go modules.  I had read it more as "we don't want to forbid people from doing so".

So if that's the case, then I think the existing language is fine, but perhaps add something about how this might be called **go**ModuleH1 it can be used for any type of directory content hashing.  The note at the bottom already covers that, perhaps it can just be strengthened?

MarkLodato:
OK, I rewrote it to hopefully make this more clear without adding complexity. I also fixed the command to be more accurate and add a worked example to show exactly what is going on. What do you think?

Another question is, should we rename to `dirHash1` since it's not specific to go?

TomHennen (approved):
Looks good!

TomHennen:
> Another question is, should we rename to dirHash1 since it's not specific to go?

Up to you.  I don't know if anyone is using or planning to use this yet.  Not sure if it's worth upsetting any existing implementations.

MarkLodato:
I'll leave it as-is for now to keep this PR small, but we might want to consider renaming it before it really gets used.

TomHennen:
> I'll leave it as-is for now to keep this PR small, but we might want to consider renaming it before it really gets used.

That makes sense to me.  I also perhaps shouldn't be too worried about existing implementations and backwards compatibility, they'll still work.  Renaming SGTM.

MarkLodato:
Thanks. I need another approver.

If I should just rename in this PR for convenience, let me know.

joshuagl (approved):
LGTM. I appreciate the detailed example, thanks.

> If I should just rename in this PR for convenience, let me know.

Please do.

Issue #117, opened 2022-12-01
Add generic object/artifact reference attestation field type


marcelamelara:
For attestation predicates that reference other attestations or artifacts, such as the `evidence` fields in SCAI (#112) or SLSA v1.0, it's beneficial for scalability to provide a reference to the object in question, rather than including the full object. The SCAI specification proposes an Object Reference format (see [SCAI spec doc](https://arxiv.org/abs/2210.05813), Sec. 3), which can serve as the foundation for a generic object reference type:

```
{ 
    "name": "<NAME>",
     "digest": { "<ALGORITHM>": "<HEX VALUE>", ... },
     "locationURI": "<URI>",
     "objectType": "<OBJECT TYPE>" // optional
}
```

This object follows a common pattern as seen in the (extended) attestation `subject` field (#5) or the artifact reference in SLSA v1.0 [Artifact Reference](https://github.com/slsa-framework/slsa/blob/e60f7426f501f34e8b41e9b4da594bb7ec1c57bb/docs/provenance/v1.0.cue#L37), so there is precedent to formalize such a field type in attestations.

Examples for how this field type might be used can be found as part of the [SCAI predicate spec](https://github.com/in-toto/attestation/blob/main/spec/predicates/scai.md#examples)

marcelamelara:
Nice! I like the idea of having the MIME type in there, instead of the less descriptive `objectType` I have in there right now. 

I hadn't considered `References` as predicates, but SBOM seems like a great use case for this. I had been imagining that even fields like `subject` may become of type `Reference`, but this is something we could discuss.

TomHennen:
Oh I guess that does seem like a different use case.  For our purposes we just need to communicate that this document exists and don't necessarily need to reference it from within some pre-existing in-toto statement.

marcelamelara:
So, in essence, the in-toto attestation then really just becomes a signature over the document without having to sign the entire document directly. I would expect this to be a pretty common use case in practice, so I think we definitely want to support this.

marcelamelara:
This issue may be relevant to #27 and #2 as well.

TomHennen:
Yes I think so, but I wonder what the location field would be if the attestation is contained within the bundle but not necessarily at any specific URI?  Just unset?  Some special value like "<in-bundle>"?

[...]

Issue #27, opened 2021-04-28
Ability to refer to one attestation from another


MarkLodato:
There are likely use cases where you want to refer to one attestation from another. For example, you could have a "Policy Decision" predicate that says "I allow subject x to run in environment Y, based on input attestation Z."

Since this is just theoretical at the moment, we'll wait until we have a few concrete use cases to actually design and implement this. Please add any use cases you might have to this issue.

The straightforward solution is to treat the entire attestation as an artifact, and thus you refer to the attestation as a hash over the envelope. The downside to this approach is that it prevents one from re-encoding the envelope, such as if you add a signature to it, because doing so changes the hash. (Counter-point: don't do that.) Perhaps this is the best option.

Alternative ideas:
- Refer to the hash of the statement, rather than the envelope. Downside: that loses who signed it, which is critical information.
- Refer to the hash of some canonicalization of the envelope. Downside: that relies on canonicalization (frowned upon, see ITE-5), needs to be invented, and adds complexity.
- Each attestation has a UUID. Downside: adds complexity and could be error prone.  For example, what if someone else creates one with the same UUID, either accidentally or on purpose?
- Store attestations in some ledger, then you can refer to the location within the ledger (transaction ID or leaf hash). Downside: requires a ledger, and is seemingly not any better than the straightforward solution.

JustinCappos:
Can you do this with sublayouts and/or treating the result of an in-toto verification as a step?  If not, can you give an example to make it clear why?

MarkLodato:
@JustinCappos To clarify, this is just tracking the general idea. We don't have concrete use cases yet, so we can't really say whether any particular solution would work. I updated the initial comment to clarify.

That said, I don't think either of those are solutions here. Layouts are policies, and this is about attestations (i.e. links). The result of an in-toto verification is itself an attestation/link, so that has the exact same problem. I'm guessing that in-toto currently solves this problem by referring to the hash of the entirely link? I've now listed as the "straightforward solution" in the first post. Maybe I'm over-thinking this.

marcelamelara:
@MarkLodato can you please take a look at the structure proposed in #117? Might this be a possible step in the right direction to support the functionality you're looking for?

MarkLodato:
#117 seems reasonable to me.

To be clear, this issue was about how we recommend that one attestation refers to another. In the original post, I said: 

> The straightforward solution is to treat the entire attestation as an artifact, and thus you refer to the attestation as a hash over the envelope.

[...]

PR #143, opened 2023-03-03
Add new ResourceDescriptor field type

marcelamelara:
This PR introduces a new field type used to describe software artifacts or resources within attestation predicates. The ResourceDescriptor field type is designed to address requirements from various issues: #2 , #27 #117, https://github.com/slsa-framework/slsa/issues/614 and https://github.com/slsa-framework/slsa/issues/649.

This PR replaces #133. Part of #130.


marcelamelara:
@MarkLodato PTAL.

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Fields
+
+Though all fields are optional, a ResourceDescriptor MUST specify at least
+one of `uri` or `digest` at a minimum.
+
+`name` _string, optional_
+
+> Human-readable identifier to distinguish the resource or artifact locally.
+> The semantics are up to the producer and consumer.
```
MarkLodato, line 30: nit: I think this can be interpreted as two distinct purposes (same problem as Subject.name):

- A human-readable "display name" for debugging, which might not be unique or might change unexpectedly
- A machine-readable "filename" for distinguishing between descriptors, which is expected to be stable so that consumers can rely on it

Can we specify that it's only the latter? In other words, that it is intended to be stable filename or similar so that consumers can rely on it?

Also it's probably worth comparing to `uri`. I interpret this as `name` being local to the attestation, while `uri` is universal.

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
+# ResourceDescriptor field type specification
+
+Version v1.0-draft
+
+A size-efficient description of any mutable or immutable software artifact
```
TomHennen, line 5: nit:

A size-efficient description of any software artifact or resource (mutable or immutable).

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+
+`representation` _object, optional_
+
+> A representation of the contents of the resource or artifact.
+> The producer MAY use this field in scenarios where including the contents
+> of the resource/artifact directly in the attestation is deemed more
+> efficient for consumers than providing a reference to another location.
+>
+> The producer and consumer must agree on the semantics and acceptable
+> formats for the `representation` object.
```
MarkLodato, line 52: I think this would be more clear as `value` with type `bytes` (encoded as base64). Otherwise you cannot digest it. (Same comment as https://github.com/slsa-framework/slsa/issues/649#issuecomment-1454078960.)

Would also be good to RECOMMEND that the size be kept small.

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+}
+```
+  
+Pointer to another in-toto attestation:
+  
+```jsonc
+ { 
+   "name": "gcc9.3.0-rebuilderd-attestation",
+   "digest": { "sha256": "abcdabcde..." },
+   "downloadLocation": "http://example.com/rebuilderd-instance/gcc_9.3.0-1ubuntu2_amd64.att",
```
TomHennen, line 122: nit: should this be .intoto.jsonl or is .att fine for a singular attestation?

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+For consistency, we RECOMMEND the following:
+
+-   A descriptor that specifies a `digest` is assumed to refer to an
+immutable resource or artifact.
+-   When `uri` AND `digest` are specified, the descriptor is bound to the
+`digest` field, and all other fields are considered informational.
+-   A descriptor without a `representation`, is assumed to serve as a
+pointer to the resource/artifact.
+
+## Examples
```
TomHennen, line 86: Should there be an example of 'representation' ?

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+> algorithms.
+
+`representation` _object, optional_
+
+> A representation of the contents of the resource or artifact.
+> The producer MAY use this field in scenarios where including the contents
+> of the resource/artifact directly in the attestation is deemed more
+> efficient for consumers than providing a reference to another location.
+>
+> The producer and consumer must agree on the semantics and acceptable
```
TomHennen, line 51: I wonder if we have any more guidance (or just a disclaimer!) for how they're supposed to agree on this?  is it via the predicateType?  The identity of producer?

One consumer might have two producers creating the same predicate type but doing different things with representation or annotations.

Maybe it's a non-issue?

TomHennen (changes requested):
Thanks for putting this together Marcela!

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+## Semantics
+
+Though the ResourceDescriptor allows for a high degree of flexibility,
+certain field combinations typically have specific semantics.
+For consistency, we RECOMMEND the following:
+
+-   A descriptor that specifies a `digest` is assumed to refer to an
+immutable resource or artifact.
+-   When `uri` AND `digest` are specified, the descriptor is bound to the
+`digest` field, and all other fields are considered informational.
```
MarkLodato, line 82: I'm not sure I agree with this statement. If a SLSA Provenance lists:

```json
"resolvedDependencies": [{
  "uri": "pkg:pypi/requests@1.2.3",
  "digest": {"sha256": "abcd1234..."}
}]
```

both `uri` and `digests` are important. The build interpreted the file with hash `abcd1234...` as the Python "requests" package version 1.2.3. This URI is important for policy decisions.

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+  "uri": "git+https://android.googlesource.com/platform/vendor/foo/bar@16244f4e7524d44a8f3060905eaf9190e96e9fb0#prebuilts/Foo/Foo.apk",
+  "digest": { "sha256": "7f4714fd..." }
+}
+```
+
+Pointer to a git repo (with annotations):
+
+```jsonc
+{
+  "uri": "git+https:github.com/actions/runner/tree/main",
```
MarkLodato, line 110:
```suggestion
  "uri": "git+https://github.com/actions/runner",
```

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,139 @@
...
+```
+  
+Pointer to another in-toto attestation:
+  
+```jsonc
+ { 
+   "name": "gcc9.3.0-rebuilderd-attestation",
+   "digest": { "sha256": "abcdabcde..." },
+   "downloadLocation": "http://example.com/rebuilderd-instance/gcc_9.3.0-1ubuntu2_amd64.att",
+   "mediaType": "application/vnd.in-toto+json"
```
MarkLodato, line 123: This isn't correct. That's the mediaType of the *payload* of the attestation. The attestation itself is a DSSE or something else.

Maybe just omit this example for now since DSSE doesn't have a media type currently?

MarkLodato (changes requested):
Thanks! Overall I like the idea, but I have a few questions.

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Yep, I have a TODO for that. I was waiting on a real usage example: https://github.com/slsa-framework/slsa/issues/649#issuecomment-1454078960

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Good point. This is pretty tricky, but agree that it would be useful to add more guidance round this. I don't think it will make sense in every case to map the `representation` to the upper-level `predicateType`. Per our discussion in https://github.com/slsa-framework/slsa/issues/649#issuecomment-1454161037 I'm thinking that the `uri and/or `mediaType` could be used as hints by the producer, which is directly related to the contents of `representation`.

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Per Mark's [comment](https://github.com/in-toto/attestation/pull/143#discussion_r1124999516) below, I'll probably just end up taking out this example.

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Fair point. Agreed that the `name` should be stable and machine-readable, even if it can also be used as display info for humans to read. Will add these details to the field description.

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Yes, makes perfect sense. I felt the name `value` would be too general, which is why I had initially gone with the slightly more specific (IMO) `representation` (akin to the `repr()` function in python). But, I definitely agree that base64 is the better choice, so the name `representation` also doesn't really make sense. Could we just go with `content` since that's really what the field is for?

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Ah. This wasn't trying to suggest that the the other fields shouldn't matter for policy, but rather provide guidance for how to handle cases where both `uri` and `digest` fields are set. This statement was more about the the assumption that the `digest` SHOULD match the file at the `uri`. This might already be implied by the first bullet point in this list, though, and the consumer might choose to match the `digest` against some other field, so this statement isn't accurate either way.

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Thanks for pointing this out, great point. To be fair, the `application/vnd.in-toto` type isn't officially registered either, to my knowledge. We could, however, allow custom/unregistered `mediaTypes` that use the `x.` or `x-` prefix for cases like this one. WDYT?

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Mark and I are having a longer [discussion](https://github.com/in-toto/attestation/pull/143#discussion_r1124999516) about this example below that you probably have thoughts on.

spec/v1.0-draft/resource_descriptor.md
marcelamelara: Thanks!

spec/v1.0-draft/resource_descriptor.md
TomHennen: Yeah that makes a ton of sense of representation.

Would it make any sense for annotations to suggest following the mime convention of using `vnd.` prefixes for non standardized annotations?  (At the moment we have no standardized annotations but maybe some will come along)

spec/v1.0-draft/resource_descriptor.md
MarkLodato: `content` sounds good to me! That's much better than `value`.

Regarding "small", I had the same thought. I think it's a judgement call but haven't been able to think of more specific guidelines.

TomHennen: Suggestion for 'small', "SHOULD be less than 1KB".

spec/v1.0-draft/resource_descriptor.md
TomHennen: Suggested language

> There are no standardized annotations yet.  Users SHOULD follow [RFC 6838](https://www.rfc-editor.org/rfc/rfc6838.html#section-3.2) convention of prefixing annotation names with `x.`, `prs.`, or `vnd.` to avoid collisions with other users.

marcelamelara: This seems to be implying that annotations more broadly should be self-describing in some way. I don't know if a MIME type will always make sense. In that case, we might even just say that annotations SHOULD include either a TypeURI OR mime type, as you described.

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,167 @@
...
+`mediaType` _string, optional_
+
+> The [MIME Type][] (i.e., media type) of the described resource or artifact.
+>
+> For resources or artifacts that do not have a standardized MIME type,
+> producers SHOULD follow [RFC 6838] (Sections 3.2-3.4) conventions of
+> prefixing types with `x.`, `prs.`, or `vnd.` to avoid collisions with other
+> producers.
+
+`annotations` _object, optional_
```
TomHennen, line 80: Oh maybe where I've been getting confused is I've assumed 'annotations' is a map<string, object> because I assumed folks might have multiple different annotations they might want to apply.  So instead of 

```jsonl
"annotations": { "_type": "x.git.review", "twoPartyReview": false }
```

I'd expect something like

```jsonl
"annotations": {
    "x.git.review": { "twoPartyReview": false },
    "vnd.acme.scan": { "fuzzed": false, "licenseCompliant": true }
}
```

If you agree we should probably strike the `_type` suggestion (since it would prevent multiple annotations).

The text might the be something like

>  The producer and consumer must agree on the semantics and acceptable
> formats for the `annotations` objects. There are currently no standardized
> annotation formats. To enable consumers to parse the field, `annotations`
> keys SHOULD be self-describing via a name following the conventions of
> custom MIME types per [RFC 6838] Sections 3.2-3.4 (e.g. `x.`, `prs.`, `vnd.`).

WDYT?

TomHennen (changes requested):

spec/v1.0-draft/resource_descriptor.md
marcelamelara: a) Agree that the format should be map <string, object>, so will make those changes
b) Given #147, I'm almost wondering if we should avoid giving any kind of formatting guidelines for the annotations or extensions fields other than "avoid dots and dollar-signs in the strings" (field names with '$' are also not well-supported before MongoDB v5).

TomHennen (approved):
Thanks!

marcelamelara:
Some final edits after the latest round of feedback and #147. Apologies if I missed any earlier review comments that might have been lost after rebasing the branch. I'll be fixing the outstanding linter issues after the final round of reviews.

MarkLodato (changes requested):
Looks good overall, thanks! (Sorry for the delay in reviewing.)

Just minor suggestions

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,168 @@
...
+>
+> To enable automated downloads by consumers, the specified location SHOULD
+> be resolvable.
+
+`mediaType` _string, optional_
+
+> The [MIME Type][] (i.e., media type) of the described resource or artifact.
+>
+> For resources or artifacts that do not have a standardized MIME type,
+> producers SHOULD follow [RFC 6838] (Sections 3.2-3.4) conventions of
```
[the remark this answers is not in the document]
marcelamelara: Thanks!

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,168 @@
...
+    "<FIELD_2>": { /* object */ },
+    ...
+  }
+}
+```
+
+## Fields
+
+Though all fields are optional, a ResourceDescriptor MUST specify one of
+`uri` or `digest` at a minimum.
```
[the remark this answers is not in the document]
marcelamelara: Yes, good catch, thank you.

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,168 @@
...
+> When a `digest`cannot be computed for the resource or artifact, the
+> producer MUST set this field.
+
+`digest` _[DigestSet], optional_
+
+> A set of cryptographic digests of the contents of the resource or artifact.
+>
+> When known, the producer SHOULD set this field to denote an immutable
+> artifact or resource. The producer and consumer must agree on acceptable
+> algorithms.
```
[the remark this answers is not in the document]
marcelamelara: I've updated the Semantics section in a way that will hopefully clarify how to handle `digest`. I'm inclined to not put such a strong requirement on the `digest` and `content` combination, because I think there might be cases where `uri`, `digest` and `content` are all set, and somehow the `digest` is computed over the `uri` rather than the `content`. In the TektonTask example, for instance, it's not immediately clear if the report pointed to by the `uri` is a larger artifact that has the `digest` and the `content` just contains a summary of the larger report.

>  a consumer MAY compute an arbitrary digest over content and consider that equivalent to being set here

Is this referring to a `digest` field _within_ the `content`?

spec/v1.0-draft/resource_descriptor.md
```diff
@@ -0,0 +1,168 @@
...
+> The semantics are up to the producer and consumer. The `name` SHOULD be
+> stable, such as a filename, to allow consumers to reliably use the `name`
+> as part of their policy.
+
+`uri` _[ResourceURI], optional_
+
+> A URI used to identify the resource or artifact globally.
+>
+> When a `digest`cannot be computed for the resource or artifact, the
+> producer MUST set this field.
```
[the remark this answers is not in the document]
marcelamelara: I went with option 2. In this case, I think re-stating this reduces ambiguity.

pxp928 (approved):
LGTM. Thanks @marcelamelara, captured very well.

spec/v1.0-draft/resource_descriptor.md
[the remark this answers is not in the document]
MarkLodato: Let's not block this PR on that. I'll might send a follow-up PR with a proposed tweak for the wording, but this is good enough for now.

MarkLodato (approved):

spec/v1.0-draft/resource_descriptor.md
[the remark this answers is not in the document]
marcelamelara: Thanks!

MarkLodato:
Friendly ping. Can this be merged now? https://github.com/slsa-framework/slsa/pull/684 is blocked by this.

PR #144, opened 2023-03-03
Relax subject name

### LINKED ISSUE #142: Relax subject.name requirements for v1
[`subject.name` currently says](https://github.com/in-toto/attestation/blob/main/spec/v1.0-draft/statement.md#:~:text=subject%5B*%5D.name%20string%2C%20required):

'If the name is not meaningful, use "_"'

and 

'MUST be non-empty and unique within subject.'

This means that for cases where the name isn't meaningful, but there are multiple subjects, producers must specify a name, even if they don't care what it is.

That's usually fine, however when there is a large number of subjects this leads to a case where the size of the statement can become _very_ large and filled with information that is "not meaningful".  E.g. we have a case with ~92K subjects.  If we populate name with the filename (which we don't really care about) the resulting statement is ~25MB.  If, however, we don't set the subject name _at all_ then the resulting statement is ~11MB, a significant savings in size and generation time.

Suggestion:
Make `subject.name` _optional_ but require "if set MUST be set for _all_ subjects and MUST be non-empty and unique within subject".

This may cause trouble for parsers who turn the repeated subject field into a map keyed by name.  We would encourage them to generate some unique value when the encounter empty/missing names.

Does this sound reasonable?

TomHennen:
Don't make producers set subject.name if it doesn't mean anything.
This simplifies their workflow and reduces the size & cost of
producing statements when there are a lot of subjects.

Retain option to use '_' for backwards compatibility, could
probably remove that without any harm.

fixes #142

### DIFF
--- a/spec/v1.0-draft/statement.md
+++ b/spec/v1.0-draft/statement.md
@@ -54,13 +54,15 @@ Additional [parsing rules] apply.
 
 > Identifier to distinguish this artifact from others within the `subject`.
 >
-> The semantics are up to the producer and consumer. Because consumers evaluate
-> the name against a policy, the name SHOULD be stable between attestations. If
-> the name is not meaningful leave the field unset or use "\_". For example, a
-> [SLSA Provenance] attestation might use the name to specify output filename,
-> expecting the consumer to only considers entries with a particular name.
-> Alternatively, a vulnerability scan attestation might leave name unset
-> because the results apply regardless of what the artifact is named.
+> The semantics are up to the producer and consumer and they MAY use it when
+> evaluating policy. If the name is not meaningful leave the field unset or
+> use "\_". For example, a [SLSA Provenance] attestation might use the name
+> to specify output filename, expecting the consumer to only consider
+> entries with a particular name. Alternatively, a vulnerability scan
+> attestation might leave name unset because the results apply regardless of
+> what the artifact is named.
+>
+> If set `name` SHOULD be unique within subject.
 
 `predicateType` _string ([TypeURI]), required_
 
--- a/spec/v1.0-draft/statement.md
+++ b/spec/v1.0-draft/statement.md
@@ -55,14 +55,14 @@ Additional [parsing rules] apply.
 > Identifier to distinguish this artifact from others within the `subject`.
 >
 > The semantics are up to the producer and consumer and they MAY use it when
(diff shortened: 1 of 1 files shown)


TomHennen:
@MarkLodato @AdamZWu FYI

spec/v1.0-draft/statement.md
```diff
@@ -50,19 +50,19 @@ Additional [parsing rules] apply.
...
 >
-> The semantics are up to the producer and consumer. Because consumers evaluate
-> the name against a policy, the name SHOULD be stable between attestations. If
-> the name is not meaningful, use "\_". For example, a [SLSA Provenance]
-> attestation might use the name to specify output filename, expecting the
-> consumer to only considers entries with a particular name. Alternatively, a
-> vulnerability scan attestation might use the name "\_" because the results
-> apply regardless of what the artifact is named.
+> The semantics are up to the producer and consumer and they MAY use it when
+> evaluating policy. If the name is not meaningful leave the field unset or
```
MarkLodato, line 58:
```suggestion
> evaluating policy. If the name is not meaningful, leave the field unset or
```

spec/v1.0-draft/statement.md
```diff
@@ -50,19 +50,19 @@ Additional [parsing rules] apply.
...
+> use "\_". For example, a [SLSA Provenance] attestation might use the name
+> to specify output filename, expecting the consumer to only consider
+> entries with a particular name. Alternatively, a vulnerability scan
+> attestation might leave name unset because the results apply regardless of
+> what the artifact is named.
 >
-> MUST be non-empty and unique within `subject`.
+> If set `name` SHOULD be unique within subject.
```
MarkLodato, line 65:
```suggestion
> If set, `name` SHOULD be unique within subject.
```

MarkLodato (approved):

spec/v1.0-draft/statement.md
TomHennen: You know, I _had_ a comma there and second guessed myself.

Done.

spec/v1.0-draft/statement.md
TomHennen: Done

marcelamelara (approved):
LGTM. Thanks, Tom!

pxp928 (approved):
LGTM! I agree that the name should be optional if not useful. Thanks @TomHennen

PR #146, opened 2023-03-07
Remove .0 from the _type uri

TomHennen:
Any minor changes should be backwards compatible so there's no need to identify type in the statement and it could actually be harmful.

Leaving it as is in docs and such because the text could change with point releases.

refs #130


pxp928 (approved):

pxp928:
Thanks @TomHennen

joshuagl (approved):

PR #148, opened 2023-03-08
Update extension fields

### LINKED ISSUE #147: Extension fields and dots
I am trying to store an SLSA Provenance which defers to the in-toto specification for extension fields:

> Extension fields: Producers MAY add extension fields to any JSON object by using a property name that is a [TypeURI](https://github.com/in-toto/attestation/blob/main/spec/v1.0-draft/scalar_field_types.md#TypeURI). The use of URI is to protect against name collisions. Consumers MAY parse and use these extensions if desired. The presence or absence of the extension field MUST NOT influence the meaning of any other field, and the field MUST follow the monotonic princple.

This format results in field names that are not legal in MongoDB.   The period is used in queries to query subdocuments.   I suspect this will be a problem with other document based stores.  It seems from the RFC that the TypeURI is meant to be used as a way of creating unique values (like its use for _type), but not intended as a way of creating unique field names.

TomHennen:
Extension fields no longer recommend using TypeURI as a name since this can cause problems for some users.

Instead we recommend avoiding `.` and `$` and simply advise users to avoid names that may collide, without telling
them how to do so.  Users are encouraged to file an issue to document extensions they're using.

fixes #147.

### DIFF
--- a/spec/v1.0-draft/README.md
+++ b/spec/v1.0-draft/README.md
@@ -33,10 +33,10 @@ model.
     with names used by other orgs. Producers MAY file an issue/PR to document
     extension fields they're using. Any consumer MAY parse and use these
     extensions if desired.
-    
+
     Field names SHOULD avoid special characters like `.` and `$` as these
     can make querying these fields in some databases more difficult.
-    
+
     The presence or absence of the extension field MUST NOT influence the
     meaning of any other field, and the field MUST follow the monotonic
     princple.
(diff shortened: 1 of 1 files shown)


TomHennen:
@danzehme does this resolve your issue?

TomHennen:
@MarkLodato FYI, I'm not sure if this will change anything for SLSA?  We don't forbid using TypeURI as field names, but it's discouraged.

danzehme:
> @danzehme does this resolve your issue?

Yes, that most certainly would as that makes the field name rather unrestricted.

marcelamelara (approved):
LGTM Tom! Thanks!

PR #150, opened 2023-03-10
Fix spdx download location URL

TomHennen:
The old link stopped working, it seems like we should link to a specific version.


pxp928 (approved):
LGTM

marcelamelara (approved):
LGTM. Thanks!

Issue #124, opened 2023-01-24
A Source Attestation for recording metadata about source code


trishankatdatadog:
Presently, there is no proposed Attestation for recording metadata about source code that developers may have written. For example, the [Datadog Agent Integrations](https://www.datadoghq.com/blog/engineering/secure-publication-of-datadog-agent-integrations-with-tuf-and-in-toto/) use [in-toto links](https://github.com/DataDog/integrations-core/tree/master/.in-toto) to record what source code was checked in by whom.

An [example](https://github.com/DataDog/integrations-core/blob/38bbffef6cefa7a12fd31961eb11521cf0e4fe5c/.in-toto/tag.5e2e6791.link) Source Attestation may look like:

```json
{
  "signatures": [
    {
      "keyid": "5e2e6791ad1dca571dd782006dc8f36eeba9870c",
      "other_headers": "04000108001d1621045e2e6791ad1dca571dd782006dc8f36eeba9870c050263c5145b",
[...]

Issue #46, opened 2021-07-16
Timestamp somewhere?


dlorenc:
We've been discussing support for vulnerability scans as a type of "attestation" over here: https://github.com/sigstore/cosign/issues/442

and it's clear that these will need some form of timestamp to work correctly. A vulnerability scan is timely, and should only be considered valid for a specific period of time after it is generated. This also helps align with the principle of "monotonicity", where the absence of an attestation should never move a decision from DISALLOW to ALLOW.

This could be done with a timestamp inside a custom scan predicate, but it might also be useful to place this at the statement layer. I'm not convinced either way yet.

cc @joshuagl @SantiagoTorres

TomHennen:
I think the argument against putting it in a generic field at the statement layer is that it would be difficult to convey what the timestamp means for all predicates.

This is [discussed a little bit here](https://github.com/in-toto/attestation/tree/main/spec#predicate-conventions) where it says

> Timestamps SHOULD use RFC 3339 syntax with timezone "Z" and SHOULD clarify the meaning of the timestamp. For example, a field named timestamp is too ambiguous; a better name would be builtAt or allowedAt or scannedAt.

That leans towards having more specific timestamp fields in the predicate layer and is why [the Provenance predicate](https://github.com/in-toto/attestation/blob/main/spec/predicates/provenance.md) defines `buildStartedOn` and `buildFinishedOn`.

Are there use cases where someone might want to make a blanket statement against multiple different predicate types based on the value of the timestamp field?  What would be the disadvantages of only having specific timestamp fields in the predicate?

dn-scribe:
Following the rational of separating statements from predicates - one could imagine a policy of "Must have a fresh attestation", or  "order of attestations to sub-processes should be 1-2-3" . 

I think there is a more important subject behind the timestamp issue - which is the context of the attestation (the timestamp is only one component of such a context). In SLSA v.2 environment object can suffice, but I would argue that any attestation requires some context, and thus it should be a part of the statement. Examples to context data: pipeline-run, job-ID, GitHub repo, machine-id. 
The challenge is to identify useful, common, mandatory context fields, and to enable user-defined fields.

mattmoor:
For context-sensitive `timestamp` properties, I completely agree that they belong inside of the `predicate` itself.

[...]

PR #151, opened 2023-03-13
Add predicates for human reviews

### LINKED ISSUE #77: Defining a generalized predicate format for "reviews"
We started discussing what code review attestations should look like, and @iamwillbar suggested checking if we could define a generalized predicate for reviews. We could then derive code review and other formats (perhaps vuln scans like #58) from this general predicate. As a starting point, we've potentially got the following sections in the predicate.

### Meta Information (better name?)

In the case of code reviews, this could identify the person performing the review, but in the case of a vuln scan, it could identify the scanner used. One thing to note is that this _may_ be unnecessary in the case of some reviews, because we could use the signature to identify the functionary as we do with links. Further, we could capture temporal information about when the review was performed.

### Artifact Information

This section essentially identifies the artifacts the review applies to. We can incorporate ITE-4 here as well. One project we're looking to work with is [crev](https://github.com/crev-dev/crev), so we could use something like purl as well to identify package-level reviews as a whole. One open question may be reviews for diffs, and how they'd chain together to apply to a file as a whole.

### Review Result

This is pretty self explanatory for both CRs and vuln scans. I'm not sure if a negative result CR should exist, except maybe for package-results as they're used in crev.

---

These are all early thoughts in defining a generalized review format, and I'm curious to hear what people think about this. Also open to hearing about other projects like crev we should be paying attention to when working on this.

adityasaky:
This is still a WIP. Crev needs a major update but I took a pass at defining a generic human review attestation type followed by system specific ones that extend the generic predicate.

Fixes #77


spec/predicates/human-review/vcs.md
```diff
@@ -0,0 +1,103 @@
...
+        {
+            "digest": { "gitCommit": "<HASH>" }
+        },
+    ],
+    "predicateType": "http://in-toto.io/attestation/human-review/vcs/v0.1",
+    "predicate": {
+        "result": "approved/changes-required/comment",
+        "reviewLink": "https://github.com/<>",
+        "timestamp": "<TIMESTAMP>"
+        "reviewer": "<STRING>"
```
TomHennen, line 49: Should reviewer be a list or is the idea there would be one attestation per reviewer?

spec/predicates/human-review/vcs.md
```diff
@@ -0,0 +1,103 @@
...
+-   Fields marked _optional_ MAY be unset or null, and should be treated
+    equivalently. Both are equivalent to empty for object or array values.
+
+### Fields
+
+Extends: [human-review](./human-review.md#fields).
+
+`result` _enum_, _required_
+
+One of `approved`, `changes-requested`, or `comment`. TODO: This is inspired by
```
TomHennen, line 75: What should folks assume based on these various fields?  What should a user do if they get one of these with the result of 'comment'?  If there are multiple attestations (because of multiple rounds of review) would the idea be "look for however many 'approved' statues from individual reviewers you require?"

Is it useful for this predicate type to cover reviews that are in-progress (vs those that are just complete and the code merged)?

spec/predicates/human-review/crev.md
```diff
@@ -0,0 +1,127 @@
...
+dependency, and can be performed by one or more of several actors in the supply
+chain. The developer importing a new dependency can perform the review or a
+dedicated security team can be tasked with it.
+
+## Schema
+
+```json
+{
+    "_type": "https://in-toto.io/Statement/v1",
+    "subject": [{...}],
```
TomHennen, line 37: I think it would be helpful to have some language about what the subjects of these things are expected to be.

It looks like for crev it's expected to be a binary and not source code itself?

I also think it's fine to delegate the details to the various projects (crev doesn't seem to define this yet so maybe it's enough to just explicitly leave it open-ended)?

spec/predicates/human-review/vcs.md
```diff
@@ -0,0 +1,103 @@
...
+is that this attestation is generated as a confirmation, an approval of some
+specific change, and is performed by developers in the supply chain who have
+code review responsibilities.
+
+## Schema
+
+```json
+{
+    "_type": "https://in-toto.io/Statement/v1",
+    "subject": [
```
TomHennen, line 39: It would probably be good to discuss what the subject is (that it should point to some code in a VCS) and probably to discuss if the review covers the referenced _change_ or a review of the entire codebase?  Maybe that could be another field in the predicate?

TomHennen (changes requested):

spec/predicates/human-review/vcs.md
adityasaky: I think attestation per reviewer, that way we can plug into the threshold semantics directly for policies like two party review.

spec/predicates/human-review/vcs.md
adityasaky: I thought about this for quite a while, my original draft had attestation generation only on approval (and monotonic). I was worried we'd lose some signal in the process though. Do you think it makes sense to go back to that model?

Also to answer the question more generally, I think assuming based on these fields should be in the realm of policies consuming the attestation. I'm fine with sketching some of that out as examples for the predicate spec once we nail it down some more.

spec/predicates/human-review/vcs.md
adityasaky: That's a great point, I considered supporting a _range_ of commits that comprise a changeset. I'll take a pass at tweaking it.

spec/predicates/human-review/human-review.md
```diff
@@ -0,0 +1,106 @@
...
+```json
+{
+    "_type": "https://in-toto.io/Statement/v1",
+    "subject": [{
+        "digest": {
+            "gitCommit": "b0bd8ecab5607e174fa403e002c74a666e7edd51"
+        }
+    }],
+    "predicateType": "https://in-toto.io/attestation/human-review/v0.1",
+    "predicate": {
```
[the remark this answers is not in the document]
TomHennen: Interesting, I would imagine that in some cases the code review system might wind up creating this attestation.  E.g. GitHub creates this attestation for the reviewer, signed with GitHub keys and that GitHub would be the identified attestor.  That's my thinking because GitHub is the system that's actually providing the UI to do the review and so it is in a position to make these attestations.

If it's the individual doing the review that's making/signing the attestation (e.g. with sigstore) that's different.  Of course then there will be the problem of mapping GitHub identities with Sigstore ones (maybe that's already a feature?).

Either way it's probably worth adding text describing who the various attestors could be?

spec/predicates/human-review/vcs.md
TomHennen: Would it make sense, given ITE-9, that this proposal should probably include some concrete use cases with example policies?

FWIW I'm not sure what I'd actually do with a 'changes-requested' review or a 'comment' review.

spec/predicates/human-review/crev.md
adityasaky: > It looks like for crev it's expected to be a binary and not source code itself?

I understand it as a version of a package. I think resource descriptors and/or purl might fit the criteria. I'll update that.

spec/predicates/human-review/vcs.md
adityasaky: Is it too complex to have the subject point to the tip and include starting commit for the range in the predicate?

spec/predicates/human-review/vcs.md
adityasaky: I think there's interesting scope here for making the vcs one:
1. approve only
2. include 1+ reviewer details.

This addresses the signed-by-github point above. For human created ones, we can use the generic one. Thoughts? I could even split vcs into github and gerrit to show off how they'd differ, perhaps?

spec/predicates/human-review/vcs.md
adityasaky: What if we leaned into ITE-4 to make the subject of a github-specific attestation a pull request itself? I think it could work but we'd then have to identify how we link the commits in the PR to that being used as a source for the build etc.

TomHennen: We've been thinking pretty hard about this on our end, but it really comes down to what you're trying to do.

My guess is that people will want to look at a specific commit/merge and ask if it was reviewed.  The attestation should probably provide a link to the review (ITE-4 can probably help there, but maybe it can just be a URI?).  Then I suspect that some users will want to perform that same check for each commit in the log.

If that's true then I think the subject should be the commit where the PR was merged.  IMO we'd only want the attestation for the final commit because any review or approval probably applies to the _sum_ of the changes and not the individual commits in the PR.

That means that for a branch to be considered reviewed you want each commit the branch pointed to to have a review that applies to it, but you don't necessarily care about the intermediate commits since the branch never pointed to them individually and you likely don't want people to use those commits directly anyways.

What goal do the users of this predicate have in mind "how do I know if the all the code in this repo is safe?", "how can I identify which things haven't been reviewed so I can take a closer look at them?", "I'd like to make sure I trust the folks that provided all reviews on this repo so I need to know who they are", "I'd like to make sure the code in this repo was reviewed for problem X".  "I'd like to have confidence that, at some point, the code in this repo was evaluated for problems X, Y, Z, but I don't need to ensure that each individual commit meets that bar".

adityasaky: I think I agree. For a v0.1.0, I suggest we leave it as applying to one or more commits, up to the user. In automated systems, I foresee it being issued at the tip of the feature branch, and I think this works from the "was this change reviewed? give me a CR issued for this PR matching the tip of the PR's branch".

adityasaky:
> Have we given any thought tho how this interacts with https://github.com/in-toto/attestation/issues/124 and https://github.com/in-toto/attestation/issues/124#issuecomment-1488677973. cc @colek42

I think they're complementary. The review predicate can apply to sources recorded in links in a prior step, say from a git clone, and we can extend that with other notions about the repository using testifysec's git attestor.

protos/in_toto_attestation/predicates/human_review/crev/v0/crev.proto
```diff
@@ -0,0 +1,30 @@
...
+
+    string url = 3;
+}
+
+message Review {
+    string result = 1;
+
+    google.protobuf.Timestamp timestamp = 2;
+
+    reviewer reviewer = 3;
```
marcelamelara, line 23: Ditto:
```suggestion
    Reviewer reviewer = 3;
```

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,123 @@
...
+## Use Cases
+
+As noted above, crev enables social review of popular open source software
+dependencies. A crev review includes information such as the thoroughness of
+the review, understanding of the source code, and a final rating.
+
+## Prerequisite
+
+Understanding of the
+[in-toto attestation specification](https://github.com/in-toto/attestation).
```
marcelamelara, line 22: This isn't really a comment specific to this predicate, but I've seen several predicate specs say this, and I feel like this ought to just be implied. The pre-reqs I think are a lot more interesting and worth noting should be the lesser known, predicate-specific ones.

marcelamelara (commented):
Thank you for getting this started, @adityasaky ! I have a general question, and apologies if this has already been addressed: what's the reasoning for needing three separate predicates for this? The human review and VCS review predicate are practically identical and frankly don't capture much information. I really like the Crev format, and wonder if we can either a) simply use the Crev format in the other use cases, or b) extract the most important properties a consumer wants out of a human review attestation, and develop a single format that will be compatible with all three use cases (Crev, generic, and VCS). If we go with option b, I suspect the predicate will end up looking a lot like Crev, possibly with a few extra fields.

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,100 @@
...
+    "subject": [{
+        "name": "in-toto-python",
+        "digest": {
+            "gitCommit": "b0bd8ecab5607e174fa403e002c74a666e7edd51"
+        },
+        "uri": "https://github.com/in-toto/in-toto"
+    }],
+    "predicateType": "http://in-toto.io/attestation/human-review/vcs/v0.1",
+    "predicate": {
+        "reviewer": ["https://github.com/lukpueh"],
```
marcelamelara, line 95: I would be a little concerned with the fact that this reviewer URL isn't entirely immutable, which is why I think I favor the slightly more detailed Crev format for the reviewer field.

spec/predicates/human-review.md
```diff
@@ -0,0 +1,90 @@
...
+{
+    "_type": "https://in-toto.io/Statement/v1",
+    "subject": [{
+        "digest": {
+            "gitCommit": "b0bd8ecab5607e174fa403e002c74a666e7edd51"
+        }
+    }],
+    "predicateType": "https://in-toto.io/attestation/human-review/v0.1",
+    "predicate": {
+        "result": "pass",
```
marcelamelara, line 85: Maybe I'm missing something: How come there is no `reviewer` field in this predicate? I also would have expected to see a `result` field in the VCS predicate, hence my earlier comment about the two predicates being practically identical.

adityasaky: This one is a more generic review type. Here, the reviewer is expected to be identified using the signature rather than a field in the predicate. The vcs type is to support, say, github or a gitlab server issuing an attestation with an online key.

adityasaky: The `result` field being omitted in the VCS one is because in the VCS case, we don't want negative reviews. An attestation => approval (and adheres to the monotonic principle).

marcelamelara: For the generic review, why couldn't these same semantics (omission of `reviewer` implies check signer) be achieved with a general predicate that makes `reviewer` optional? 

Two comments on the VCS case: First, why wouldn't we ever want to record a negative review? What if we need forensic evidence of this review or to prove that the threshold number by approved reviewers was not achieved? 

Second, as I understand the monotonic principle, it's the consumer side that should be ensuring their policies follow the monotonic principle; a consumer policy that follows the monotonic principle should be one that says "deny unless results=pass". So by this, an omission of `results` isn't a monotonic way of expressing a positive result. While the producer should enable the consumer to make these types of policy decisions, I don't think the producer should be the one dictating the policy of the consumer. So, I still don't see why all the different use cases can't be achieved with a single code review predicate that makes the appropriate fields optional, allowing for as-rich or as-simple semantics as the consumer may need.

marcelamelara:
CC'ing @pdxjohnny as this in-toto predicate/topic may be of interest re: code ingestion validation.

spec/predicates/human-review-vcs.md
adityasaky: Could you elaborate? Do you mean because GitHub usernames can change and / or be registered by other people in future? To address that, we could use the user ID instead in the example and specify inline that it must be immutable for a user.

spec/predicates/human-review.md
adityasaky: > For the generic review, why couldn't these same semantics (omission of reviewer implies check signer) be achieved with a general predicate that makes reviewer optional? This is how SCAI handles the producer field as well, for example.

I think that complicates validating VCS attestations though, as the reviewer is compulsory there.

Does it make more sense to only submit VCS and Crev right now? The generic one is supposed to be more of a catch-all for users to issue reviews but if our use cases largely fall in the VCS / Crev buckets, we can revisit the generic one later.

> VCS case

Also see earlier discussion on this: https://github.com/in-toto/attestation/pull/151#discussion_r1147963407, cc @TomHennen 

Recording a negative review didn't make sense from use cases such as "was this PR reviewed and approved before it was merged". This is in part because we run into the question of when we care about PR reviews. I personally think that only the reviews (approvals) that apply at merge time matter, and that reviews (from the same users often) before the approvals that requested changes and left comments don't affect PR-specific workflows.

protos/in_toto_attestation/predicates/human_review/v0/human_review.proto
```diff
@@ -0,0 +1,16 @@
...
+
+package in_toto_attestation.predicates.human_review.v0;
+
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/in-toto/attestation/go/predicates/human_review/v0";
+option java_package = "io.github.intoto.attestation.predicates.human_review.v0";
+
+message Review {
+    string result = 1;
```
pxp928, line 11: missing `Reviewer reviewer`?

protos/in_toto_attestation/predicates/human_review/vcs/v0/vcs.proto
```diff
@@ -0,0 +1,16 @@
...
+
+package in_toto_attestation.predicates.human_review.vcs.v0;
+
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/in-toto/attestation/go/predicates/human_review/vcs/v0";
+option java_package = "io.github.intoto.attestation.predicates.human_review.vcs.v0";
+
+message Review {
+    repeated string reviewer = 1;
```
pxp928, line 11: why is this a string? while the others are missing `Reviewer reviewer`?

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,113 @@
...
+    "subject": [{
+        "name": "in-toto-python",
+        "digest": {
+            "gitCommit": "b0bd8ecab5607e174fa403e002c74a666e7edd51"
+        },
+        "uri": "https://github.com/in-toto/in-toto"
+    }],
+    "predicateType": "http://in-toto.io/attestation/human-review/vcs/v0.1",
+    "predicate": {
+        "reviewer": [{
```
pxp928, line 105: This does not follow the `REVIEWER` types?

spec/predicates/human-review.md
```diff
@@ -0,0 +1,85 @@
...
```
pxp928, line 80: missing `REVIEWER`?

protos/in_toto_attestation/predicates/human_review/v0/human_review.proto
adityasaky: Nope, this generic one was intended to be signed by the reviewer themself, so it didn't include the reviewer field which is needed when github or gerrit issues the attestation.

protos/in_toto_attestation/predicates/human_review/vcs/v0/vcs.proto
adityasaky: Updated to use RD based on today's discussion.

spec/predicates/human-review-vcs.md
adityasaky: This was previously set to be opaque per system but now uses RD.

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,156 @@
...
+Specifies if the overall rating of the dependency is `positive` or `negative`.
+
+`reviewer` _object_, _required_
+
+Identifies the reviewer. This has some meaning for crev's trust proliferation
+aspects, but the identity of the reviewer can also be mapped based on in-toto's
+functionary handling. `idType` is used to determine the contents of `reviewer`.
+The `url` is a reference to the reviewer's crev-proofs repository.
+
+`timestamp` _Timestamp_, _required_
```
TomHennen, line 70: nit: instead of calling this 'timestamp' which leaves what time it refers to ambiguous, can it be named review_time or something more descriptive?

See https://github.com/in-toto/attestation/issues/46 for some discussion

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,156 @@
+# Predicate type: Dependency Reviews (crev)
+
+Type URI: (tentative) https://in-toto.io/attestation/human-review/crev/v0.1
+
+Version: Do we use Crev's versioning? Currently -1? What about their URI?
```
TomHennen, line 5: Do we have a resolution for this?  If we don't know maybe just stick with v0.1?

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,122 @@
...
+review system. Each entry must include immutable information that is always
+mapped to the reviewer. The digest may be skipped in the ResourceDescriptor,
+using the name and URI fields to record the username and immutable ID.
+
+`reviewLink` _URI_, _optional_
+
+Contains a link to the full review. Useful to point to information that cannot
+be captured in the attestation.
+
+`timestamp` _Timestamp_, _optional_
```
TomHennen, line 88: As above can we give this a more descriptive name?

Also note that with multiple reviewers we might want to clarify which review this refers to.  First? Last?

TomHennen (changes requested):

spec/predicates/human-review-crev.md
adityasaky: I'll reach out to the crev developers again.

protos/in_toto_attestation/predicates/human_review/crev/v0/crev.proto
```diff
@@ -0,0 +1,30 @@
...
+
+    string id = 2;
+
+    string url = 3;
+}
+
+message Review {
+    string result = 1;
+
+    google.protobuf.Timestamp timestamp = 2;
```
TomHennen, line 21: nit: I think this field name changed?

protos/in_toto_attestation/predicates/human_review/vcs/v0/vcs.proto
```diff
@@ -0,0 +1,20 @@
...
+
+option go_package = "github.com/in-toto/attestation/go/predicates/human_review/vcs/v0";
+option java_package = "io.github.intoto.attestation.predicates.human_review.vcs.v0";
+
+message Review {
+    repeated in_toto_attestation.v1.ResourceDescriptor reviewers = 1;
+
+    string review_link = 2;
+
+    google.protobuf.Timestamp timestamp = 3;
```
TomHennen, line 17: nit: I think this field name changed?

TomHennen (changes requested):
Just two minor changes.

I think that's it besides resolving the question of including the version in the type?

spec/predicates/human-review-crev.md
adityasaky: I've added a note for now. The crev developers are open to us reimplementing crev without necessarily being backwards compatible. IMO we should start in a place of compatibility (or close-to-compatibility) and see where we go from there. For now, I've added a note about crev's versioning and updated to using in-toto specific URIs.

adityasaky:
I've added a new field to the VCS one called `target` that records the base branch and its state. This is, IMO, valuable when judging a changeset as it matters where it's being submitted to and when.

spec/predicates/human-review-vcs.md
adityasaky: I think I've addressed this by including immutable user IDs.

spec/predicates/human-review.md
adityasaky: I've since taken this file out, the PR submits one VCS specific predicate and one for crev.

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,138 @@
...
+### Parsing Rules
+
+This predicate follows the
+[in-toto Attestation Framework's parsing rules](../v1/README.md#parsing-rules).
+
+### Fields
+
+`reviewers` _array of ResourceDescriptor_, _required_
+
+Indicates uniquely the identity of one or more reviewers. MUST NOT be used over
```
zachariahcox, line 77: The source of the attestation is always the pull request product, right?

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,138 @@
...
+            "name": "lukpueh",
+            "uri": "https://api.github.com/user/589324"
+        }],
+        "target": {
+            "name": "refs/heads/develop",
+            "digest": {
+                "gitCommit": "330500b54433de4f6f9575676b67738b98ba5e54",
+            },
+        },
+        "reviewLink": "https://github.com/in-toto/in-toto/pull/503#pullrequestreview-1341209941",
```
zachariahcox, line 131: This spec doesn't describe the actual change set that was reviewed, and therefore places a lot of the load on this `reviewLink`.

For a given PR, the diffs reviewed and voted upon by each reviewer regularly change (unless rules are enforced to dismiss "stale" reviews).

Mechanically this review link is reverencing the critical "provenance" data maintained by the PR product (GitHub in this case). 
Pull request data is saved as evidence justifying the existence of the new repo version.

PR data is not quite as nice as signed attestation statements like this though.  
PR data can change over time, either by-design (EG: by rerunning checks on the same refs) or maliciously if an attacker gains access to the DB.

adityasaky: The idea is to infer the diff from the subject (which identifies the tip of the reviewed topic branch) and the base recorded in the predicate. I leaned away from including the full diff inline due to length considerations. `reviewLink` is included ATM for completeness, any information that isn't considered vital to the review process for the changeset but may still be necessary.

You make a good point about the iterative nature of reviews and that's where a tighter alignment with GitHub mechanics would be helpful to answer questions like when the attestation is signed and generated. If an attestation is generated at merge time, for example, it captures the final states which is _probably_ what we care about?

spec/predicates/human-review-vcs.md
adityasaky: Can you elaborate on `source`? Do you mean the subject? The subject of the attestation identifies the tip of the PR feature branch.

spec/predicates/human-review-vcs.md
adityasaky: TODO: `target` needs to be expanded to handle current tip of the base branch and the merge base for diff inference.

marcelamelara:
@adityasaky Can you help clarify this? As I re-review these predicates, I have a few questions and one major concern around Crev specifically. 

First, I really like the intent of the project and would definitely like to support it in in-toto, and I noticed that they actually support more SW supply chain info beyond the source code review proofs that this predicate introduces. Specifically, they also seem to support [package-level reviews](https://github.com/crev-dev/cargo-crev/blob/master/crev-lib/rc/doc/editing-package-review.md) as well as "meta-reviews" that they call [trust proofs](https://github.com/crev-dev/cargo-crev/blob/master/crev-lib/rc/doc/editing-trust.md). Do you intend to add predicates for these other types of Crev claims at a later point in time? AFAIU, the package review proof format is similar, if not identical to the code review proof format. So, I wonder if it would be worth creating a separate, more encompassing Crev "dependency review" predicate that can cover both types of dependency reviews?

Second, my major concern. The project do not seem to have an authoritative spec anywhere, i.e. I found a spec for a review proof in a few different places: [main repo](https://github.com/crev-dev/crev) vs [cargo-crev](https://github.com/crev-dev/cargo-crev/tree/master/crev-lib/rc/doc). According to the main Crev project repo, the [Rust implementation is the most-well maintained, up-to-date version of Crev](https://github.com/crev-dev/crev#further-reading), but I have no idea if the other implementations are truly consistent or anything. I am still in support of having a predicate for Crev, but I would feel a lot more comfortable if we can be unambiguously clear on our end about which specific version of the Crev review proof formats the predicate follows.

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,161 @@
...
+[in-toto Attestation Framework's parsing rules](../v1/README.md#parsing-rules).
+
+### Fields
+
+The subject of this predicate type is a specific package and its version in some
+ecosystem.
+
+`result`, _enum_, _required_
+
+Specifies if the overall rating of the dependency is `positive` or `negative`.
```
marcelamelara, line 65: We should probably support the full set of values for the "rating" field of the [Code Review Proofs](https://github.com/crev-dev/cargo-crev/blob/master/crev-lib/rc/doc/editing-code-review.md).

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,161 @@
...
+
+`reviewer` _object_, _required_
+
+Identifies the reviewer. This has some meaning for crev's trust proliferation
+aspects, but the identity of the reviewer can also be mapped based on in-toto's
+functionary handling. `idType` is used to determine the contents of `reviewer`.
+The `url` is a reference to the reviewer's crev-proofs repository.
```
marcelamelara, line 72: A question and a suggestion. Q: What are possible `idType` values, and who determines these? Rec: I'd add in here that this field is intended to correspond to the `from` field in the Code Review Proofs format when the `idType` matches the ID for Crev.

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Model
+
+Most modern software have external dependencies. Dependency review is the
+process of reviewing and verifying the source code of a particular
+dependency, and can be performed by one or more of several actors in the supply
+chain. The developer importing a new dependency can perform the review or a
+dedicated security team can be tasked with it.
+
+## Schema
```
marcelamelara, line 31: To my more general comment about which Crev spec is authoritative, I suggest indicating here which format this predicate follows. As indicated on the main project repo, I'm assuming the spec specified in [cargo-crev](https://github.com/crev-dev/cargo-crev/blob/master/crev-lib/rc/doc/editing-code-review.md) is the authoritative one.

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,161 @@
...
+
+Identifies the reviewer. This has some meaning for crev's trust proliferation
+aspects, but the identity of the reviewer can also be mapped based on in-toto's
+functionary handling. `idType` is used to determine the contents of `reviewer`.
+The `url` is a reference to the reviewer's crev-proofs repository.
+
+`reviewTime` _Timestamp_, _required_
+
+Indicates time of review creation. `timestamp` in the original crev
+specification.
```
marcelamelara, line 77: This field seems to have been renamed.
```suggestion
Indicates time of review creation. `date` in the original crev
specification.
```

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,161 @@
...
+
+`thoroughness` _enum_, _required_
+
+Describes how thorough the reviewer was. Must be set to one of `low`, `medium`,
+or `high`.
+
+`understanding` _enum_, _required_
+
+Describes the reviewer's understanding of the dependency code. Must be set to
+one of `low`, `medium`, or `high`.
```
marcelamelara, line 87: As I mentioned in my comment above, I'd suggest we support using all possible values for these fields according to the [cargo-crev](https://github.com/crev-dev/cargo-crev/blob/master/crev-lib/rc/doc/editing-code-review.md#data-fields) data fields spec.

spec/predicates/human-review-crev.md
```diff
@@ -0,0 +1,161 @@
...
+
+`comment` _string_, _optional_
+
+Optional field with any other comments a reviewer may have about the
+dependency.
```
marcelamelara, line 92: Which consumers, if any, of this predicate need to be able to parse/care about the contents of the `comment` field? This would be helpful info.

marcelamelara (changes requested):
Thanks for these changes @adityasaky ! This review only covers the Crev predicate, and I'll come back to review the VCS predicate.

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,155 @@
+# Predicate type: VCS Code Reviews
+
+Type URI: (tentative) https://in-toto.io/attestation/human-review/vcs/v0.1
+
+Version: 0.1.0
+
+## Purpose
```
marcelamelara, line 7: Given the Crev dependency review predicate is available as well, I think it would be helpful to add a short comment here about the complementary relationship between that and this predicate.

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,155 @@
...
+Changes to source code are reviewed by other members of the development team.
+The review may be performed using systems like Gerrit or via reviewing
+mechanisms built into version control forges such as GitHub and GitLab.
+
+### Threshold of Reviews on Pull Requests
+
+GitHub allows repository owners to create protection rules that require a
+threshold of approvals before a pull request can be merged. This predicate can
+be used to express meeting such a policy. In this case, GitHub would issue a
+signed attestation identifying all the reviewers for a pull request. The
```
marcelamelara, line 24: The idea is that this sort of attestation wouldn't be generated until the threshold is met, right?
```suggestion
signed attestation identifying all accepting reviewers for a pull request. The
```

marcelamelara: Also, this may be getting too much into the weeds.... the current GitHub rule for requiring approvals before a PR can be merged reads "When enabled, pull requests targeting a matching branch require a number of approvals _and no changes requested_ before they can be merged" (emphasis mine). 

I've personally experienced this: even though I had the threshold number of needed approvals, a third reviewer had requested changes in the past and not re-reviewed the PR. It wasn't until their review was either dismissed or updated that the PR was able to be merged. This may be a GH-specific rule, but what is the expectation for generating this attestation in this sort of case? The current text only describes how the attestation subject is chosen wrt a PR merge rule, but it may be worth clarifying the exact point or operation upon which this attestation should be generated (e.g., after the approved PR has been merged).

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,155 @@
...
+            "digest": { "gitCommit": "<HASH>" },
+            "uri": "<LINK TO REPO>"
+        },
+    ],
+    "predicateType": "http://in-toto.io/attestation/human-review/vcs/v0.1",
+    "predicate": {
+        "reviewers": ["<ResourceDescriptor>", ...],
+        "targetTip": "<ResourceDescriptor>",
+        "mergeBase": "<ResourceDescriptor>",
+        "reviewLink": "<LINK TO REVIEW>",
```
marcelamelara, line 62: Nit:
```suggestion
        "reviewLink": "<URI OF REVIEW>",
```

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,155 @@
...
+
+Contains a link to the full review. Useful to point to information that cannot
+be captured in the attestation.
+
+`reviewTime` _Timestamp_, _optional_
+
+Indicates the time of _last review_ creation. If the attestation is for a single
+review, this is equivalent to when that review was created. With multiple
+reviewers, the attestation is re-created for each new review and the time
+reflects that of the latest review.
```
marcelamelara, line 111: What happens to the old attestations that are superseded by the more recent review attestations? I'm wondering if we should add the ability to link to prior attestations related to the same subject PR to maintain full visibility into the review history.

marcelamelara (commented):
I really like where this is going. @adityasaky I mostly have clarifying comments for the VCS review predicate.

spec/predicates/human-review-crev.md
adityasaky: There's a bit of a disconnect here, because the crev [code review](https://github.com/crev-dev/cargo-crev/blob/master/crev-lib/rc/doc/editing-code-review.md) type includes `from` but missing from the [package review](https://github.com/crev-dev/cargo-crev/blob/master/crev-lib/rc/doc/editing-package-review.md) type. Same for `date`. I suspect `date` may be omitted as a package-wide review is meant as an ongoing document with updates as new versions / advisories emerge. On the other hand, I'm not sure about the `from` field. I wonder if in the in-toto context we want to drop it and lean entirely on the signature. That takes us to a broader conversation about how in-toto supports crev, though.

spec/predicates/human-review-vcs.md
adityasaky: These are great questions and have (somewhat) come up in the SLSA Source track doc as well. It all comes down to when the attestation is created and signed + what it's associated with. I authored this take on the VCS review predicate thinking it'd be generated at merge time. But this obfuscates cases where stale approvals aren't dismissed (which is I think the more common configuration). @zachariahcox has made this point in the Source track doc if I'm not mistaken, and his take on a similar predicate identifies associates each individual review of the changeset with the commit at the tip of the topic branch. Settling when the predicate is generated will help us decide what and how its contents must be presented IMO.

spec/predicates/human-review-vcs.md
adityasaky: Related: https://github.com/in-toto/attestation/pull/151#discussion_r1330386183

I think it depends on the intended policies. I'm not sure we have policies / use cases that require full review history, but I'm not opposed to including the information. This is currently written with the admittedly simpler rule `k of n approve changes` in mind.

marcelamelara: For a v0.1, I think the `k out of n` assumption is actually fine. I would just make it explicit, if it isn't stated anywhere yet.

spec/predicates/human-review-crev.md
```diff
@@ -71,26 +77,30 @@ aspects, but the identity of the reviewer can also be mapped based on in-toto's
...
+`date` _Timestamp_, _required_
 
 Indicates time of review creation. `timestamp` in the original crev
 specification.
 
 `thoroughness` _enum_, _required_
 
 Describes how thorough the reviewer was. Must be set to one of `low`, `medium`,
-or `high`.
+`high`, or `none`.
```
marcelamelara, line 88: Semantically, what does a `none` value tell a consumer of the attestation? Could this be interpreted as an "endorsement" of a particular package?

spec/predicates/human-review-crev.md
```diff
@@ -12,21 +12,25 @@ Version: v0.1.0
...
+implementation.
 
 ## Use Cases
 
-As noted above, crev enables social review of popular open source software
-dependencies. A crev review includes information such as the thoroughness of
-the review, understanding of the source code, and a final rating.
+crev enables social review of popular open source software dependencies. A crev
+review includes information such as the thoroughness of the review,
+understanding of the source code, and a final rating. The ratings for these
```
marcelamelara, line 24: Being very nitpicky here, though I think this conveys the expectations more clearly.
```suggestion
understanding of a package's source code, and a final rating. The ratings for these
```

spec/predicates/human-review-crev.md
```diff
@@ -12,21 +12,25 @@ Version: v0.1.0
...
+fields are self-identified by each individual reviewer.
 
 ## Model
 
 Most modern software have external dependencies. Dependency review is the
-process of reviewing and verifying the source code of a particular
-dependency, and can be performed by one or more of several actors in the supply
-chain. The developer importing a new dependency can perform the review or a
-dedicated security team can be tasked with it.
+process of reviewing and verifying the source code of a particular dependency,
```
marcelamelara, line 30: Nit:
```suggestion
process of reviewing and verifying the source code of a particular open-source package,
```

marcelamelara (approved):
Only some minor nits, but LGTM otherwise. Once the final outstanding comments are resolved, I think this is ready for a v0.1 release.

spec/predicates/human-review-vcs.md
```diff
@@ -0,0 +1,155 @@
...
+        },
+    ],
+    "predicateType": "http://in-toto.io/attestation/human-review/vcs/v0.1",
+    "predicate": {
+        "reviewers": ["<ResourceDescriptor>", ...],
+        "targetTip": "<ResourceDescriptor>",
+        "mergeBase": "<ResourceDescriptor>",
+        "reviewLink": "<URI OF REVIEW>",
+        "reviewTime": "<TIMESTAMP>",
+        "annotations": {...}
```
jkjell, line 64: Perhaps this falls under the annotations (or could better fit into the CREV predicate) but, I wonder if there's some quantitative that would be useful to include here. For example, number of lines of code added/removed/modified. I've seen some projects require additional reviewers for larger feature work, and fewer reviewers for minor bug fixes. It may be hard to craft a policy around a metric like that but, most other options I can think of turn VCS specific pretty quickly (things like commit messages that link to specific issues with a label or tag for [feat] or [bug]).

marcelamelara:
@adityasaky Where have we landed on this? At the last attestation maintainer's meeting, we decided to close this issue if it's stalled while we wait for the SLSA Source Track to progress.

marcelamelara:
I propose that we close this PR since we haven't seen any updates in over a year, unless @adityasaky or someone else is able to pick this back up and get it across this finish line. We can always reopen this topic when there's more momentum behind it again.

TomHennen:
SGTM

adityasaky:
I think it's reasonable to close this. I think we might see something like the VCS predicate get proposed soon through https://github.com/slsa-framework/slsa/issues/1161, so I'll preserve this branch in case we want to build off that.

PR #152, opened 2023-03-13
Add predicate capturing results of test runs

adityasaky:


TomHennen:
FYI one of my colleagues who has done work in this area is going to take a look and provide some comments. Stay tuned.

spec/predicates/test-results.md
```diff
@@ -0,0 +1,148 @@
...
+
+Authors:
+
+## Purpose
+
+This predicate type defines a generic schema to express the result of running
+tests in software supply chains. The schema may be extended to support different
+types of testing or specific test harnesses.
+
+## Use Cases
```
clmccarthy5, line 15: Out of curiosity, do you have immediate use-cases in mind?

spec/predicates/test-results.md
```diff
@@ -0,0 +1,148 @@
...
+
+Understanding of the
+[in-toto attestation specification](https://github.com/in-toto/attestation).
+
+## Model
+
+This predicate type includes one compulsory field, `result`, that describes the
+result of the test run. The `testRun` object can be used to communicate a link
+to the test run and list tests that passed, failed, and passed with warnings. It
+also may optionally contain a reference to the configuration for the test run.
```
clmccarthy5, line 37: What's the rationale for making the configuration reference optional?

It seems that having only the result with no indication of which tests were run may not be very helpful (e.g. it'd be good to know if the configured tests were from trusted config or untrusted config). I suppose this relates to my question above, about the potential use-cases for the predicate

spec/predicates/test-results.md
adityasaky: I'm working with an adopter to deploy in-toto and I started sketching this out based on their test framework.

spec/predicates/test-results.md
adityasaky: Continuing from comment about implementer, I made the configuration optional right now but I'm waiting to hear what parts they're willing to share with the results. I think there's value with configurations as well, so I'm okay to change this.

Also, do you have any thoughts on other things we'd include here?

clmccarthy5: SGTM. Re: other things to include, I can't think of anything immediately -- IMO the reference to the test configuration and link to the test execution are the most important bits. It might be useful to consider the use-cases more explicitly, i.e. who is the consumer of test attestations, to see if any relevant info is missing (but feel free to ignore if there aren't concrete use-cases yet).

adityasaky: Done!

adityasaky: I just added some verification scenarios. Let me know if these make sense!

clmccarthy5: Those all LGTM, thanks for the additional details!

clmccarthy5 (approved):

TomHennen (approved):
Thanks Aditya!

adityasaky:
Thanks for the reviews! I've squashed my commits now.

spec/predicates/test-results.md
```diff
@@ -0,0 +1,172 @@
+# Predicate type: Test Results
+
+Type URI: https://in-toto.io/attestation/test-result/v0.1
+
+Version: 0.1.0
+
+Authors:
```
joshuagl, line 7: we should update this before merging :-)

spec/predicates/test-results.md
```diff
@@ -0,0 +1,172 @@
...
+2.  that all required tests passed
+
+Therefore, each attestation corresponds to one invocation of a test suite, and
+may include the results of several individual tests.
+
+### Asserting Test Configurations Used
+
+The supply chain owner creates a policy that records the expected test
+configurations. During verification, the policy checks that the test attestation
+used the right configurations. A custom inspection may optionally parse the
```
joshuagl, line 31: This is the first predicate to reference an inspection, should we be avoiding in-toto specification terms in the attestations repository (which aims to be policy engine agnostic)?

spec/predicates/test-results.md
```diff
@@ -0,0 +1,172 @@
...
+            "warnedTests": ["<TEST_NAME>", ...],
+            "failedTests": ["<TEST_NAME>", ...]
+        }
+    }
+}
+```
+
+### Parsing Rules
+
+This predicate follows the in-toto attestation parsing rules. Summary:
```
joshuagl, line 85: Broader question, do we want every predicate to include a summarised copy of the framework's parsing rules? Or just to note where the predicate's parsing rules differ from the framework's parsing rules?

WDYT @in-toto/attestation-maintainers ? We can move this to a separate issue, the discussion should not block this PR.

spec/predicates/test-results.md
```diff
@@ -0,0 +1,172 @@
...
+    equivalently. Both are equivalent to empty for object or array values.
+
+### Fields
+
+`result` _boolean_ , _required_
+
+Indicates the result of the test run. If true, it indicates _all_ tests passed
+in the corresponding run.
+
+`testRun` _object_, _optional_
```
joshuagl, line 105: Should `testRun` be required if the `configuration` child field is required?

joshuagl (commented):
This looks useful. I had a few questions (inline) about this predicate specifically, and a couple of higher-level questions: one about the relationship between in-toto attestations and the broader in-toto spec, and one about inclusion of attestation framework parsing rules in each predicate.

The higher level questions can be addressed separately and should not block this PR.

spec/predicates/test-results.md
adityasaky: Yep! I missed this when making configuration required.

spec/predicates/test-results.md
adityasaky: Speaking with my author of predicates hat on:

It _was_ a bit of work to go back and update each of them as the attestation framework wide rules changed. I could propose a change in ITE-9 that this is only to be used to define deviations from the regular rules.

joshuagl: Yeah, maintainability seems much easier with this in one place (that we link to from the predicate?)

spec/predicates/test-results.md
```diff
@@ -0,0 +1,172 @@
...
+
+Understanding of the
+[in-toto attestation specification](https://github.com/in-toto/attestation).
+
+## Model
+
+This predicate type includes one compulsory field, `result`, that describes the
+result of the test run. The `testRun` object can be used to communicate a link
+to the test run and list tests that passed, failed, and passed with warnings. It
+also may optionally contain a reference to the configuration for the test run.
```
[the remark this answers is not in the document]
adityasaky: Updated.

spec/predicates/test-results.md
adityasaky: I think it's okay to mention them in use cases. I can clarify that it's with in-toto layouts / policy engine. We could similarly add a non in-toto verifier and how it'd do something...

spec/predicates/test-results.md
adityasaky: Updated this one to only link to the spec.

spec/predicates/test-results.md
```diff
@@ -0,0 +1,163 @@
...
+
+In addition to the previous use case, the supply chain owner creates a policy
+verifying that test results passed. In the simplest case, the policy applies to
+all tests. Therefore, it asserts the contents of `result` and that
+`testRun.failedTests` is empty. In more nuanced cases, a subset of tests may
+matter. For example, the tested artifact may be an OS image that's to be
+deployed to three types of devices. As such, the test harness tests the new
+image on an instance of each device. When verifying the attestation prior to an
+image being installed on a device, it only matters that the tests passed on the
+corresponding test device and not necessarily the others.
```
marcelamelara, line 45: Could you please provide an example for this particular use case where only a subset of tests matter? Is the expectation that this information will be conveyed via the test names in the entries of `passedTests` etc? If so, I think that needs to be expressed more explicitly.

spec/predicates/test-results.md
```diff
@@ -0,0 +1,163 @@
...
+`testRun.link` _URI_, _optional_
+
+Contains a link to the test run. Useful to point to information that cannot be
+captured in the attestation.
+
+`testRun.configuration` _list of ResourceDescriptor_, _required_
+
+Reference to the configuration used for the test run.
+
+`testRun.passedTests` _list_, _optional_
```
marcelamelara, line 107: Ditto below.

```suggestion
`testRun.passedTests` _list of strings_, _optional_
```

spec/predicates/test-results.md
```diff
@@ -0,0 +1,163 @@
...
+
+Each entry corresponds to a single test that passed.
```
marcelamelara, line 109: I would add a comment here (as well as for `warnedTests` and `failedTests`) that the strings in this list are test names, and that their semantics are up to the producer and consumer to decide.

marcelamelara (changes requested):
Thanks @adityasaky !! I really like the simplicity of this predicate. I have several questions/requests for updates to improve clarity.

spec/predicates/test-results.md
```diff
@@ -0,0 +1,163 @@
...
+
+`testRun.warnedTests` _list_, _optional_
+
+Each entry corresponds to a single test that expressed a warning.
```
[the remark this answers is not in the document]
TomHennen: That makes me wonder if instead of having a top level true/false field it would make sense to have top level fields like:

```
testRun.numPassedTests
testRun.numWarnedTests
testRun.numFailedTests
```

Then evaluators are suggested to check:

1. numPassedTests > 0
2. numFailedTests == 0

But can check other stuff if they want too?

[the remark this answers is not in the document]
adityasaky: I think this can be calculated using `len(testRun.{passed,warned,failed}Tests)` correct?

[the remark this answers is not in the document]
TomHennen: Yeah that would work (details would differ depending on language).

If you take this approach would you drop `result`?

spec/predicates/test-results.md
adityasaky: Added an example with the device types named.

spec/predicates/test-results.md
```diff
@@ -0,0 +1,163 @@
...
+
+Indicates the result of the test run. If true, it indicates _all_ tests passed
+in the corresponding run.
+
+`testRun` _object_, _required_
+
+`testRun.link` _URI_, _optional_
+
+Contains a link to the test run. Useful to point to information that cannot be
+captured in the attestation.
```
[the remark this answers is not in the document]
adityasaky: Updated the text here.

spec/predicates/test-results.md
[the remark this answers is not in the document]
adityasaky: I think the result is useful as for simple cases, we can drop the passed, warned, failed lists altogether and just include the result field and the configuration.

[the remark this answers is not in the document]
adityasaky: Separately, I wonder if we can move all of the testRun fields to the top level? Now that I think about it, I'm not sure we need that separation.

adityasaky:
Does it make more sense for the schema to be as follows? Since `testRun` is now required because of the configuration, we can remove that intermediate field.

```json
{
    "_type": "https://in-toto.io/Statement/v1",
    "subject": [{...}],
    "predicateType": "https://in-toto.io/attestation/test-result/v0.1",
    "predicate": {
        "result": "pass|fail",
        "url": "<URL>",
        "configuration": ["<ResourceDescriptor>", ...],
        "passedTests": ["<TEST_NAME>", ...],
        "warnedTests": ["<TEST_NAME>", ...],
        "failedTests": ["<TEST_NAME>", ...]
    }
}
```

adityasaky:
nit: I'll reorder the fields to move required ones up. Configuration should be first since it's now required.

spec/predicates/test-results.md
```diff
@@ -91,11 +91,9 @@ This predicate follows the
 
 ### Fields
 
-`result` _boolean_ , _required_
+`result` _enum_ , _required_
 
-Indicates the result of the test run. If true, it indicates _all_ tests passed
-in the corresponding run. This means that `warnedTests` and `failedTests` must
-both be empty when they are used.
+Indicates the result of the test run. One of `PASSED`, `WARNED`, and `FAILED`.
```
TomHennen, line 96: nit: "or `FAILED`"

adityasaky: fixed

Issue #127, opened 2023-01-31
Clean up documentation


marcelamelara:
There's a lot of redundant and/or outdated information in the files in the [docs/](https://github.com/in-toto/attestation/tree/main/docs) folder. Would be great to get that cleaned up and updated.

TODO: 
- [x] Address: https://github.com/in-toto/attestation/pull/135#discussion_r1114674137

marcelamelara:
Include: https://github.com/in-toto/attestation/pull/135#discussion_r1114674137

joshuagl:
We should ensure that the 'monotonic' principle is clearly articulated cc @SantiagoTorres

joshuagl:
Some streamlining was merged in #153. Still to do:
- [ ] ensure that the 'monotonic' principle is clearly articulated 
- [ ] remove redundant examples
- [ ] fix motivating use cases to use specified attestation format

TomHennen:
This has languished for a while and @marcelamelara has done a lot of other work improving docs, so we think we can close this.  If anyone has other doc cleanup they'd like to see please file new issues on the specific things you'd like to see?

There might be more work to do on the monotonic principle and something DevRel related to improve adoption of the problem but those seem like larger issues?  FYI @jkjell?

PR #153, opened 2023-03-14
First pass at streamlining and improving consistency

joshuagl:
In #127 I promised to clean up redundant information in docs/ — well, I didn't get very far before running out of cycles. Here's the current WIP state.

Still to do:
- [ ] ensure that the 'monotonic' principle is clearly articulated
- [ ] remove redundant examples
- [ ] fix motivating use cases to use specified attestation format


TomHennen (changes requested):
Thanks! just one small comment

joshuagl:
> Thanks! just one small comment

I can convert this from draft if you think the changes here are useful?

TomHennen:
> > Thanks! just one small comment
> 
> I can convert this from draft if you think the changes here are useful?

Yes, I think they're useful.  So I think we should merge.

Do we think the "still to do" are blockers to v1?

joshuagl:
> Do we think the "still to do" are blockers to v1?

I am inclined towards no. Not least of all because these changes wouldn't require a version bump as they are not changes to the spec.

docs/README.md
```diff
@@ -1,12 +1,35 @@
...
-An **in-toto attestation** is authenticated metadata about one or more
-software artifacts[^1]. The intended consumers are automated policy engines,
-such as [in-toto-verify] and [Binary Authorization].
+The **in-toto Attestation Framework**
 
-**TODO: Clean this up; add a docs index.**
+1.  defines a standard [format](../spec/v1.0-draft/) for attestations which
+    bind subjects, the artifacts being described, to arbitrary authenticated
+    metadata about the artifact
+2.  provides a set of [pre-defined statements](../spec/predicates/) for
```
marcelamelara, line 8: Statements are something specific in the format, so we probably want to use a different term here, like "pre-defined claims formats" or "pre-defined metadata formats".

docs/README.md
```diff
@@ -1,12 +1,35 @@
...
+engines, such as [in-toto-verify] and [Binary Authorization]. However, any
+system that creates or consumes verifiable claims about how a piece of software
+is produced can benefit from the in-toto Attestation Framework.
+
+## Project goals
+
+-   Standardize artifact metadata without being specific to the producer or
+    consumer. This way CI/CD pipelines, vulnerability scanners, and other
+    systems can generate a single set of attestations that can be consumed by
+    anyone, such as [in-toto] or [Binary Authorization].
```
marcelamelara, line 22: This repeats the last paragraph in the previous section. Maybe merge them?

docs/README.md
```diff
@@ -1,12 +1,35 @@
...
+-   Make it possible to write automated policies that take advantage of
+    structured information.
+-   Fit within the [SLSA Framework][SLSA].
+-   Support an ecosystem of verifiable metadata about software artefacts to
```
marcelamelara, line 26: Nit: for consistency across the doc

```suggestion
-   Support an ecosystem of verifiable metadata about software artifacts to
```

docs/background.md
```diff
@@ -0,0 +1,95 @@
...
+We capture motivational context for the creation of the in-toto Attestation
+Framework here.
+
+## Motivation
+
+This project was initiated with two main motivations:
+
+1.  Support use cases where the prior in-toto link schema was a poor fit. For
+    example, test steps and vulnerability scans are not about "producing" a new
+    artifact so they are awkward to represent in the current format.
```
marcelamelara, line 12:
```suggestion
    artifact so they are awkward to represent in the original format.
```

docs/background.md
```diff
@@ -0,0 +1,95 @@
...
+2.  Support interoperability with [Binary Authorization], which will support
+    the agreed-upon format once finalized. This way we have a single ecosystem
+    of software supply chain security.
```
marcelamelara, line 15: Do we know that this is true?

docs/background.md
```diff
@@ -0,0 +1,95 @@
...
+## Requirements
+
+Functional requirements:
+
+-   Must support user-defined types and schemas, for two reasons:
+    -   To allow in-toto users to more naturally express attestations, as
+        explained above.
+    -   Because Binary Authorization does not want to require its users to use
+        the existing in-toto link schema, which is overly specific.
+-   Should allow indexing of attestations by artifact ID, without having to
```
marcelamelara, line 26: What's an "artifact ID"?

docs/background.md
```diff
@@ -0,0 +1,95 @@
...
+        attestation types.
+    -   Example: Given a provenance attestation for a docker image, it should be
+        possible to identify all the materials generically.
+    -   Implication: the association between attestation and related artifact
+        IDs must be standardized.
+
+Nonfunctional requirements:
+
+-   Must support backwards compatible links that can be consumed by existing
+    layout files.
```
marcelamelara, line 49: To clarify we mean the original in-toto formats.

```suggestion
-   Must support backwards compatible Links that can be consumed by existing
    Layout files.
```

marcelamelara (changes requested):
Thanks @joshuagl ! I mostly have minor wording changes, but there are a couple terms that we might need to better define (either as part of this PR or a separate one).

marcelamelara:
> > Do we think the "still to do" are blockers to v1?
> 
> I am inclined towards no. Not least of all because these changes wouldn't require a version bump as they are not changes to the spec.

The only TODO I would suggest we do try to target for v1 is the rewrite/clarification of the monotonic principle just because that does pertain directly to the usage of the spec.

pxp928 (commented):
looks good just a suggestion.

docs/motivating_use_case.md
```diff
@@ -119,7 +119,7 @@ allowed_artifact_id_types = [
 
 Let's take a look at one example team's software supply chain.
 
-![drawing](images/attestation_supply_chain.png)
+![drawing](../images/attestation_supply_chain.png)
```
pxp928, line 122: does it make sense to remove this for the time being as it's out of date? It might cause confusion.

docs/background.md
joshuagl: this part of the doc was [originally authored](https://github.com/in-toto/attestation/commit/9c61d1825f492dca00d6ee12f0f21ae695df3ee0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R627) by @MarkLodato, perhaps he can comment?

cc @TomHennen

docs/background.md
joshuagl: I believe this is a subject, or a product in the old Link format.

docs/README.md
joshuagl: I'll remove the "such as in-toto or Binary Authorization" but leave the rest as-is. We could certainly clean this up further, but I like the current split between design intent and design goals. WDYT?

docs/background.md
TomHennen: I'd suggest changing 'will' to 'can' just to cover our bases.

docs/background.md
TomHennen: Would it be worth changing 'artifact ID' to digest?

Perhaps a higher level question: can the intro to this doc just clearly indicate this is for historical reference and terms & statements made here were up-to-date at the time it was authored but may be out-of-date now?

docs/background.md
joshuagl: Done, thanks!

docs/background.md
joshuagl: I already changed to "artifact identifier, i.e., a digest", but can just swap this for digest if preferred?

I've updated the background section to highlight that this is historical reference, good suggestion -- thanks!

docs/motivating_use_case.md
pxp928: addressed by issue #157

marcelamelara (approved):
Thanks for the updates!

docs/README.md
```diff
@@ -21,110 +44,17 @@ Examples of hypothetical attestations:
...
+-   [ideas for future schemas](schema_ideas.md)
+-   [protobuf definitions](protos.md)
 
 [Binary Authorization]: https://cloud.google.com/binary-authorization
-[Predicate]: spec/README.md#predicate
-[SLSA Attestation Model]: https://slsa.dev/attestation-model
 [SLSA Provenance]: https://slsa.dev/provenance
 [SLSA]: https://github.com/slsa-framework/slsa
-[Statement]: spec/README.md#statement
 [in-toto]: https://in-toto.io
```
marcelamelara, line 59:
```suggestion
```

Issue #155, opened 2023-03-16
Capture proposed new predicates from repo


pxp928:
Currently, under the `docs` there are a bunch of future schema ideas that do not below. In order to clean this up, I have moved them to this issue for future tracking. 

## Terminology and concepts

-   **Artifact:** blob of data, identified solely by contents
-   **Resource:** uniquely identifiable thing that can be associated with an
    artifact

The following example shows various resources associated with the software
[curl](https://curl.se). Each is a way that a user may retrieve the software,
and each maps to a specific artifact at a point in time.

| Description    | Example Resource IDs                            | Example Artifact ID     |
| -------------- | ----------------------------------------------- | -------------------     |
| Git commit     | https://github.com/curl/curl, branch “master”   | Commit ID “9d954…”      |
|                | https://github.com/curl/curl, tag “curl-7_72_0” |                         |
| Source tarball | https://curl.se/download/curl-7.72.0.tar.gz     | SHA256 “d4d58…”         |
| Debian package | [curl][curl-debian] (latest version)            | SHA256 “aec36e…”        |
|                | [curl 7.72.0-1_amd64][curl-debian]              |                         |
| Docker image   | [docker.io/curlimages/curl:latest][curl-docker] | Digest “sha256:09a374…” |
|                | [docker.io/curlimages/curl:7.72.0][curl-docker] |                         |

[curl-debian]: https://packages.debian.org/search?keywords=curl
[curl-docker]: https://hub.docker.com/r/curlimages/curl

## Use cases

Attestation archetypes:

-   **Provenance:** how an artifact (or set of artifacts) came into being.
    -   Subject: artifact
-   **Artifact Analysis:** metadata about an artifact
-   **Artifact Equivalence:** two artifacts are equivalent
    -   Example: git commit == git tree == tar.gz == zip

**Open Question:** Are there any use cases that don't cleanly fit into one of
the archetypes above?

## Current proposal ([link](README.md))

Code review:

```json
{
  "attestation_type": "https://example.com/CodeReview/v1",
  "subject": { "git_commit_id": "859b387b985ea0f414e4e8099c9f874acb217b94" },
  "details": {
    "timestamp": "2020-04-12T13:50:00Z",
    "repo_type": "git",
    "repo_url": "https://github.com/my-company/my-product",
    "repo_branch": "master",
    "code_reviewed": true
  }
}
```

Provenance:

```json
{
  "attestation_type": "https://example.com/GitHubActionProduct/v1",
[...]

PR #156, opened 2023-03-16
Remove schema ideas file and store as an issue instead

pxp928:
Building on the work done by the other maintainers, we should remove future schema ideas from `docs` and add them to an issue to avoid confusion. I have created https://github.com/in-toto/attestation/issues/155 to capture this.


TomHennen (approved):
LGTM, thanks!

PR #162, opened 2023-03-18
Add proto and example for ResourceDescriptor

TomHennen:
Also updates go protos.

We're supposed to create these anytime we update protos. It looks like the vsa one was supposed to move when we put it in a subfolder, but we missed it.  Moving it now.

Example output

```jsonl
{
  "name": "theName",
  "uri": "http://example.com",
  "digest": {
    "sha256": "abc123"
  },
  "content": "Ynl0ZXNjb250ZW50",
  "downloadLocation": "http://example.com/test.zip",
  "mediaType": "theMediaType",
  "annotations": {
    "a1": {
      "keyNum": 13,
      "keyStr": "value1"
    },
    "a2": {
      "keyObj": {
        "subKey": "subVal"
      }
    }
  }
}
```


go/example/main.go
```diff
@@ -49,6 +49,31 @@ func createVsa(subName string, subSha256 string, vsaBody *ppb.VerificationSummar
 	return createStatementPb(subName, subSha256, "https://slsa.dev/verification_summary/v0.2", vsaStruct), nil
 }
 
+func testResourceDescriptor() (*spb.ResourceDescriptor, error) {
```
marcelamelara, line 52: Nit: Rename to `createResourceDescriptorPb` to match.

spec/v1.0/resource_descriptor.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
+package github.intoto.attestation.spec.v1_0;
```
marcelamelara, line 3: Q: From a package name standpoint, I've been wondering if we would want to simplify it a bit to something like `intoto.attestation.v1_0`? This way, folks could import `intoto.attestation.v1_0.Statement` and `intoto.attestation.v1_0.predicates.VerificationSummary` directly. Seems slightly more intuitive... Just a question, not sure if this would break any proto conventions.

marcelamelara (commented):
Thank you @TomHennen ! I mostly have some nits.

spec/v1.0/resource_descriptor.proto
TomHennen: That makes sense but it is a bigger change (since we'd want to change the other protos too).  Perhaps it could be a separate PR?

spec/v1.0/resource_descriptor.proto
```diff
@@ -0,0 +1,26 @@
...
+
+import "google/protobuf/struct.proto";
+
+option go_package = "github.com/in-toto/attestation/go/spec/v1.0";
```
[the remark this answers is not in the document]
TomHennen: From what I can tell the go package does double-duty.  The convention seems to be that it's just a single word here (we'd have `attestation`) or something like that, but then users would still use github.com/... to import it in their code because that's how they tell Go where to find it?

I can try making that change, but it would be larger and should probably be in a separate PR.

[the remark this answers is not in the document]
marcelamelara: I just went to see how others do it, and you're right that the `go_package` should point to the download location of the .go files, so we probably keep what we have.

spec/v1.0/resource_descriptor.proto
marcelamelara: Yes, doing this in a separate PR makes sense.

spec/v1.0/resource_descriptor.proto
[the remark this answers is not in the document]
marcelamelara: Yep, agreed!

spec/v1.0/resource_descriptor.proto
adityasaky: I've raised this point before but I'm a little iffy as we don't control intoto.io. Is that a realistic concern?

adityasaky: We (well the CNCF tbf) control in-toto.io, but not intoto.io. The hyphen's been dropped in the dsse envelope, our java package and so on. I'm very anti-hyphen for all new projects as a result. If we think it's not an issue (and it hasn't been for the other instances so far I guess), we can go ahead. But https://intoto.io is an actual tech company so...

I registered intoto.link a while ago so we'd have one without a hyphen (and because the pun was great). I've also been trying to find out if intoto.dev was registered by someone we know. Perhaps we can continue with io.intoto until we know more? I'll defer to the rest of you, I just wanted to note it.

TomHennen: > But https://intoto.io/ is an actual tech company so...

Oh yikes, maybe I should put it back the way it was?  WDYT @marcelamelara ?

adityasaky: Nope, it's taken.

TomHennen: Maybe we should use `_` ? https://stackoverflow.com/questions/63225957/how-to-allow-minus-sign-for-package-name-in-proto-file

marcelamelara: Oh ha... good to know! I'd be fine with using an underscore. Would `in_toto.attestation.` work as a package prefix for the protos then? `in_toto` is also the Python package name I believe...

adityasaky: I'm fine with `_` if that's not expressly disallowed!

TomHennen: Done.  It is supported and everything seems to work.

marcelamelara (approved):

pxp928 (approved):
LGTM

PR #163, opened 2023-03-18
More motivating use case cleanup.

TomHennen:
relations should be a member of the 'predicate'.
Note that 'attestation' in the pseudocode is a Statement.
git_commit -> gitCommit


pxp928 (approved):
LGTM

PR #164, opened 2023-03-21
Some more docs cleanup

marcelamelara:
Making small changes to the docs related to protos and the DSSE envelope.

### DIFF
--- a/docs/protos.md
+++ b/docs/protos.md
@@ -1,7 +1,32 @@
-# Statement protobuf definitions
+# Spec protobuf definitions
 
-Protobuf definitions for the in-toto statement type and some predicates are provided in the
-spec/ directory.  Pre-generated Go implementation of those protos are available in the go dir.
+Protobuf definitions for the in-toto spec and some predicates are
+provided in the spec/ directory.  Pre-generated Go implementation of those
+protos are available in the go/ directory.
+
+## Pre-requisites
+
+On an Ubuntu-based system, install the following dependencies.
+
+```shell
+sudo apt install protobuf-compiler golang
+```
+
+## Regenerating Go proto libraries
+
+[It's typical to keep generated Go code in the repository itself](https://go.dev/doc/articles/go_command#:~:text=and%20then%20check%20those%20generated%20source%20files%20into%20your%20repository)
+since it makes users' lives much easier.
+
+Proto libraries should be regenerated & commited after any change to the
+proto files:
+
+```shell
+$ make go_protos
+$ git commit -asm "update protos"
+...
+```
+
+## Run the Go example
 
 go/example/main.go provides an example of how these protos can be used.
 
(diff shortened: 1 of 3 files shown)


docs/protos.md
```diff
@@ -1,33 +1,40 @@
...
+go/example/main.go provides an example of how these protos can be used.
+
+To try it:
+
+```shell
+$ make run
+...
+Read statement with predicateType https://example.com/unknownPred2
+Predicate fields:{key:"foo"  value:{struct_value:{fields:{key:"bar"  value:{string_value:"baz"}}}}}
+```
```
TomHennen, line 40: nit: should there be a newline at the end of this?  Is there already and GitHub is hiding it?

spec/predicates/README.md
```diff
@@ -12,17 +12,20 @@ community! Please see our [New Predicate Guidelines].
 This directory contains predicate specification types that have gone through
 our [vetting process], and may be of general interest:
 
--   [SLSA Provenance]: To describe the origins of a software artifact.
+-   [SLSA Provenance]: To describe the build process of a software artifact.
```
TomHennen, line 15: Not sure I agree with this change.  The provenance does also describe the source location. Also, the definition of provenance includes 'origin'.

Alternatively perhaps we can reword [what SLSA says about it](https://slsa.dev/provenance/v1-rc1#:~:text=Describe%20how%20an%20artifact%20or%20set%20of%20artifacts%20was%20produced%20so%20that%3A) -> "Describes how an artifact or set of artifacts was produced"

TomHennen, line 15: Not sure I agree with this change.  The provenance does also describe the source location. Also, the definition of provenance includes 'origin'.

Alternatively perhaps we can reword [what SLSA says about it](https://slsa.dev/provenance/v1-rc1#:~:text=Describe%20how%20an%20artifact%20or%20set%20of%20artifacts%20was%20produced%20so%20that%3A) -> "Describes how an artifact or set of artifacts was produced"

spec/v1.0/envelope.md
```diff
@@ -1,38 +1,12 @@
...
-[DSSE] and adopted by in-toto in [ITE-5].
+authentication and serialization.
 
-## Schema
+The format and protocol are defined per [DSSE v1.0]. In the in-toto
+Attestation Framework, the DSSE payload is a base64-encoded JSON [Statement].
 
-```jsonc
-{
-  "payloadType": "application/vnd.in-toto+json",
```
TomHennen, line 13: I think it's fine to refer people to the DSSE spec for the format, but what this change misses is that we no longer specify what the payloadType for an in-toto DSSE should be.  I.e. we probably do want to say

```
> Identifier for the encoding of the payload. Always
> `application/vnd.in-toto+json`, which indicates that it is a JSON object
> with a `_type` field indicating its schema.
```

spec/predicates/README.md
marcelamelara: I was trying to reflect the current SLSA v1 spec focus on build, but I see what you're saying. Will update.

docs/protos.md
adityasaky: IIRC github highlights the absence of a newline with a red minus in a circle or something like that? It's a little annoying because this view is not in fact the source view after all..

adityasaky: image
here's an example

TomHennen (approved):
Thanks!

pxp928 (approved):
LGTM. Thanks

PR #169, opened 2023-03-27
Remove extension fields from versioning.md

### LINKED ISSUE #167: versioning.md#extension-fields still talks about using TypeUri
Even though in v1 we went with something less prescriptive and advise against things like `.` in the names.

Link: https://github.com/in-toto/attestation/blob/main/spec/versioning.md#extension-fields

TomHennen:
The individual versions already indicate how extension fields can be added, and it should be possible to change that method with various versions.  Having versioning.md include that logic also could lead to confusion and disagreement.

fixes #167


marcelamelara (approved):
Thanks! LGTM.

PR #172, opened 2023-03-28
Add python bindings for protobufs

### LINKED ISSUE #171: Include generated python bindings for protobufs
In an effort to keep all of the different language implementations of `in-toto` consistent within the ITE-6 spec, a strategy we can use is to have each language client reference the Protobufs generated within this repository (Note: Go bindings have already started to [take shape](https://github.com/in-toto/attestation/blob/main/go/spec/v1.0/statement.pb.go)). For the [Python client](https://github.com/in-toto/in-toto) of `in-toto`, we can continue down this path laid by the Go bindings ([impetus](https://github.com/in-toto/in-toto/pull/570)). 

Goal: 

Implement Protobuf Python bindings for the [protobufs in the V1.0 spec](https://github.com/in-toto/attestation/blob/main/spec/v1.0/).

chasen-bettinger:
- Add command to generate python bindings in Makefile.
- Included the generated python bindings that generated from that command in the Makefile.

This closes #171 .


marcelamelara:
Thanks for this PR @chasen-bettinger ! I ran the `protoc --python_out` command on my local machine, and saw the same `v1/0/` directory structure, so I'm sure it's not something on your end... I suspect we might have the same problem when we create the bindings for other languages. 

We definitely don't want a hack. I'm thinking we might have to change the directory structure for the protobufs. @TomHennen any ideas?

marcelamelara (commented):
Thanks! I'm also realizing a few other obstacles. (1) The `__init__.py` files are needed at a minimum to make this a proper python package. We probably want a `setup.py` file to actually allow installs. (2) The package base name `io.in_toto.attestation` isn't preserved for the Python bindings right now. I think that's related to our current directory structure (per https://github.com/in-toto/attestation/pull/172#issuecomment-1487030140).

TomHennen (changes requested):
Thanks for doing this!

I think we definitely need the Makefile rule.

As for the generated code, we had checked in the go stubs because it's pretty standard for go code to check that stuff in directly (and it's how go can find the stuff IIUC).

For other languages (like Python) I wonder if it would be preferable to publish a package to PyPI?

Consider this optional.  I'll defer to others.

adityasaky:
> For other languages (like Python) I wonder if it would be preferable to publish a package to PyPI?

I think we will want to but we have to be mindful of the interplay with https://pypi.org/project/in-toto/. The other alternative is to store generated py files in in-toto/in-toto but I don't like the sound of that...

adityasaky:
Perhaps an in-toto-attestations package in-toto/in-toto uses as a dependency is fine. @lukpueh wdyt?

marcelamelara:
> > For other languages (like Python) I wonder if it would be preferable to publish a package to PyPI?
> 
> I think we will want to but we have to be mindful of the interplay with https://pypi.org/project/in-toto/. The other alternative is to store generated py files in in-toto/in-toto but I don't like the sound of that...

I was thinking about this too. Would it make sense to have an `in_toto.attestation` sub-package that is pushed to PyPI? I don't know enough about python packaging to know if this would cause name collisions, but the attestation sub-package does reasonably belong in the `in_toto` namespace imo.

marcelamelara:
FWIW, the sources for the [protobuf](https://github.com/protocolbuffers/protobuf/tree/main/python) package live in the same repo as the protos. I wonder if we could have a setup similar to them. 

Also, to have a consistent package structure for a language bindings in the future, I recommend we move the protos to a separate directory structure: `protos/in_toto/attestation/` and `protos/in_toto/attestation/predicates`. WDYT?

adityasaky:
> I was thinking about this too. Would it make sense to have an in_toto.attestation sub-package that is pushed to PyPI? I don't know enough about python packaging to know if this would work as intended, but the attestation sub-package does reasonably belong in the in_toto namespace imo.

Note: I'm not necessarily proficient with Python packaging.

My understanding is PyPI doesn't support namespaces like NPM does. To have it be `in_toto.attestation`, the generated code would need to live _in_ in-toto/in-toto and uploaded with releases of that implementation, I believe. We could just release a separate package built from this repository called `in-toto-attestations`. Then, `in_toto.attestation` could `import *`? Not sure what the benefits of doing that are as folks are more likely to directly use `in-toto-attestations` anyway...

IMO, we should make `in-toto-attestations` the home of all models. in-toto/in-toto then uses that and defines user friendly workflows like `in-toto-run` and `in-toto-verify` to interact with metadata.

marcelamelara:
> My understanding is PyPI doesn't support namespaces like NPM does. To have it be in_toto.attestation, the generated code would need to live in in-toto/in-toto and uploaded with releases of that implementation, I believe.

I feared as much.

> IMO, we should make in-toto-attestations the home of all models.

Completely agree.  And it looks like we don't get much of a choice to release them in a standalone package, I'm coming around to the idea that it might make more sense to do this anyway.

python/v1/0/statement_pb2.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+from google.protobuf import struct_pb2 as google_dot_protobuf_dot_struct__pb2
+
+
+DESCRIPTOR = _descriptor.FileDescriptor(
+  name='v1.0/statement.proto',
+  package='io.in_toto.attestation.v1_0',
+  syntax='proto3',
+  serialized_options=b'Z+github.com/in-toto/attestation/go/spec/v1.0',
```
adityasaky, line 21: As someone who isn't very experienced with protobuf, is this supposed to point into the go package? Is this some default carrying over?

TomHennen: FWIW all this code is autogenerated by protoc, so who knows?

marcelamelara:
@chasen-bettinger We just merged #180. Could you please rebase your branch and regenerate the python bindings given the new directory structure? If you could please update the Makefiles (there's a top-level and one in protos now), that would be really helpful!

chasen-bettinger:
@marcelamelara Thanks for the update. Ready for your re-review!

Makefile
```diff
@@ -7,4 +7,4 @@ protos: go_setup
 go_run:
 	go run examples/go/main.go
 
-.PHONY: protos go_setup go_run
```
adityasaky, line 10: Let's leave a trailing newline. I suspect this PR doesn't need to modify this file at all unless we're adding a python example.

protos/Makefile
```diff
@@ -2,4 +2,9 @@
 go:
 	protoc --go_out=../go --go_opt=module=github.com/in-toto/attestation/go $(shell find ./in_toto_attestation -name "*.proto")
 
+# Creates the python libs in the python/ directory.
+python: 
+	mkdir -p ../python
```
adityasaky, line 7: For consistency, let's take this out? If the dir is made the first time, that likely suffices.

python/v1/statement_pb2.py
adityasaky, line 1: To be clear, a dependent would import this as `from in_toto_attestation.v1.statement_pb2 import Statement`, yes?

Makefile
marcelamelara: +1 to leaving the trailing newline. But I do think this PR should add a `make -C protos python` command under the `protos` target.

python/v1/statement_pb2.py
marcelamelara: yes, I believe we need the `in_toto_attestation` directory as part of the python package directory hierarchy.

protos/Makefile
```diff
@@ -2,4 +2,9 @@
...
+	protoc -I ./in_toto_attestation --python_out=../python $(shell find ./in_toto_attestation -name "*.proto")
```
marcelamelara, line 8: I don't think we need the `-I` flag for this one. This will possibly also resolve not having the `in_toto_attestation` directory as part of the package. 

```suggestion
	protoc --python_out=../python $(shell find ./in_toto_attestation -name "*.proto")
```

protos/Makefile
```diff
@@ -2,4 +2,9 @@
...
+
 .PHONY: go
```
marcelamelara, line 10: Add `python` target to .PHONY.

marcelamelara (changes requested):
Thanks @chasen-bettinger ! I have a few minors changes.

Makefile
adityasaky: Ah I forgot that target was in this file, my bad.

protos/Makefile
chasen-bettinger: @marcelamelara I actually included this to maintain consistency with the Go bindings. If we exclude this option then the bindings are placed in a child directory of `in_toto_attestation`.

e.g.

- python
  - in_toto_attestion
    - predicates
    - v1

chasen-bettinger: @marcelamelara What is your preference?

chasen-bettinger: @marcelamelara I see this comment in a later review 

`yes, I believe we need the in_toto_attestation directory as part of the python package directory hierarchy.` so I will operate under the assumption that you are expecting the directory to be present.

chasen-bettinger:
@marcelamelara Tag, you're it!

marcelamelara (approved):
Thanks @chasen-bettinger !! LGTM :) This sets us up really nicely to get the python package released!

protos/Makefile
marcelamelara: Yes, I think having that directory will make creating the python package a bit easier. Thanks again for making the changes!

adityasaky (approved):
Thanks @chasen-bettinger!

marcelamelara:
Will merge after #192 is merged.

TomHennen (approved):
Thanks!

PR #180, opened 2023-03-29
Move protos to their own directory for more consistency between different language bindings

marcelamelara:
This creates a directory structure for all protos. The main goal is to have a directory structure that mirrors the package name/structure we want the different language bindings to have (per https://github.com/in-toto/attestation/pull/172#issuecomment-1487430341). This also seems like the most sustainable solution to the issue described [here](https://github.com/in-toto/attestation/pull/172#issue-1643070784).

Cons: The proto package name is changing from what we tagged in v1.0! This will be a breaking change in case there's anyone who is already using our protos. Need to discuss how to handle this divergence from our tagged v1.0 package name.

Blocking #172


protos/in_toto_attestation/predicates/vsa/vsa.proto
adityasaky, line 1: Side note: shouldn't we also version predicate protos?

adityasaky (approved):

protos/in_toto_attestation/predicates/vsa/vsa.proto
marcelamelara: This one in particular is versioned in the name `VerificationSummaryV02`, but do you mean in the directory structure?

adityasaky: Yep! That'd then get versioned in the generated code too, right?

TomHennen: IMO having a separate directory for each version of a predicate that is already in a subdir seems annoying.

Maybe it's needed though in case they have subprotos that are defined in other files?

adityasaky: We should probably limit directories using the major version only, perhaps. vsa/v_0, vsa/v_1 and so on. Likely the same for statement too.

adityasaky: Then tooling would support the latest available thing for the major version and because we're enforcing semver, we'd maintain backwards compatibility on patch / minor bumps.

marcelamelara: So did we decide to add major version directories for predicates? I tend to agree that it could be cumbersome, but more importantly, having a `/v0` package seems a bit awkward. I actually think it's fine not to have versioned subdirectories for predicates (we don't in spec/) since predicates are versioned independently.

The other thing to consider is how the different language bindings will be shipped. While having the versioned subdirectories is helpful for the spec, in a future with multiple major versions of the spec, would we want to be shipping old APIs with newer APIs? I don't know if that's common practice.

protos/in_toto_attestation/predicates/vsa/v0_2/vsa.proto
```diff
@@ -1,10 +1,10 @@
 syntax = "proto3";
 
-package io.in_toto.attestation.predicates;
+package in_toto_attestation.predicates;
 
 import "google/protobuf/timestamp.proto";
 
-option go_package = "github.com/in-toto/attestation/go/spec/predicates";
+option go_package = "github.com/in-toto/attestation/go/predicates";
```
adityasaky, line 7:
```suggestion
option go_package = "github.com/in-toto/attestation/go/predicates/vsa/v0";
```

I don't know if we've taken a call on the versioning but we should mark separate packages for each predicate IMO.

adityasaky: If we're versioning by dir, `message VerificationSummaryV02 {` can be changed to drop the version.

marcelamelara: ping @TomHennen for this as this might be what the slsa-verifier expects.

TomHennen: I think @laurentsimon knows best (and is working on VSA features now).  WDYT Laurent?

laurentsimon: I've been waiting for the v1.0 cleanups on VSA to add support in slsa-verifier, so no particular requirement / opinion from me.

TomHennen: In that case I don't believe anyone is depending on this behavior yet.

protos/in_toto_attestation/v1/resource_descriptor.proto
```diff
@@ -1,13 +1,13 @@
 syntax = "proto3";
 
-package io.in_toto.attestation.v1_0;
+package in_toto_attestation.v1;
 
 import "google/protobuf/struct.proto";
 
-option go_package = "github.com/in-toto/attestation/go/spec/v1.0";
+option go_package = "github.com/in-toto/attestation/go/v1";
```
TomHennen, line 7: Why drop the `.0`?

We dropped the `.0` from the type string because any new fields are supposed to be backwards compatible unless or else they need to increment the major number.  However we may want to make backwards compatible changes to these messages, which would rev the minor version, even though the major version and type string would be unchanged.

So I think it's appropriate to keep `.0` here so that we can introduce `1.1` in the future with new fields.

Or maybe, the argument goes, if it really is backwards compatible, then producers of the v1 proto would be unaffected by new fields and could just transparently take the new definition.  Is that right?

docs/protos.md
```diff
@@ -33,8 +36,10 @@ go/example/main.go provides an example of how these protos can be used.
 To try it:
 
 ```shell
-$ make run
+$ make -C protos go_run
```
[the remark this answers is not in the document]
marcelamelara: I wanted the Makefile closer to the protos, but it also seemed like a reasonable place to trigger the tests for the protos. As we add more test suites and language bindings, I imagine we'll want to have a structure to support that; ideally, I'd like to put that in place now. Would it make sense to have a root Makefile from which we can trigger protos and/or test Makefiles? The would also allow us to de-couple the language bindings from the tests, and move all tests into a tests/ directory. WDYT?

protos/in_toto_attestation/v1/resource_descriptor.proto
marcelamelara: > if it really is backwards compatible, then producers of the v1 proto would be unaffected by new fields and could just transparently take the new definition. Is that right?

Yes, this is the intent. I do realize now that we probably want to document basic protos rules to facilitate this: 
* no field re-ordering within a major version
* no field type changes within a major version
* no required fields

Would this work? Also, is there a way to encode the current proto spec version into the definition?

TomHennen: Yeah that makes sense.  Maybe we can just link to this https://protobuf.dev/programming-guides/proto3/#updating ?

> is there a way to encode the current proto spec version into the definition?

What do you mean by proto spec version?

docs/protos.md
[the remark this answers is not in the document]
TomHennen: That all makes sense.  I think timing of the update is up to you.

protos/in_toto_attestation/predicates/vsa/vsa.proto
TomHennen: > I actually think it's fine not to have versioned subdirectories for predicates

This SGTM

> would we want to be shipping old APIs with newer APIs

If we're talking about the various versions of the statements and predicate types then I think this is exactly what we want.  Consumers will likely want to support multiple versions of all of these messages for some time.  They may have multiple producers creating these messages where each producer creates a different version.  So they'll likely want all the message versions available to them?

protos/in_toto_attestation/v1/resource_descriptor.proto
marcelamelara: > What do you mean by proto spec version?

The statement spec is currently at version 1.0, so is there a way to encode/validate this version number in the corresponding proto? I suspect the answer is no.

TomHennen: Oh I see.  That's something I had wanted to use cuelang.org and vet for.  Theoretically you could create cuelang definitions to validate all the rules in the spec.  It would be nice to have them for all predicates.  If your still looking for project ideas for that event you're organizing I meant to add that to the list. :)

protos/in_toto_attestation/predicates/vsa/vsa.proto
marcelamelara: > They may have multiple producers creating these messages where each producer creates a different version. So they'll likely want all the message versions available to them?

Good point. This actually sounds like an argument in favor of having versioned subdirectories for predicates.

docs/protos.md
```diff
@@ -15,13 +15,38 @@ On an Ubuntu-based system, install the following dependencies.
...
 
 [It's typical to keep generated Go code in the repository itself](https://go.dev/doc/articles/go_command#:~:text=and%20then%20check%20those%20generated%20source%20files%20into%20your%20repository)
 since it makes users' lives much easier.
 
 Proto libraries should be regenerated & committed after any change to the
-proto files:
+proto files. From the root directory, run:
 
 ```shell
 $ make protos
```
adityasaky, line 52: this has changed and needs to be updated. I also think it may be better to remove the git commit line, we probably want better commit messages describing the actual changes...

adityasaky: nvm, i missed the double makefiles

adityasaky (commented):
LGTM!

pxp928 (approved):
LGTM. Had a question about the double makefiles but that was already answered in the previous comments.

PR #182, opened 2023-04-04
Add codeowners file

adityasaky:
Not sure about this one. The governance model says codeowners file but IMO we should have a maintainers.md file to contain information beyond github usernames. This one may still be useful for automating review requests and so on. Thoughts?


pxp928 (approved):
LGTM. Good thought.

marcelamelara:
Agreed that it would be helpful to add more info to MAINTAINERS.md at some point, though that lands very low on my priority list right now.

PR #187, opened 2023-04-12
Update SPDX predicate specification

danbev:
This commit contains a suggestion for updating the SPDX predicate specification, to include the version in the `predicateType`.

The motivation for this change is same to the motivation that was mentioned for including the version in the CycloneDX predicate specification which is that it allows a consumer to determine the version without having to parse the predicate itself.

### DIFF
--- a/spec/predicates/spdx.md
+++ b/spec/predicates/spdx.md
@@ -4,6 +4,11 @@ Type URI: https://spdx.dev/Document
 
 Version: 2.3
 
+TODO: Ask SPDX project to choose a URI and to review this spec. Ideally the URI
+would resolve to this file. Also, decide whether we want the version number to
+reflect the spdxVersion (e.g. 2.2) or have them be independent (no version
+number in URI).
+
 ## Purpose
 
 A Software Bill of Materials type following the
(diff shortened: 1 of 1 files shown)


spec/predicates/spdx.md
```diff
@@ -1,47 +1,69 @@
 # Predicate type: SPDX
 
-Type URI: (tentative) https://spdx.dev/Document
+Type URI: https://spdx.dev/Document
```
joshuagl, line 3: Should we somehow indicate that the Type URI MUST include a version?
```suggestion
Type URI: https://spdx.dev/Document/v${SPDX_MAJOR}.${SPDX_MINOR}
```

spec/predicates/spdx.md
```diff
@@ -1,47 +1,69 @@
...
 
-Version: 1.0.0
-
-TODO: Ask SPDX project to choose a URI and to review this spec. Ideally the URI
```
joshuagl, line 7: Perhaps we should leave the TODO suggesting we get someone from the SPDX community to review the spec?

joshuagl (commented):
Thanks for the suggestion, I completely agree with the rationale.

spec/predicates/spdx.md
marcelamelara: I like this idea, though we haven't required this in other predicate definitions either. 

This may be a question beyond the scope of this PR, but it seems like we might want to be clearer in our documentation for predicateTypes.

joshuagl (approved):
LGTM, thank you

PR #190, opened 2023-04-13
Small docs tweaks

### LINKED ISSUE #188: Drop base64 encoding requirement from ResourceDescriptor.contents?
(originally from https://github.com/slsa-framework/slsa/issues/834)

Minor piece of feedback / question:

ResourceDescriptor.contents is currently described as:

> base64-encoded bytes, optional

I think it's reasonable to enforce character class constraints on this field, but do we need to go as far as to say this should always be base64 encoded? I'm thinking about simple key/value pairs contained in SLSA `byproducts` that would use ASCII encoding - requiring base64 around these may be unnecessary. 🤔

The description also says:

> The semantics are up to the producer and consumer. The uri or mediaType MAY be used by the producer as hints for how consumers should parse content.

RFC 6838 already has a section for encoding requirements: https://www.rfc-editor.org/rfc/rfc6838.html#section-4.8. Perhaps this is enough to inform to clients what the encoding expectations are for this field?

Or is the right way to read this "contents need to be base64-compatible" - e.g. if the value itself is a valid base64 string, that's okay? (though this is also tricky because there are different base64 encodings)

marcelamelara:
Committing more tweaks to docs based on feedback.

Fixes #188.

### DIFF
--- a/spec/v1.0/README.md
+++ b/spec/v1.0/README.md
@@ -46,7 +46,7 @@ this model.
     meaning of any other field, and the field MUST follow the monotonic
     principle.
 
--   **Monotonic:** A policy is considered monotonic if ignoring an
+-   **Monotonic principle:** A policy is considered monotonic if ignoring an
     attestation, or a field within an attestation, will never turn a DENY
     decision into an ALLOW. A predicate or field follows the monotonic
     principle if the expected policy that consumes it is monotonic.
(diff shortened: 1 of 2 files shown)


TomHennen:
@wlynch WDYT?  Does this do what you're looking for?

wlynch (approved):

TomHennen (approved):

pxp928 (approved):
LGTM. Thanks for catching that @wlynch

joshuagl (approved):

PR #192, opened 2023-04-14
Autogenerate attestation libraries using protoc

adityasaky:
This PR adds a GitHub Actions workflow to regenerate libraries using protoc. The action is triggered once a week as well as when something is pushed to main.

(We need the scheduled run as protoc is configured to use version `3.x` so we may have newer tooling even if there are no changes here for a week to main.)


.github/workflows/make-protos.yml
```diff
@@ -0,0 +1,39 @@
+name: Generate attestation libraries
+on:
+  push:
+    branches:
+      - "main"
+  schedule:
```
marcelamelara, line 6: Can you please add a comment describing the justification for needing the schedule?

marcelamelara (commented):
Thanks! Do we need to update the documentation in docs/protos.md to remove the explicit requirement to regenerate the language bindings manually?

.github/workflows/make-protos.yml
```diff
@@ -0,0 +1,39 @@
...
+    - cron: '0 14 * * 1'
+jobs:
+  run-make-protos:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install protoc
+        uses: arduino/setup-protoc@ab6203da1c3118e4406048171b09238ad31ad73e
```
TomHennen, line 13: Do we really want the _arduino_ version of protoc?  Seems odd?

.github/workflows/make-protos.yml
TomHennen: +1 to some more comments, even describing what it does a bit more would be nice.  E.g. It just sends the PR, right?  It doesn't auto-merge it?

.github/workflows/make-protos.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: setup env
+        run: |
+          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+      - name: Checkout repo
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
+      - name: Generate libraries
+        run: |
+          make protos
+      - name: Create PR
+        uses: peter-evans/create-pull-request@5b4a9f6a9e2af26e5f02351490b90d01eb8ec1e5
```
TomHennen, line 29: Is it secure to use this (random guy's?) thing for modifying code?  My guess is no one is going to provide much review on these PRs since they're machine generated.

TomHennen (changes requested):
Thank you!

Also, my apologies if my other questions are daft, I've never done GitHub Actions before.

Finally, did we have some other documentation that suggests people regenerate the libs when modifying the protos?  I'd assume that would change now?

.github/workflows/make-protos.yml
adityasaky: It's not the arduino version, rather they're the creators of the setup-protoc action. It's certainly preferable to maintaining our own setup-protoc action. Presumably, they have similar codegen requirements and created the action for their needs?

.github/workflows/make-protos.yml
adityasaky: It's like the arduino thing above, it's a random guy but it's a pretty popular action for the job (https://github.com/peter-evans/create-pull-request/). Note that the action is only creating a commit and PR if the prior step `make protos` resulted in something to commit.

.github/workflows/make-protos.yml
adityasaky: Added comments for the schedule and the create-pull-request step.

adityasaky:
Also amended the protos doc to instruct contributors not to regenerate libraries.

adityasaky:
Updated `go_setup` target in Makefile to be a `go install` of the Go generator. Also removed the explicit versioning so the one in go.mod is used by default.

marcelamelara (approved):
LGTM, thank you!

.github/workflows/make-protos.yml
TomHennen: Got it.  Thanks for the explanation.

TomHennen (approved):

PR #205, opened 2023-04-26
Add Java bindings

adityasaky:


marcelamelara (changes requested):
Thank you @adityasaky ! Could you please add a java/README in [this](https://github.com/in-toto/attestation/blob/f87580431d93fcf57ecf7e5117e3af1159065222/python/README.md) style ?

adityasaky:
Updated with java_package as well. I'm not a Java person per se, so I hope someone can help me figure out what the package structure should look like.

TomHennen:
I thought we didn't like using io.intoto since intoto.io is owned by some other company?  Can you use 'io.in-toto' or 'io.in_toto'?

adityasaky:
True, but I believe there are best practices concerns as well. I took a look at in-toto-java along the way. @alos perhaps you could help out too?

marcelamelara:
+1 to Tom's comment about the io.intoto package namespace. Looks like in-toto-java uses the `io.github.intoto` namespace (per [these docs](https://github.com/in-toto/in-toto-java#installation))?

adityasaky:
Updated the package name.

marcelamelara (approved):
Thanks for the updates! I have one more docs-related request: Can you please update protos/README.md to add java?

TomHennen (approved):
Thanks!

PR #211, opened 2023-05-03
Allow uses of ResourceDescriptor to require fields

adityasaky:
Currently, the `ResourceDescriptor` only mandates one of `uri`, `digest`, or `content`. However, some predicates (links for example) and/or the subject field in the Statement should be able to mandate one or more fields separately.


MarkLodato (approved):

marcelamelara (approved):
Thanks @adityasaky ! LGTM.

TomHennen (approved):
Thanks!

PR #212, opened 2023-05-03
Use ResourceDescriptor for statement subjects

adityasaky:
Updates statement.proto to use ResourceDescriptor which changes the number of `digest` from 2 to 3.

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -34,17 +34,16 @@ Steps:
     -   `statement` := decode `envelope.payload` as a JSON-encoded
         [Statement]; reject if decoding fails
     -   Reject if `statement.type` != `https://in-toto.io/Statement/v1`
-    -   `artifactNames` := empty set of names
-    -   For each `s` in `statement.subject`:
-        -   For each digest (`alg`, `value`) in `s.digest`:
-            -   If `alg` is in `acceptableDigestAlgorithms`:
-                -   If `hash(alg, artifactToVerify)` == `hexDecode(value)`:
-                    -   Add `s.name` to `artifactNames`
-    -   Reject if `artifactNames` is empty
+    -   `matchedSubjects` := the subset of entries `s` in `statement.subject`
+        where:
+        -   there exists at least one `(alg, value)` in `s.digest` where:
+            -   `alg` is in `acceptableDigestAlgorithms` AND
+            -   `hash(alg, artifactToVerify)` == `hexDecode(value)`
+    -   Reject if `matchedSubjects` is empty
 
 Output (to be fed into policy engine):
 
 -   `statement.predicateType`
 -   `statement.predicate`
--   `artifactNames`
+-   `matchedSubjects`
 -   `attesterNames`
--- a/protos/in_toto_attestation/v1/statement.proto
+++ b/protos/in_toto_attestation/v1/statement.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 package in_toto_attestation.v1;
 
 import "google/protobuf/struct.proto";
+import "in_toto_attestation/v1/resource_descriptor.proto";
 
 
 option go_package = "github.com/in-toto/attestation/go/v1";
@@ -15,12 +16,7 @@ message Statement {
   // Expected to always be "https://in-toto.io/Statement/v1"
   string type = 1 [json_name = "_type"];
 
-  message Subject {
-    string name = 1;
-    map<string, string> digest = 2;
-  }
-
-  repeated Subject subject = 2;
+  repeated in_toto_attestation.v1.ResourceDescriptor subject = 2;
 
   string predicateType = 3;
 
--- a/spec/v1/statement.md
+++ b/spec/v1/statement.md
@@ -33,36 +33,26 @@ Additional [parsing rules] apply.
 > Identifier for the schema of the Statement. Always
 > `https://in-toto.io/Statement/v1` for this version of the spec.
 
-`subject` _array of objects, required_
+`subject` _array of [ResourceDescriptor] objects, required_
 
 > Set of software artifacts that the attestation applies to. Each element
-> represents a single software artifact.
+> represents a single software artifact. Each element MUST have `digest` set.
+>
+> The `name` field may be used as an identifier to distinguish this artifact
+> from others within the `subject`. Similarly, other ResourceDescriptor fields
+> may be used as required by the context. The semantics are up to the producer
+> and consumer and they MAY use them when evaluating policy. If the name is not
+> meaningful, leave the field unset or use "\_". For example, a
+> [SLSA Provenance] attestation might use the name to specify output filename,
+> expecting the consumer to only consider entries with a particular name.
+> Alternatively, a vulnerability scan attestation might leave name unset because
+> the results apply regardless of what the artifact is named.
+>
+> If set, `name` and `uri` SHOULD be unique within subject.
 >
 > IMPORTANT: Subject artifacts are matched purely by digest, regardless of
 > content type. If this matters to you, please comment on
 > [GitHub Issue #28](https://github.com/in-toto/attestation/issues/28)
-
-`subject[*].digest` _object ([DigestSet]), required_
-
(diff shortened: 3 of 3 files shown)


spec/v1.0/statement.md
```diff
@@ -33,37 +33,24 @@ Additional [parsing rules] apply.
 > Identifier for the schema of the Statement. Always
 > `https://in-toto.io/Statement/v1` for this version of the spec.
 
-`subject` _array of objects, required_
+`subject` _array of [ResourceDescriptor] objects, required_
 
 > Set of software artifacts that the attestation applies to. Each element
-> represents a single software artifact.
+> represents a single software artifact. Each object MUST have `digest` set.
+> The `name` field may be used as an identifier to distinguish this artifact
```
MarkLodato, line 40: Increase readability by starting a new paragraph here.

```suggestion
>
> The `name` field may be used as an identifier to distinguish this artifact
```

spec/v1.0/statement.md
```diff
@@ -33,37 +33,24 @@ Additional [parsing rules] apply.
...
+> from others within the `subject`. The semantics are up to the producer and
+> consumer and they MAY use it when evaluating policy. If the name is not
+> meaningful, leave the field unset or use "\_". For example, a
+> [SLSA Provenance] attestation might use the name to specify output filename,
+> expecting the consumer to only consider entries with a particular name.
+> Alternatively, a vulnerability scan attestation might leave name unset because
+> the results apply regardless of what the artifact is named. If set, `name`
+> SHOULD be unique within subject.
```
adityasaky, line 48: Is there a reason the uniqueness of `name` is only a SHOULD?

spec/v1.0/statement.md
```diff
@@ -33,37 +33,24 @@ Additional [parsing rules] apply.
...
 > content type. If this matters to you, please comment on
 > [GitHub Issue #28](https://github.com/in-toto/attestation/issues/28)
 
-`subject[*].digest` _object ([DigestSet]), required_
-
-> Collection of cryptographic digests for the contents of this artifact.
->
-> Two DigestSets are considered matching if ANY of the fields match. The
-> producer and consumer must agree on acceptable algorithms. If there are no
-> overlapping algorithms, the subject is considered not matching.
```
MarkLodato, line 51: We need to describe and maintain the semantics of what an attestation is bound to. Currently it is definitely the `digest` (using the matching rules described here) and optionally `name` if the producer and consumer agree on it. Since we're adding more fields, I think basically we'd be saying that it MUST match on `digest` and MAY match on other fields if the producer and consumer agree on that?

spec/v1.0/statement.md
MarkLodato: Because we don't have a reason to make be a MUST 😜 . What is the rationale for making the file invalid if two names are the same?

/cc @TomHennen

adityasaky: I was thinking about it from the perspective of policies that rely on the `name` field. Taken with #211, consider the link predicate. The subject field is essentially the products field in the old scheme. To apply in-toto's artifact rules, we rely on `name` before using the `digests`. In the original schema, the verifier is only going to get one entry for a `name` as it uses a map but in the new model, we'd have to separately enforce uniqueness of `name`.

I think this ties into a broader question. Can predicates set expectations on the `Statement` wrapping them? If a link predicate can mandate that `digest` is compulsory for its materials field (within the predicate), can it do the same for unique `name` for the subjects?

spec/v1.0/statement.md
adityasaky: I don't know that the rule belongs here actually. We have the matching rules for digest objects in the `DigestSet` definition. As for how to evaluate if the resource descriptor object as a whole should be matched, it's perhaps context specific? In some cases we may be matching on name and allowing digest to change (eg: `MODIFY foo/*`).

spec/v1.0/statement.md
TomHennen: >  Can predicates set expectations on the Statement wrapping them?

This seems useful, and a good compromise?

adityasaky: I think it could work indeed.

Validation with something like https://www.kusari.dev/blog/spector-announcement/ could get more complicated, though. Especially for conditions like unique `name` across all subjects rather than just requiring some field in every subject entry. cc @mlieberman85, @pxp928

protos/in_toto_attestation/v1/statement.proto
```diff
@@ -15,12 +16,7 @@ message Statement {
   // Expected to always be "https://in-toto.io/Statement/v1"
   string type = 1 [json_name = "_type"];
 
-  message Subject {
-    string name = 1;
-    map<string, string> digest = 2;
-  }
-
-  repeated Subject subject = 2;
+  repeated in_toto_attestation.v1.ResourceDescriptor subject = 2;
```
adityasaky, line 19: We cannot set expectations here about the required `digest` field. Validation has to occur separately, I'm again thinking of something like https://www.kusari.dev/blog/spector-announcement/.

adityasaky: TBF the old definition wasn't setting that as mandatory either.

spec/v1.0/statement.md
MarkLodato: Oops, I missed that and also the "subject artifacts are matched purely by digest". Please disregard this comment.

spec/v1.0/statement.md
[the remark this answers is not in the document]
adityasaky:
```
    -   `matchedSubjects` := empty set of subject entries
    -   For each `s` in `statement.subject`:
        -   For each digest (`alg`, `value`) in `s.digest`:
            -   If `alg` is in `acceptableDigestAlgorithms`:
                -   If `hash(alg, artifactToVerify)` == `hexDecode(value)`:
                    -   Add `s` to `matchedSubjects`
                    -   Break
    -   Reject if `matchedSubjects` is empty

Output (to be fed into policy engine):

-   `statement.predicateType`
-   `statement.predicate`
-   `matchedSubjects`
-   `attesterNames`
```

Went with this for a smaller diff, is that ok?

spec/v1.0/statement.md
adityasaky: I'm going to resolve this, I think the current consensus is to let predicates specify required fields. It's probably a breaking change to require `name` in the spec? :thinking:

docs/validation.md
```diff
@@ -34,17 +34,18 @@ Steps:
...
-    -   `artifactNames` := empty set of names
+    -   `matchedSubjects` := empty set of subject entries
     -   For each `s` in `statement.subject`:
         -   For each digest (`alg`, `value`) in `s.digest`:
             -   If `alg` is in `acceptableDigestAlgorithms`:
                 -   If `hash(alg, artifactToVerify)` == `hexDecode(value)`:
-                    -   Add `s.name` to `artifactNames`
-    -   Reject if `artifactNames` is empty
+                    -   Add `s` to `matchedSubjects`
+                    -   Break
```
MarkLodato, line 43: remove; more than one entry can match

spec/v1.0/statement.md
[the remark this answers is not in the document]
MarkLodato: Yup.

MarkLodato (approved):

docs/validation.md
adityasaky: This breaks on hash check, not entries.

MarkLodato: It's too confusing since there are multiple loops, and this is just pseudocode anyway. The logic is correct without the break, so I suggest omitting it.

In a future PR you might want to implement my suggestion to simplify the wording.

adityasaky: I see what you mean. I'll update the wording now.

adityasaky: Done, thanks!

MarkLodato (approved):

TomHennen (approved):
Thanks for this work!

marcelamelara (approved):
Thank you @adityasaky ! I believe this closes #5 as well!

PR #214, opened 2023-05-03
Use only major version for spec docs

adityasaky:
This change makes the repository consistent with our versioning for protos / generated code. Readers should use tagged versions so that the spec can be actively developed in the v1 namespace.


marcelamelara (approved):
LGTM, thank you @adityasaky ! Wondering if this is a good time to move the parsing rules out of the /v1 directory and make them version-agnostic.

TomHennen (commented):
I get only using the major version in the attestations (since they're supposed to be backwards compatible) but does that make sense for the documentation that needs to track changes over time?

marcelamelara:
It's a good question. Though to track changes over time (beyond the commits) I feel like we should've created a v1.1-draft directory right after cutting the v1.0 release, right? On the other hand, I think it might be enough to signal a transition in version using the tagged releases, which I think is what @adityasaky is suggesting.

Perhaps a happy medium would be to add a "What's new" or CHANGELOG section at the bottom of the spec documents within a major version. We already do this for the predicates when they are updated.

TomHennen:
That sounds good to me.

TomHennen:
Do we actually need to change anything in this PR?  Maybe not and I can just approve/merge?

marcelamelara:
Yeah, I think this is good to go.

TomHennen (approved):

PR #220, opened 2023-05-10
Add Golang wrapper APIs and tests

### LINKED ISSUE #175: Create test suite for Go bindings
Currently, the [Go bindings](https://github.com/in-toto/attestation/blob/main/go) have a single example application. 
Ideally, the test suite would test the Go bindings for the [attestation layers](https://github.com/in-toto/attestation/blob/main/go/spec/v1.0/) and [predicates](https://github.com/in-toto/attestation/blob/main/go/spec/predicates) in a more modular way.

Tasks:
- [x] Add individual test files for each Go object
- [x] Update the Makefile to invoke the test suite via `go test`
- [x] Update the docs/protos.md to include instructions for running the test suite

marcelamelara:
This PR introduces basic wrapper APIs for the Golang language bindings for the Statement layer. I also adds basic tests and documentation for testing the Golang language bindings.

Fixes #175


go/v1/statement.go
```diff
@@ -0,0 +1,57 @@
...
+	
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func NewStatementPb(subject []*ResourceDescriptor,
+	predicateType string,
+	predicate *structpb.Struct) (*Statement, error) {
+	
+	s := &Statement{
+		Type:          "https://in-toto.io/Statement/v1",
```
adityasaky, line 18: Can this be a `const` in this package?

go/v1/statement.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if !IsValidStatement(s) {
+		return nil, errors.New("malformed Statement")
+	}
+	
+	return s, nil
+}
+
+func IsValidStatement(s *Statement) bool {
+	if s.GetType() != "https://in-toto.io/Statement/v1" {
```
adityasaky, line 32: `const` again here.

go/v1/statement.go
```diff
@@ -0,0 +1,57 @@
...
+	if s.GetSubject() == nil || len(s.GetSubject()) == 0 {
+		return false
+	}
+	
+	// check all resource descriptors in the subject
+	subject := s.GetSubject()
+	for _, rd := range subject {
+		if !IsValidResourceDescriptor(rd) {
+			return false
+		}
```
adityasaky, line 45: There should be a statement specific check here that `digest` is set, actually.

go/v1/resource_descriptor_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+package v1
+
+import (
+	"testing"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
+	"github.com/stretchr/testify/assert"
```
adityasaky, line 13: can you run `go mod tidy`? I'm surprised this wasn't added to `go.mod` and `go.sum`.

adityasaky (commented):
Nice! Some general comments / thoughts.

go/v1/resource_descriptor_test.go
adityasaky:
```
go test ./...
# github.com/in-toto/attestation/go/v1
go/v1/resource_descriptor_test.go:13:2: no required module provides package github.com/stretchr/testify/assert; to add it:
        go get github.com/stretchr/testify/assert
FAIL    github.com/in-toto/attestation/go/v1 [setup failed]
?       github.com/in-toto/attestation/examples/go      [no test files]
?       github.com/in-toto/attestation/go/predicates/link/v0    [no test files]
?       github.com/in-toto/attestation/go/predicates/vsa/v0     [no test files]
FAIL
make: *** [Makefile:13: go_test] Error 1
```

adityasaky: I suggest adding a github workflow that runs `make go_test` as well.

marcelamelara: Ah, yes, I'll push the updated go.mod and go.sum.

go/v1/resource_descriptor.go
```diff
@@ -0,0 +1,42 @@
+/*
+ * Wrapper APIs for in-toto attestation ResourceDescriptor protos.
+ */
```
adityasaky, line 3:
```suggestion
*/
```

go/v1/resource_descriptor_test.go
```diff
@@ -0,0 +1,62 @@
+/*
+ * Tests for in-toto attestation ResourceDescriptor protos.
+ */
```
adityasaky, line 3:
```suggestion
*/
```

go/v1/statement.go
```diff
@@ -0,0 +1,57 @@
+/*
+ * Wrapper APIs for in-toto attestation Statement layer protos.
+ */
```
adityasaky, line 3:
```suggestion
*/
```

go/v1/statement_test.go
```diff
@@ -0,0 +1,113 @@
+/*
+ * Tests for in-toto attestation ResourceDescriptor protos.
+ */
```
adityasaky, line 3:
```suggestion
*/
```

go/v1/resource_descriptor_test.go
marcelamelara: Done.

go/v1/statement.go
marcelamelara: Done.

go/v1/statement.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"errors"
+	
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+const StatementTypeUri = "https://in-toto.io/Statement/v1"
+
+func NewStatementPb(subject []*ResourceDescriptor,
```
migmartri, line 15: Thanks @marcelamelara for working on making our life (library consumers) easier, I really appreciate it :)

My two cents, but please take them with a grain of salt, I do not have all the context of your vision on how these wrappers might evolve over time.

That said, I have mixed feelings about the usefulness of these "constructors". Apart of the validation part, which I like, creating this additional piece of indirection might result in some maintenance burden. i.e if the statement schema changes, modifications on the function signature might be required.

I'd argue too that this function doesn't simplify much the life of a `caller`, In the end is limiting itself to assemble in this case a statement directly from the function args, and `structpb.Struct`s are quite code intensive to assemble in my opinion. 

What I really like though is the validation logic, and this is something I'd like to use from Chainloop itself.

So I am wondering if as an alternative, we could just extend the generated proto message with a `validation function`. That way, I, as a consumer of these proto bindings, could decide how to assemble the `ResourceDescriptor` or `Statement` while still be able to leverage this validation logic.  

FWIW; I do something similar [here](https://github.com/chainloop-dev/chainloop/blob/ac7d2f5df369a6000185bf17a49f86557b6c8b1c/app/controlplane/api/workflowcontract/v1/crafting_schema_validations.go#L24). Where a protoc generated message (`v1.CraftingSchema` ) is "extended".

In any case, in the current state, this PR LGTM!

Thank you!

go/v1/statement_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func TestJsonUnmarshalStatement(t *testing.T) {	
+	got := &Statement{}
+	err := protojson.Unmarshal([]byte(wantSt), got)
+
+	assert.Nil(t, err, "Error during JSON unmarshalling")
+
+	want, err := createTestStatement()
+
+	assert.Nil(t, err, "Error during test Statement creation")
```
migmartri, line 45: nitpick, you might be able to use

`assert.Nil` => `assert.NoError`  https://pkg.go.dev/github.com/stretchr/testify/assert#NoError

go/v1/statement.go
marcelamelara: @migmartri Thank you for your comments! I must admit I'm not a very proficient Go programmer, so some of the oddities in my implementation might stem from that.

For simple Statements/RDs, I think the constructors can be convenient shorthand; cases needing more customizability can still always build the messages directly (as we do in many of the tests). I think the issue of added fields in messages requiring changes to the function signature is much more likely to occur for predicates, as we don't expect the Statement layer to change very often, especially within the same major version.

That said, we (the attestation maintainers) have decided it may be best for the in-toto-golang implementation to host any wrapper APIs, so we will leave that aspect up to the maintainers there. I really like your suggestion to extend the pre-generated message (I tried to find whether this is possible, and couldn't find any info), so I will likely implement your suggestion instead.

> structpb.Structs are quite code intensive to assemble in my opinion.

Don't disagree... what do you suggest as an alternative? This seems like a general problem with using structpb.Struct rather than this implementation. We didn't go with anypb in this case because we do want Struct fields to map to JSON Objects, and the Statement cannot dictate the format of the Struct.

go/v1/statement_test.go
marcelamelara: Yep, thanks!

go/v1/statement.go
migmartri: > Thank you for your comments! I must admit I'm not a very proficient Go programmer, so some of the oddities in my implementation might stem from that.

Hi! No problem, happy to help. And don't worry, I am fairly new with Go too :)

> I think the issue of added fields in messages requiring changes to the function signature is much more likely to occur for predicates, as we don't expect the Statement layer to change very often, especially within the same major version.

That's fair.

> I really like your suggestion to extend the pre-generated message (I tried to find whether this is possible, and couldn't find any info)

There is a way to add validations annotations to your proto fields by using [proto-gen-validate](https://github.com/bufbuild/protoc-gen-validate), once you add the annotations, the generated code will have a `Validate` method. You can find an example of how I use that [here](https://github.com/chainloop-dev/chainloop/blob/ac7d2f5df369a6000185bf17a49f86557b6c8b1c/app/controlplane/api/controlplane/v1/integrations.proto#L56), then this [validation file](https://github.com/chainloop-dev/chainloop/blob/main/app/controlplane/api/controlplane/v1/integrations.pb.validate.go) gets generated.

That said, `proto-gen-validate` mostly focuses on unary per-field validations and in your case (based on the validation code in this PR), you might need something more custom, in which case a custom method is what you might need.

> Don't disagree... what do you suggest as an alternative? This seems like a general problem with using structpb.Struct rather than this implementation.

Indeed, nothing specific about this implementation, my implicit point was that the wrapper was not helping with one of the hardest parts of the construction, the creation of protopb.Structs, specially in the case of annotations. If that was not the case I would have seen a good reason for having this wrapper. In the future, the wrappers (wherever the are) could maybe accept a subset of `map[x]interface` and then internally it will generate the structpb, check the error and so on, simplifying the call.  

> We didn't go with anypb in this case because we do want Struct fields to map to JSON Objects, and the Statement cannot dictate the format of the Struct.

re alternative: honestly I don't know, you have a convincing point on the JSON compatibility. Now I am curious, I'll do some research :)

Thanks for this, looking good! Can't wait to adopt it :)

marcelamelara: @migmartri Made some updates addressing several of your comments. Hopefully that works!

> There is a way to add validations annotations to your proto fields by using [proto-gen-validate](https://github.com/bufbuild/protoc-gen-validate), once you add the annotations, the generated code will have a Validate method. You can find an example of how I use that [here](https://github.com/chainloop-dev/chainloop/blob/ac7d2f5df369a6000185bf17a49f86557b6c8b1c/app/controlplane/api/controlplane/v1/integrations.proto#L56), then this [validation file](https://github.com/chainloop-dev/chainloop/blob/main/app/controlplane/api/controlplane/v1/integrations.pb.validate.go) gets generated.

Learn something new every day :) Very cool, thanks!

> If that was not the case I would have seen a good reason for having this wrapper. In the future, the wrappers (wherever the are) could maybe accept a subset of map[x]interface and then internally it will generate the structpb, check the error and so on, simplifying the call.

Really good point. Yeah, I imagined this is the sort of thing you had in mind given your previous comment. In many cases, we imagine that the protobufs will actually be populated based on some serialized data, so I wonder about the broader utility of a wrapper that populates the proto messages directly, given libraries like `protojson` and such.

migmartri (commented):
Thanks for applying some of the feedback! This looks great to me!

I'll leave the approval to some of the maintainers :)

LGTM!

go/v1/resource_descriptor.go
```diff
@@ -0,0 +1,14 @@
+/*
+Wrapper APIs for in-toto attestation ResourceDescriptor protos.
+*/
+
+package v1
+
+func (d *ResourceDescriptor) Validate() bool {
+	// at least one of name, URI or digest are required
+	if d.GetName() == "" && d.GetUri() == "" && len(d.GetDigest()) == 0 {
+		return false
```
TomHennen, line 10: Would it make sense, now or in a future version, to have this return an error indicating why it failed?  Perhaps this could return `error` instead of bool?

go/v1/statement.go
```diff
@@ -0,0 +1,40 @@
...
+Wrapper APIs for in-toto attestation Statement layer protos.
+*/
+
+package v1
+
+const StatementTypeUri = "https://in-toto.io/Statement/v1"
+
+func (s *Statement) Validate() bool {
+	if s.GetType() != StatementTypeUri {
+		return false
```
TomHennen, line 11: As with ResourceDescriptor perhaps this could return an error instead of bool?

TomHennen (changes requested):
Thanks for this work Marcela, it seems pretty valuable!

go/v1/resource_descriptor.go
marcelamelara: Yeah, I think this makes total sense, and I'd rather change the function signature now, than later :)

go/v1/statement.go
marcelamelara: done.

TomHennen (approved):

TomHennen:
Thanks Marcela!

marcelamelara:
Just going to note: We may be able to replace per-language validators with something [more portable](https://github.com/bufbuild/protoc-gen-validate/issues/741) in the future.

go/v1/resource_descriptor_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func TestBadResourceDescriptor(t *testing.T) {
+	got := &ResourceDescriptor{}
+	err := protojson.Unmarshal([]byte(badRd), got)
+
+	assert.NoError(t, err, "Error during JSON unmarshalling")
+
+	err = got.Validate()
+
+	assert.Error(t, err, "Error: created malformed ResourceDescriptor")
```
pxp928, line 68: nit: can also use `errors.Is()` to check that we same error back but this is good for now.

pxp928 (approved):
LGTM. A small nit that can be improved later if needed if more error messages are possible. Thanks for all the work!

PR #227, opened 2023-05-24
Add Python validation APIs and unit tests

### LINKED ISSUE #176: Create test suite for Python bindings
This issue is for implementing a basic but comprehensive test suite for the Python bindings for the attestation layers and predicates.

Tasks:
- [x] Write individual test files for each Python object
- [x] Update the Makefile to invoke the Python test suite
- [x] Update the docs/protos.md to include instructions for running the test suite

Depends on completion of #171 .

marcelamelara:
This PR introduces basic wrapper APIs for the Python language bindings for the Statement layer. I also adds basic tests and documentation for testing the Python language bindings.

Fixes #176


python/in_toto_attestation/v1/resource_descriptor.py
```diff
@@ -0,0 +1,15 @@
...
+class ResourceDescriptor:
+    def __init__(self, rd=None):
+        self.pb = rdpb.ResourceDescriptor()
+
+        if rd:
+            self.pb.CopyFrom(rd)
+
+    def validate(self):
+        # at least one of name, URI or digest are required
+        if (not self.pb.name and not self.pb.uri and not self.pb.digest) or len(self.pb.digest) == 0:
```
TomHennen, line 14: If self.pb.name is set but len(self.pb.digest) is 0, that's probably valid, but I think we'd through a value error?

(Or am I mis-reading the logic?)

python/in_toto_attestation/v1/statement.py
```diff
@@ -0,0 +1,36 @@
+#  Wrapper class for in-toto attestation Statement protos.
+
+import in_toto_attestation.v1.statement_pb2 as spb
+from in_toto_attestation.v1.resource_descriptor import ResourceDescriptor
+
+STATEMENT_TYPE_URI = 'https://in-toto.io/Statement/v1'
+
+class Statement:
+    def __init__(self, s=None):
+        self.pb = spb.Statement()
```
TomHennen, line 10: Should this set self.pb.type to STATEMENT_TYPE_URI so that users don't have to?

python/in_toto_attestation/v1/statement.py
```diff
@@ -0,0 +1,36 @@
+#  Wrapper class for in-toto attestation Statement protos.
+
+import in_toto_attestation.v1.statement_pb2 as spb
+from in_toto_attestation.v1.resource_descriptor import ResourceDescriptor
+
+STATEMENT_TYPE_URI = 'https://in-toto.io/Statement/v1'
+
+class Statement:
+    def __init__(self, s=None):
```
TomHennen, line 9: I'm not sure if we want to follow any particular style conventions, so I guess these comments are optional :)

It's not obvious from the method signature what `s` is.  Statements require subjects so it could conceivably be that.  Perhaps it could be given a more descriptive name (or comments)?  Alternatively, I might have expected __init__ to take the subject of the statement (since it's required).  If we want to have the ability to copy from an existing statement perhaps that could be a separate function?  Perhaps something like

```python
@staticmethod
def copy_from_pb(s):
  # ...
```

python/in_toto_attestation/v1/statement.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # check all resource descriptors in the subject
+        subject = self.pb.subject
+        for rdpb in subject:
+            rd = ResourceDescriptor(rdpb)
+            rd.validate()
+
+            # v1 statements require the digest to be set in the subject
+            if len(rd.pb.digest) == 0:
+                raise ValueError('At least one digest required')
```
TomHennen, line 30: Would it be possible to give the user feedback on which subject failed this check?

python/in_toto_attestation/v1/statement.py
```diff
@@ -0,0 +1,36 @@
...
+
+        if not self.pb.predicateType:
```
TomHennen, line 32: Should this check for empty predicateTypes too?

tests/python/test_resource_descriptor.py
```diff
@@ -0,0 +1,59 @@
...
+        self.test_rd = create_test_desc()
+
+    def test_create_resource_descriptor(self):
+        self.test_rd.validate()
+        
+    def test_json_parse_resource_descriptor(self):
+        got_pb = pb_json.Parse(self.want_full_rd, rdpb.ResourceDescriptor())
+        got = got_pb.SerializeToString(deterministic=True)
+
+        want = self.test_rd.pb.SerializeToString(deterministic=True)
```
TomHennen, line 40: I think `got` and `want` might be switched here?

tests/python/test_resource_descriptor.py
```diff
@@ -0,0 +1,59 @@
...
+    desc.content = b'bytescontent'
+    desc.downloadLocation = 'https://example.com/test.zip'
+    desc.mediaType = 'theMediaType'
+    desc.annotations['a1'].update({'keyStr': 'val1', 'keyNum': 13})
+    desc.annotations['a2'].update({'keyObj': {'subKey': 'subVal'}})
+    return ResourceDescriptor(desc)
+
+class TestResourceDescriptor(unittest.TestCase):
+    def setUp(self):
+        self.want_full_rd = '{"name":"theName","uri":"https://example.com","digest":{"alg":"abc123"},"content":"Ynl0ZXNjb250ZW50","downloadLocation":"https://example.com/test.zip","mediaType":"theMediaType","annotations":{"a1":{"keyNum": 13,"keyStr":"val1"},"a2":{"keyObj":{"subKey":"subVal"}}}}'
```
TomHennen, line 25: optional: It looks like these are all only used in single tests.  My usual preference would be to define them in those tests. That way it's easier for folks reading the test to understand what's going on.  If you follow that approach you won't need setUp at all.

tests/python/test_statement.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_json_parse_statement(self):
+        got_pb = pb_json.Parse(self.want_st, stpb.Statement())
+        got = got_pb.SerializeToString(deterministic=True)
+
+        want = self.test_st.pb.SerializeToString(deterministic=True)
+
+        self.assertEqual(got, want, 'Protos do not match')
+
+    def test_bad_statement_type(self):
+        bad_st = '{"_type":"https://in-toto.io/Statement/v0","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}'
```
TomHennen, line 43: optional: I think this would be a bit easier to read if instead you did something like:

```
bad_st = create_test_statement()
bad_st.pb.type = 'https://in-toto.io/Statement/v0'
```

This would make it clear that the only difference from the 'passing' test and the failing one is the single change that was made.

The same approach could probably be taken for the other 'bad' tests.

WDYT?

TomHennen (changes requested):

tests/python/test_statement.py
marcelamelara: Yes, this makes total sense to me... not sure why I didn't think of this... thanks!

adityasaky:
> I have two ideas for addressing this: 1) The statement class provides a signable_bytes API as a replacement for the existing Signable API, or 2) /in-toto uses the serialized protobuf message for signing, though this may break verifiers. Thoughts?

I think option 1 probably makes sense for compatibility. FWIW, we should test across implementations as soon as we implement statements. cc @pradyumnakrishna who's beginning to look at cross implementation compatibility for in-toto

adityasaky:
I suggest we proceed with this PR without supporting anything specific re signing. If in-toto/in-toto implements an intermediate class for the Statement that builds on what's provided here, it can use the existing `Signable` class for consistency.

python/in_toto_attestation/v1/resource_descriptor.py
marcelamelara: Ah... you're right. This check currently follows Statement.subject requirements. Thanks for catching this!

tests/python/test_statement.py
marcelamelara: I need to think about this more, actually. Switching to using the statement object returned by `create_test_statement()` to create bad statements would change the test setup everywhere, as we currently build the bad proto from a JSON string. And if we want to be consistent with test_resource_descriptor.py, we'd have to change the setup over there, too. Anyway, making these changes isn't hard, so I definitely don't mind doing it, if we think these changes would significantly clean up/clarify the test cases.

python/in_toto_attestation/v1/statement.py
marcelamelara: done.

TomHennen (approved):
Thanks Marcela!

adityasaky:
LGTM!

docs/testing.md
```diff
@@ -24,3 +24,28 @@ files in the `go/` directory tree.
...
+### Writing new Python tests
+
+Please use the standard [Python unittest package] to write tests
+for new predicates. For example tests, take a look at the `test_*.py`
+modules in the `tests/python/` directory tree.
+
+At a minimum, we suggest testing JSON marshalling and unmarshalling
+of the Python language bindings.
+
+[Golang testing package]: https://pkg.go.dev/testing
```
joshuagl, line 50: Nit: I know you are fixing an existing use of the term, but it's Go _not_ Golang :-)

python/in_toto_attestation/v1/resource_descriptor.py
```diff
@@ -0,0 +1,28 @@
+#  Wrapper class for in-toto attestation ResourceDescriptor protos.
+
+import in_toto_attestation.v1.resource_descriptor_pb2 as rdpb
+
+class ResourceDescriptor:
+    def __init__(self, name='', uri='', digest={}, content=bytes(), download_location='', media_type='', annotations={}):
```
joshuagl, line 6: Why do we have default arguments which are empty values here? If these fields are required, the API user should supply them. If they are not required, we should not be providing faux values.

python/in_toto_attestation/v1/statement.py
```diff
@@ -0,0 +1,45 @@
+#  Wrapper class for in-toto attestation Statement protos.
+
+import in_toto_attestation.v1.statement_pb2 as spb
+from in_toto_attestation.v1.resource_descriptor import ResourceDescriptor
+
+STATEMENT_TYPE_URI = 'https://in-toto.io/Statement/v1'
+
+class Statement:
+    def __init__(self, subjects=[], predicate_type='', predicate={}):
```
joshuagl, line 9: Same comment on default arguments

python/in_toto_attestation/v1/statement.py
```diff
@@ -0,0 +1,45 @@
...
+    def validate(self):    
+        if self.pb.type != STATEMENT_TYPE_URI:
+            raise ValueError('Wrong statement type')
+
+        if len(self.pb.subject) == 0:
+            raise ValueError('At least one subject required')
+
+        # check all resource descriptors in the subject
+        subject = self.pb.subject
+        for i, rdpb in enumerate(subject):
```
joshuagl, line 31: As we never use `i`, why not:
```suggestion
        for rdpb in subject:
```

joshuagl (changes requested):
Could we add type annotations to the public methods (especially constructors)? 🙏

python/in_toto_attestation/v1/resource_descriptor.py
marcelamelara: Most of these fields are optional, but we still want to give API users the ability to set them through the constructor, hence the keyword arguments. We could use the more generic `*kwargs` parameter, though the onus would be on the API user to know which are acceptable `kwargs` to pass, and would make the constructor more complex too.

The only required fields are _one of_ name, uri, or digest. Unless there's something I'm fundamentally misunderstanding about how to code in Python, if we make these three fields positional arguments, they'll always be required, even if only one is needed, making the API user to always pass in a dummy value themselves.

So, how do you suggest API users set optional fields in the protobuf? The default protobuf constructor will set all fields to an empty/dummy value, so the default values passed in through the function arguments don't really change the contents of optional fields.

python/in_toto_attestation/v1/statement.py
joshuagl: 🤦 sorry, I was sure I'd carefully looked for a use before adding the comment. Not carefully enough!

python/in_toto_attestation/v1/resource_descriptor.py
adityasaky: So the options are:
* no defaults, make the user set an empty value / None for the ones they don't want to set
* defaults for all, user overrides the ones they want to set
* kwargs

I think each one has some pain points, but I do think it should be between option 1 and 2. Defaults for all looks strange but it might be the easiest for the user?

python/in_toto_attestation/v1/resource_descriptor.py
```diff
@@ -0,0 +1,37 @@
...
+        self.pb.mediaType = media_type
+
+        # HACK
+        if annotations:
+            # Value type fields are considered Messages in Python.
+            # this is needed to set Value fields into the annotations map
+            # without having to know its value type a priori.
+            for k, v in annotations.items():
+                val_pb = pb_json.Parse(json.dumps(v), Value())
+                self.pb.annotations[k].CopyFrom(val_pb)
```
marcelamelara, line 26: If any one has a suggestion for cleaning this up, I'm all ears. The switch to using Values in `annotations` since #251  actually complicated the Python API for ResourceDescriptors, as Value is an actual protobuf message (i.e., an arbitrary value can't just be assigned via dicts as was the case when it was a Struct/object). 

After a fair amount of searching, using json marshalling seems like the most succinct way of setting arbitrary Values in the `annotations` map. The alternative, as far as I found, is to set the individual `int_value/string_value/etc` properties in the Value proto, which would require us to check the type of each `annotations` field to set the appropriate property.

marcelamelara: cc @joshuagl @adityasaky

adityasaky: I didn't think about it before in https://github.com/in-toto/attestation/pull/251 and the corresponding issue but should we quickly amend it again to just use a Struct instead of map[string]Value? Struct is essentially a map[string]Value anyway, but it's easier to handle from a dict and such.

If we don't want to tweak it again, I think this hack is probably the best option.

joshuagl: I'm not sure why this is considered a hack?

I might avoid the additional nesting under the `if` by doing something like:
```Python
annotations = annotations or {}
# Value type fields are considered Messages in Python.
# this is needed to set Value fields into the annotations map
# without having to know its value type a priori.
for k, v in annotations.items():
    # ...
```

marcelamelara: I don't think we should revert the ResourceDescriptor, though I do feel we should have thought a little more carefully about the implications of the use of Value in the `annotations` field. 

I think the current solution feels very much like a hack because its a lazy workaround for the fact that the protobuf object cannot handle the conversion of arbitrary Python values to the appropriate Value internally. The Struct type we had in `annotations` earlier does do this, so either I'm missing something or this isn't how protobuf expects Value fields to be used.

adityasaky: So we had `map[string]Struct`, and updated it to `map[string]Value` which is essentially the structure of `Struct`. Had we updated to using `Struct` for annotations instead of `map[string]Value`, I think we'd get the `Value` handling from the library.

marcelamelara: Ah, I misunderstood what you meant. You're suggesting that the type for `annotations` should just be `Struct`. That makes perfect sense to me. I think we should change the protobuf definition, and note that it corresponds to the spec, which defines `annotations as a `map[string]Value`. I'll open an issue for this.

marcelamelara: Ok, I opened #263 to address this.

tests/python/test_resource_descriptor.py
```diff
@@ -0,0 +1,59 @@
+'''
```
adityasaky, line 1: Can this go in a `tests/` directory in `/python`? So all the Python code is contained in the `python` dir, similar to how the Go tests are in the `go` dir with the corresponding packages.

python/in_toto_attestation/v1/resource_descriptor.py
```diff
@@ -0,0 +1,30 @@
+#  Wrapper class for in-toto attestation ResourceDescriptor protos.
+
+import in_toto_attestation.v1.resource_descriptor_pb2 as rdpb
+from google.protobuf.struct_pb2 import Value
+import google.protobuf.json_format as pb_json
+import json
+
+class ResourceDescriptor:
```
adityasaky, line 8: Is this strictly used for validation? So we shadow the pb.RD instance to validate its fields? Because the Statement class defined in this PR expects pb.RD in its constructor if I'm reading this right.

adityasaky (commented):
Couple of thoughts I'd missed before, apologies.

tests/python/test_resource_descriptor.py
marcelamelara: Yep, this sounds good.

python/in_toto_attestation/v1/resource_descriptor.py
marcelamelara: Yes, this is strictly for validation. This sort of wrapping seems to be the regular practice for validating Python proto bindings, AFAIK.

tests/python/test_resource_descriptor.py
marcelamelara: Not related to this PR, but should we also move the `examples/go` dir into the `go/` directory at some point? I've been thinking (in general) about where to house any existing and future examples for in-toto attestation use cases.

marcelamelara:
@joshuagl Can you please re-review to see if I've addressed your previous comments? There are a few unopened PRs that depend on this PR being merged.

pxp928 (approved):
LGTM!

PR #241, opened 2023-06-01
Set dependabot policy for actions to weekly

marcelamelara:


TomHennen (approved):

TomHennen:
I assume this is because you're tired of getting dependabot PRs every day?

PR #243, opened 2023-06-02
Update supported protoc version in CI

marcelamelara:
Fixes a bug in our current make-protos.yml file that's causing the language bindings regeneration to fail.


TomHennen (approved):

pxp928 (approved):
LGTM

Issue #242, opened 2023-06-02
Expand ResourceDescriptor `annotations` to allow any type?


MarkLodato:
The ResourceDescriptor `annotations` field is currently `map<string, object>` which means that it's not possible to have an annotation of type string, number, boolean, or array. For example, you cannot do:

```json
"annotations": {
  "canonicalSource": "https://..."
}
```

Instead you have to do:

```json
"annotations": {
  "canonicalSource": { "value": "https://..." }
}
```

Is there a reason for this restriction? If not, is it possible to remove it (replacing `google.protobuf.Struct` with `google.protobuf.Value`)? Not sure if this would require a version bump.

Background: https://github.com/slsa-framework/slsa/pull/870. In SLSA, we copied the definition of ResourceDescriptor but used `Value`, not realizing this restriction was in place. So now the SLSA spec disagrees with in-toto. Since it's not clear to me why the restriction exists in the first place, I wanted to check here to see if it's worth making in-toto match SLSA.

MarkLodato:
Is the reason for extensibility? For example, if you decide to change the format of the `canonicalSource`, you can add a new field with the new semantics without having to define a new extension?

TomHennen:
I don't recall a reason for this restriction.  I think the original suggestion was that it just be `object`.  I believe I suggested `map<string, object>` as an improvement.  I never thought of `map<string, value>`.

adityasaky:
If I'm not mistaken, `Value` supports embedding objects as well so it shouldn't be backwards incompatible. I think it makes sense to support non-object annotations.

marcelamelara:
+1 to switching to `map<string, value>` for annotations. I believe making the field a map was in part for better extensibility, but I think that's orthogonal to switching map values from object to `Value` type.

TomHennen:
This sounds fine to us.  Not sure if we'll have time to do it soon.  Happy to review PRs!

chuangw6:
[in-toto-golang](https://github.com/in-toto/in-toto-golang/blob/6c516fe9aade75ffea3e026f533dbd1f54b7347d/in_toto/slsa_provenance/v1/provenance.go#L118-L121) has defined the annotation as `map[string]interface{}` which allows pointing to any type i.e. string, number, boolean, or array. 

[...]

PR #251, opened 2023-06-16
Change annotations fields in ResourceDecriptor from objects to values

joshuagl:
This change expands support from objects to all JSON values. This allows maximum flexibility for producers and better matches the implicit reading of the spec many, including the SLSA project, had.

Fixes: #242

### DIFF
--- a/protos/in_toto_attestation/v1/resource_descriptor.proto
+++ b/protos/in_toto_attestation/v1/resource_descriptor.proto
@@ -23,5 +23,5 @@ message ResourceDescriptor {
 
   string mediaType = 6;
 
-  map<string, google.protobuf.Struct> annotations = 7;
+  map<string, google.protobuf.Value> annotations = 7;
 }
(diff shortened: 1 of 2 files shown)


marcelamelara (changes requested):
Thanks @joshuagl ! One change that would be help would be to update/add to the examples of the resource descriptor spec to show a more accurate use of value type fields.

joshuagl:
Whoops, I missed the example. I just pushed 48e03ef to change the existing example. I can try and add one or two more next week if that's desirable?

marcelamelara:
Thanks! Yeah, _I_ think it would be helpful to have a second example on there that shows different possible values, but I don't think it's crucial. The other thing to consider is how these changes will affect the language bindings/unit tests. But we can resolve that in a separate PR, I figure.

marcelamelara:
> Can we consider the PR as is? I can file an issue for more examples.

I think so. Thanks!

marcelamelara (approved):

pxp928 (approved):
LGTM. +1 with @marcelamelara comment about adding other examples that shows different possible values. Can be in another PR.

Issue #250, opened 2023-06-13
CamelCased protos


sherzberg-1:
The protos in https://github.com/in-toto/attestation/tree/main/protos/in_toto_attestation/v1 are camelCased (e.g. `downloadLocation`, `mediaType`, `predicateType`) when they should be snake_cased (as they are in https://github.com/in-toto/attestation/blob/main/protos/in_toto_attestation/predicates/provenance/v1/provenance.proto).

See https://protobuf.dev/programming-guides/style/#message-field-names.

joshuagl:
Thank you for filing this! We are mostly protobuf novices on the project. :-)

Our current field definitions are inconsistent across the project, but we should standardise on the conventions documented in the protobuf programming guide. In fact, we already [claim to be following the conventions there](https://github.com/in-toto/attestation/blob/461c6b92b7f79f4bca73e6c69b68bf6121052221/docs/protos.md#protobuf-programming-practices).

PR #257, opened 2023-06-27
Use lower_camel_case consistently for fields in protobuf definitions

joshuagl:
We [claim](https://github.com/in-toto/attestation/blob/461c6b92b7f79f4bca73e6c69b68bf6121052221/docs/protos.md) to adhere to the protobuf programming practices, and thus should also follow the [style guide](https://protobuf.dev/programming-guides/style/#message-field-names) there.

The style guide states that message field names should be lower_snake_case, not lowerCamelCase. Our protobuf definitions are somewhat inconsistent, this PR normalises to lower_snake_case for field names in protobuf definitions.

Fixes: #250

### DIFF
--- a/docs/protos.md
+++ b/docs/protos.md
@@ -49,9 +49,10 @@ the message type [update guidelines].
 
 ## Regenerating proto libraries
 
-[It's typical to keep generated Go code in the repository itself](https://go.dev/doc/articles/go_command#:~:text=and%20then%20check%20those%20generated%20source%20files%20into%20your%20repository)
+[It's typical](https://go.dev/doc/articles/go_command#:~:text=and%20then%20check%20those%20generated%20source%20files%20into%20your%20repository)
+to keep code generated from protobuf definitions in the repository itself,
 since it makes users' lives much easier. However, do NOT manually regenerate
-and check in the libraries, if your change modifies or adds protos,
+and check in the libraries if your change modifies or adds protos.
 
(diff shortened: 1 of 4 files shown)


marcelamelara (approved):
Thanks @joshuagl ! These changes shouldn't break the tests, right?

joshuagl:
> Thanks @joshuagl ! These changes shouldn't break the tests, right?

This PR does not appear to break the tests.
Alas, the tests are already broken. I've proposed a fix in #259. With that fix and this branch combined, the tests pass.

joshuagl:
Alas, no. The generated API does change. Presumably the changed API is not covered by current tests, though? `make go_test` succeeds after my change is merged (it even uses cached results, further indicating the changed API is not covered by the tests).

marcelamelara:
Hmm, the changed APIs should be covered in the tests AFAIK (I wrote them), in that the tests do some (un)marshalling of RD and Statement protos. I wonder now if the JSON protobuf API used in the tests treats the updated API as equivalent? I can look into this, but since the tests are working fine, we can probably leave this for another PR.

joshuagl:
Apologies, it turns out I was wrong. I just did a manual review of the proto changes with this series applied, you can see the diff I've reviewed in a gist https://gist.github.com/joshuagl/ae743664f4422391d6e88010b1d167c0

The Go API is unchanged: struct names remain the same

PR #259, opened 2023-06-28
fix(tests): fix go tests since #251

joshuagl:
The changes in #251 to update the type of the annotation value in a resource descriptor was not reflected in the tests, causing them to fail. Update the tests to use the new annotation value type.

---

I'm not sure why the the follow-on auto-PR to update the generated code #258 did not trigger the [run-go-tests](https://github.com/in-toto/attestation/actions/workflows/run-go-tests.yml) workflow?


pxp928 (approved):
LGTM. It is strange that the workflow did not trigger.

TomHennen (approved):

PR #260, opened 2023-06-28
add missing trigger for push main branch in go-test workflow

pxp928:
Fixes the issues where an auto-PR to update the generated code https://github.com/in-toto/attestation/pull/258 does not trigger the [run-go-tests](https://github.com/in-toto/attestation/actions/workflows/run-go-tests.yml) workflow.

### DIFF
--- a/.github/workflows/run-go-tests.yml
+++ b/.github/workflows/run-go-tests.yml
@@ -1,7 +1,11 @@
 name: Run Go tests
 on:
   workflow_dispatch:
-
+  push:
+    branches:
+      - main
+    paths:
+       - 'go/**'
   # Want to trigger these tests whenever the attestation
   # libraries are regenerated, or new modules/tests are
   # added to the language bindings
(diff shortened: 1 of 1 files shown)


.github/workflows/run-go-tests.yml
```diff
@@ -1,7 +1,11 @@
 name: Run Go tests
 on:
   workflow_dispatch:
-
+  push:
+    branches:
+      - main
+    paths:
+       - 'go/**'
```
TomHennen, line 8: nit: is there an errant space in there?

TomHennen (approved):

.github/workflows/run-go-tests.yml
pxp928: ah yes. Fixed.

marcelamelara (approved):
Thanks for this fix @pxp928 !

joshuagl:
Thanks for fixing this!

PR #263, opened 2023-07-05
Define annotations field type in ResourceDescriptor proto as Struct

### LINKED ISSUE #262: The protobuf field type for `annotations` in the ResourceDescriptor should be `Struct`
Prior to #251, the `annotations` field in the ResourceDescriptor was of type `map[string]Struct`. We updated it to `map[string]Value` to allow mapping of arbitrary values. However, we are now essentially re-implementing the structure of the [`Struct` protobuf type](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L51). Using `Struct` has the advantage of `Value` handling from the protobuf library, rather than having to do so explicitly in our language bindings.

Since the `Struct` protobuf type corresponds to the intended `map[string]Value` type for `annotations` in the spec, we should update the ResourceDescriptor proto to define the `annotations` field as a `Struct`, and highlight that this still meets the spec.

_Originally posted by @adityasaky in https://github.com/in-toto/attestation/pull/227#discussion_r1253422134_

marcelamelara:
Fixes #262.


marcelamelara:
It seems the CI workflow for the Go tests is still using the old protobuf definitions :/ Perhaps we should re-make the protos in the workflow, or merge the updated Go tests in a separate PR....

adityasaky (commented):
LGTM!

TomHennen (approved):

pxp928 (approved):

PR #267, opened 2023-07-14
Update Go tests

### LINKED ISSUE #231: Use `errors.Is()` in Golang tests
Instead of using the `assert.Error()` function, which checks that an error occurred, in a future where our validation functions return more errors, we might also use `errors.Is()` to check that we get the same error back.

_Originally posted by @pxp928 in https://github.com/in-toto/attestation/pull/220#discussion_r1205883501_

adityasaky:
Fixes #231 

* Create named errors for reuse and testing
* Assert in tests expected named errors
* Add test for subject-specific restriction on subject RD

### DIFF
--- a/go/v1/statement_test.go
+++ b/go/v1/statement_test.go
@@ -5,6 +5,7 @@ Tests for in-toto attestation ResourceDescriptor protos.
 package v1
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


go/v1/statement_test.go
```diff
@@ -84,7 +84,20 @@ func TestBadStatementSubject(t *testing.T) {
 
 	err = got.Validate()
 
-	assert.Error(t, err, "Error: created malformed Statement (bad subject)")
+	// With this input, we encounter the RD object's validation error
+	assert.ErrorIs(t, err, ErrRDRequiredField, "Error: created malformed Statement (bad subject)")
```
TomHennen, line 88: optional: I'm wondering if these should be separate tests?  Alternatively maybe this should be a table driven test? https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

TomHennen, line 88: nit: can the error message between these tests distinguish which field is the problem?  (This and the following test have the same messages).  Maybe resolved if they're separate tests?

TomHennen (changes requested):
Looks good!  Just a couple comments.

go/v1/statement_test.go
adityasaky: Updated to use table, I was trying to keep the diff small but this is def better.

adityasaky: Done!

TomHennen (approved):
Thanks!

marcelamelara (approved):
Thank you @adityasaky !

Issue #58, opened 2021-08-03
[Help Wanted] Determine attestation format for vuln scans


Dentrax:
This is a follow-up issue for the https://github.com/sigstore/cosign/issues/442.

I thought it would be more appropriate to continue the discussion for the aforementioned issue, since the spec is non-cosign-related.

---

With @developer-guy, we are currently trying to determine a _vuln scan spec_, as far as we can do best. We can enrich the following brand-new attestation SPEC:

```json
{
  "_type": "https://in-toto.io/Statement/v0.1",
  "subject": [
    {
      "name": "alpine",
      "git_commit": "a1b2c3",
      "digest": {
        "sha256": "c201c331d6142766c866"
      }
    }
  ],
  "predicateType": "SARIF",
  "predicate": {
    "timestamp": "1627564731",
    "owner": {
      "name": "<WHO_RAN_THIS_SCAN>"
    },
    "environment": {
      "name": "GitHub",
      "by": "<PIPELINE_ID>",
      "type": "<CI/CD?> (i.e., GitLab Runner)",
      "cluster": "us-east1-a.prod-cluster",
      "namespace": "<namespace>"
    },
    "success": true,
    "scanners": [
      {
        "name": "trivy",
        "version": "0.19.2",
        "db": {
          "name": "trivydb",
          "version": "v1-2021072912"
        },
        "timestamp": "1627564731",
        "result": "<SARIF RESULT HERE?>"
      },
      {
        "name": "dockle",
        "version": "v0.3.15",
        "timestamp": "1627564731",
        "result": "<SARIF RESULT HERE?>"
      }
    ]
  }
}
```

We called the _predicateType_ as _SARIF_. But I think that name, not fits this type since the content is not in [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html) format. We may have to reconsider the name.

It's obvious that it's a bit hard to think of best practices during the implementation of the first version of the spec. It would be great if you maintainers get involved and give a hand to us to improve the overall structure. So we can easily implement the model into **in-toto** project in order to do _validate_ and _generate_ the attestation. Is that make sense to you? Thanks! We are waiting for your feedback about this.

FYI @dlorenc @NitinJain2 @trishankatdatadog

developer-guy:
We (w/@Dentrax) have a concern about this spec because as you can see there is a `result` section in the list of scanners.
[...]

PR #268, opened 2023-07-23
vulnerability attestation: ITE-9 specification

hectorj2f:
As a continuation of the discussion initiated in https://github.com/in-toto/attestation/issues/58 and the Slack [thread](https://cloud-native.slack.com/archives/CM46K2VT2/p1689167384025049), I decided to propose a vulnerability scan-specific in-toto predicate.

This ITE-9 document describes a vulnerability predicate type to represent vulnerability reports from the scanners in an "exportable" manner and independently of the format chosen to output the scanning results.

Note that, the fields proposed for this new predicate type are inspired by the [cosign specification for the vulnerability attestations](https://github.com/sigstore/cosign/blob/main/specs/COSIGN_VULN_ATTESTATION_SPEC.md).

### DIFF
--- a/spec/predicates/vuln.md
+++ b/spec/predicates/vuln.md
@@ -42,25 +42,25 @@ The fields that make up this predicate type are:
 The `subject` contains whatever software artifacts are to be associated with this vulnerability report document.
 The `predicate` contains a JSON-encoded data with the following fields:
 
-**scanner** object
+**scanner, required** object
 
 > There are lots of container image scanners such as Trivy, Grype, Clair, etc.
 > This field describes which scanner is used while performing a container image scan,
 > as well as version information and which Vulnerability DB is used.
 
-**scanner.uri** string (ResourceURI), optional
+**scanner.uri, required** string (ResourceURI)
 
 > > URI indicating the identity of the source of the scanner.
 
-**scanner.version** string (ResourceURI), optional
+**scanner.version, optional** string (ResourceURI)
 
 > > The version of the scanner.
 
-**scanner.db.uri** string (ResourceURI), optional
+**scanner.db.uri, optional** string (ResourceURI)
 
 > > > URI indicating the identity of the source of the Vulnerability DB.
 
-**scanner.db.version** string, optional
+**scanner.db.version, optional** string
 
 > > > The version of the Vulnerability DB.
 
@@ -68,41 +68,41 @@ The `predicate` contains a JSON-encoded data with the following fields:
 
 > > > The timestamp of when the vulnerability DB was updated last time.
 
-**scanner.result** list
+**scanner.result, required** list
 
-> > The result contains a list of vulnerabilities.
+> > The result contains a list of vulnerabilities. Note that an empty list means the **scanner** found no vulnerabilities.
 > > This is the most important part of this field because it'll store the scan result as a whole. So, people might want
 > > to use this field to take decisions based on them by making use of Policy Engines tooling whether allow or deny these images.
 
-**scanner.result.[*].vulnerability** object
+**scanner.result.[*].vulnerability, optional** object
 
 > > > The vulnerability object defines information about each one of the vulnerabilities found by the scanner.
 
-**scanner.result.[*].vulnerability.id** string
+**scanner.result.[*].vulnerability.id, required** string
 
-> > > > This is the identifier of the vulnerability, e.g. GHSA-r9p9-mrjm-926w, CVE-123.
+> > > > This is the identifier of the vulnerability, e.g. [GHSA-fxph-q3j8-mv87](https://github.com/advisories/GHSA-fxph-q3j8-mv87) whose CVE id is [CVE-2017-5645](https://nvd.nist.gov/vuln/detail/CVE-2017-5645).
 
-**scanner.result.[*].vulnerability.severity** object
+**scanner.result.[*].vulnerability.severity, required** object
 
 > > > > The severity contains a list to describe the severity of a vulnerability using one or more quantitative scoring method.
 
-**scanner.result.[*].vulnerability.severity.type** string
+**scanner.result.[*].vulnerability.severity.method, required** string
 
-> > > > > The type describes the quantitative method used to calculate the associated.
+> > > > > The method describes the quantitative method used to calculate the associated severity score such as nvd, cvss and others.
 
-**scanner.result.[*].vulnerability.severity.score** string
+**scanner.result.[*].vulnerability.severity.score, required** string
 
 > > > > > This is a string representing the severity score based on the selected method.
 
-**scanner.result.[*].vulnerability.annotations** list
+**scanner.result.[*].vulnerability.annotations, optional** list
 
 > > > > > This is a list of key/value pairs where scanners can add additional custom information.
 
-**metadata.scanStartedOn, required** string (timestamp)
+**metadata.scanStartedOn, required** Timestamp
 
 > > The timestamp of when the scan started.
 
-**metadata.scanFinishedOn, required** string (timestamp)
+**metadata.scanFinishedOn, required** Timestamp
 
 > > The timestamp of when the scan completed.
 
@@ -110,54 +110,44 @@ The `predicate` contains a JSON-encoded data with the following fields:
 
 ```jsonc
 {
-  "_type": "https://in-toto.io/Statement/v0.1",
+  "_type": "https://in-toto.io/Statement/v1",
   "subject": [
     {
-      ...
+      "name": "foo.jar",
+      "digest": {"sha256": "fe4fe40ac7250263c5dbe1cf3138912f3f416140aa248637a60d65fe22c47da4"}
     }
   ],
   // Predicate:
-  "predicateType": "https://in-toto.io/attestation/vulns/attribute-report/v0.1",
+  "predicateType": "https://in-toto.io/attestation/vulns/v0.1",
   "predicate": {
     "invocation": {
       "parameters": [],
-      // [ "--format=json", "--skip-db-update" ]
-      "uri": "",
-      // https://github.com/developer-guy/alpine/actions/runs/1071875574
-      "event_id": "",
-      // 1071875574
-      "builder.id": ""
-      // GitHub Actions
+      "uri": "https://github.com/developer-guy/alpine/actions/runs/1071875574",
+      "event_id": "1071875574",
+      "builder.id": "GitHub Actions"
     },
     "scanner": {
-      "uri": "",
-      // pkg:github/aquasecurity/trivy@244fd47e07d1004f0aed9
-      "version": "",
-      // 0.19.2
+      "uri": "pkg:github/aquasecurity/trivy@244fd47e07d1004f0aed9",
+      "version": "0.19.2",
       "db": {
-        "uri": "",
-        // pkg:github/aquasecurity/trivy-db/commit/4c76bb580b2736d67751410fa4ab66d2b6b9b27d
-        "version": "",
-        // "v1-2021080612"
-        "lastUpdate": ""
-        // 2021-08-06T17:45:50.52Z
+        "uri": "pkg:github/aquasecurity/trivy-db/commit/4c76bb580b2736d67751410fa4ab66d2b6b9b27d",
+        "version": "v1-2021080612",
+        "lastUpdate": "2021-08-06T17:45:50.52Z"
       },
       "result": [
         {
          "id": "CVE-123",
          "severity": [
-          { "type": "nvd", "score": "medium"},
-          { "type": "cvss_score", "score", "5.2" }
+          { "method": "nvd", "score": "medium"},
+          { "method": "cvss_score", "score", "5.2" }
          ]
         },
         {...}
       ]
     },
     "metadata": {
-      "scanStartedOn": "",
-      // 2021-08-06T17:45:50.52Z
-      "scanFinishedOn": ""
-      // 2021-08-06T17:50:50.52Z
+      "scanStartedOn": "2021-08-06T17:45:50.52Z",
+      "scanFinishedOn": "2021-08-06T17:50:50.52Z"
     }
   }
 }
--- a/spec/predicates/vuln.md
+++ b/spec/predicates/vuln.md
@@ -1,0 +1,136 @@
+# Predicate type: Vulnerabilities
+
+Type URI: https://in-toto.io/attestation/vulns/attribute-report
+
+Version: 0.1
+
+## Purpose
+
+The definition of a vulnerability attestation type has been discussed in the past in [in-toto attestation](https://github.com/in-toto/attestation/issues/58) and [issue](https://github.com/sigstore/cosign/issues/442). However we need to identify two different purposes from these initial conversations:
+
+* The definition of a common format to represent the results of a vulnerability report.
+
+* The definition of a certain set of metadata fields that could help to consume these vulnerability attestations from the different scanning tools.
+
+Obviously the first goal is quite challenging and requires a bigger community to agree upon a specific format. As a consequence, the following attestation type focuses on the definition of that common metadata which could enable the beginning of an exportable and manageable vulnerability attestation.
+
+This document describes a vulnerability attestation type to represent vulnerability reports from the scanners in an "exportable" manner and independently of the format chosen to output the results.
+
+## Prerequisites
+
+The in-toto [attestation] framework and a [Vulnerability scanner tool].
+
+## Use Cases
+
+When sharing the results of a vulnerability scan using an attestation, there is certain metadata that is crucial to trust and reuse this information.
+Information about the scanner used during the scanning is relevant to trust these resuls. The state of the vulnerability database used to search for vulnerabilities defines the accuracy of the results. Other metadata information such as the timestamp when the scan finished could define the reusability of these results.
+
+## Model
+
+This is a predicate type that fits within the larger [Attestation] framework.
+
+## Schema
+
+The schema of this predicate type is documented below.
+
+### Fields
+
+The fields that make up this predicate type are:
+
+The `subject` contains whatever software artifacts are to be associated with this vulnerability report document.
+The `predicate` contains a JSON-encoded data with the following fields:
+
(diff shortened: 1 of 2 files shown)


spec/predicates/vuln.md
```diff
@@ -0,0 +1,136 @@
...
+
+**scanner.db.version** string, optional
+
+> The version of the Vulnerability DB.
+
+**scanner.db.lastUpdate string (Timestamp), required**
+
+> The timestamp of when the vulnerability DB was updated last time.
+
+**scanner.result** object
```
pxp928, line 69: As this is the most important, we should have a set struct that people should follow. For example, filed to specify vulnerabilityIDs and such.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,136 @@
...
+```jsonc
+{
+  "_type": "https://in-toto.io/Statement/v0.1",
+  "subject": [
+    {
+      ...
+    }
+  ],
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/vulns/attribute-report/v0.1",
```
pxp928, line 93: Thanks @hectorj2f this is great. We had almost exactly the same attestation we generated for vulnerability reporting on GUAC: https://github.com/guacsec/guac/blob/main/internal/testing/testdata/exampledata/certify-vuln.json

Once this one finalizes, I would be happy to change over GUAC to use this.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,136 @@
...
+
+Obviously the first goal is quite challenging and requires a bigger community to agree upon a specific format. As a consequence, the following attestation type focuses on the definition of that common metadata which could enable the beginning of an exportable and manageable vulnerability attestation.
+
+This document describes a vulnerability attestation type to represent vulnerability reports from the scanners in an "exportable" manner and independently of the format chosen to output the results.
+
+## Prerequisites
+
+The in-toto [attestation] framework and a [Vulnerability scanner tool].
+
+## Use Cases
```
TomHennen, line 23: nit: for some reason Markdown doesn't seem to identify this as a header?

spec/predicates/vuln.md
```diff
@@ -0,0 +1,136 @@
...
+
+When sharing the results of a vulnerability scan using an attestation, there is certain metadata that is crucial to trust and reuse this information.
+Information about the scanner used during the scanning is relevant to trust these resuls. The state of the vulnerability database used to search for vulnerabilities defines the accuracy of the results. Other metadata information such as the timestamp when the scan finished could define the reusability of these results.
```
TomHennen, line 26: nit: resuls ->  results

spec/predicates/vuln.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Model
+
+This is a predicate type that fits within the larger [Attestation] framework.
```
TomHennen, line 30: I think this section could use more detail about how you expect the attestation to be used.

spec/predicates/vuln.md
TomHennen: Agreed.

I'd probably expect a list of vulnerability identifiers coupled with severity.  If you'd like to allow scanners to add additional custom information perhaps let them add annotations?

E.g.

```
scanner.result = [{"id": "CVE-123", "severity": "medium", "annotations": { "cvss_score": 5.2 }}, {...}]
```

I could see having different lists for CVEs vs OSV results?

TomHennen (changes requested):

spec/predicates/vuln.md
hectorj2f: It sounds good to me 👍🏻 !

hectorj2f: @TomHennen @pxp928 This suggestion makes sense to me. But I wanted to stay away from the defined format to represent the results at least for now. Users may prefer to use SARIF,or any other custom JSON format to set the results of the scan. I'd say a second iteration should define the format of this `result` struct.

wdyt ?

spec/predicates/vuln.md
hectorj2f: having a look at this!

spec/predicates/vuln.md
TomHennen: IMO the real value from predicates is having the predicates be well defined so that consumers know how to use it.  Without result being well defined I don't think there's much value.  Perhaps I'm missing something?

hectorj2f: @TomHennen Sure! I highlighted it on the `Purpose` section. My initial goal was to define the metadata around the results which could help to exchange these attestations without having to solve the problem of which format to use when sharing the results of a vulnerability attestation.

However I am happy to move on defining a structure for the result's section if we feel we are ready to come to an agreement. cc @znewman01

hectorj2f:
@colek42 Without going into details of defining a specific output format here (e.g. Sarif) for the scanner results, the main purpose consists on defining a set of metadata fields to exchange these attestations without having to worry about which output format was used by the scanner to share the results here. These metadata fields will help anyone consuming these attestations to know things about the used scanner, the state of the database and other values. It is more of a standarization of a set of fields that could help to exchange and trust the results from these attestations. For instance, if I get a vulnerability attestation but I know the vulnerability database used by this scanner hasn't been updated since months. I'll discard these results. Other aspects, such as the version/source of the scanner can represent for certain customers and reason to ignore these attestations. 

However if I now focus on the benefits of a different format versus Sarif. I'd say Sarif is a language conceived to render data in a GUI and consequently it ignores additional data that you can only get from custom output formats (e.g. grype json or trivy json output formats). In addition to that, the sarif specification is massive because it is used to render all kind of scanning results (not just vulnerability scanning).

spec/predicates/vuln.md
pxp928: Hey @hectorj2f, yes we should define a starting structure for the results for an initial release that we can iterate on if needed. For example, having a list of vulnerabilityIDs might be a start. Not sure if we want to include this but we could also have severity:

```
{
	"severity": [ {
		"type": string,
		"score": string
	} ]
}
```
similar to the osv implementation: https://ossf.github.io/osv-schema/#severity-field

hectorj2f: Sgtm! I'll add these suggestions.

colek42:
@hectorj2f thank you.  That makes complete sense.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,173 @@
+# Predicate type: Vulnerabilities
+
+Type URI: https://in-toto.io/attestation/vulns/attribute-report
```
marcelamelara, line 3: Nit: Any reason the "attribute-report" part is in the TypeURI? Shorter/simpler would be preferable.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,173 @@
...
+
+* The definition of a certain set of metadata fields that could help to consume these vulnerability attestations from the different scanning tools.
+
+Obviously the first goal is quite challenging and requires a bigger community to agree upon a specific format. As a consequence, the following attestation type focuses on the definition of that common metadata which could enable the beginning of an exportable and manageable vulnerability attestation.
+
+This document describes a vulnerability attestation type to represent vulnerability reports from the scanners in an "exportable" manner and independently of the format chosen to output the results.
+
+## Prerequisites
+
+The in-toto [attestation] framework and a [Vulnerability scanner tool].
```
marcelamelara, line 21: Seems like the vulnerability scanner tool link is missing? It would be helpful to provide some examples for tools here.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,173 @@
...
+
+When sharing the results of a vulnerability scan using an attestation, there is certain metadata that is crucial to trust and reuse this information.
+Information about the scanner used during the scanning is relevant to trust these results. The state of the vulnerability database used to search for vulnerabilities defines the accuracy of the results. Other metadata information such as the timestamp when the scan finished could define the reusability of these results.
+
+## Model
+
+This is a predicate type that fits within the larger [Attestation] framework.
+The following model aims to provide a well defined list of fields so that consumers know how to start exchaging their scanner results.
+
+This predicate model is inspired by [cosign vulnerability attestation](https://github.com/sigstore/cosign/blob/main/specs/COSIGN_VULN_ATTESTATION_SPEC.md).
```
marcelamelara, line 33: Sorry if this has been discussed before: What is the relationship between this predicate and the cosign vuln attestation? Is this one meant to replace/supersede the Cosign format?

hectorj2f: We want to move the cosign vuln attestation here instead of keeping there. The in-toto attestation repository is the right place to host this type of attestation and to expose it for any future discussions.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+> > This is the most important part of this field because it'll store the scan result as a whole. So, people might want
+> > to use this field to take decisions based on them by making use of Policy Engines tooling whether allow or deny these images.
+
+**scanner.result.[*].vulnerability** object
+
+> > > The vulnerability object defines information about each one of the vulnerabilities found by the scanner.
+
+**scanner.result.[*].vulnerability.id** string
+
+> > > > This is the identifier of the vulnerability, e.g. GHSA-r9p9-mrjm-926w, CVE-123.
```
pxp928, line 83: do we want to include known aliases (list) here for example: https://github.com/advisories/GHSA-fxph-q3j8-mv87 is also CVE-2017-5645

pxp928 (approved):
LGTM! This looks good for an initial release. Thanks @hectorj2f!

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+
+**metadata.scanFinishedOn, required** string (timestamp)
+
+> > The timestamp of when the scan completed.
+
+## Example
+
+```jsonc
+{
+  "_type": "https://in-toto.io/Statement/v0.1",
```
adityasaky, line 113:
```suggestion
  "_type": "https://in-toto.io/Statement/v1",
```

Is there any blocker to using Statement v1?

spec/predicates/vuln.md
hectorj2f: It sounds good to me.

spec/predicates/vuln.md
adityasaky: Thanks!

adityasaky:
@hectorj2f do you think you could add a proto definition for this predicate in https://github.com/in-toto/attestation/tree/main/protos/in_toto_attestation/predicates?

spec/predicates/vuln.md
pxp928: oh sorry, I think my comment was not specific enough. I was thinking about adding a new field that calls out known aliases (if known). Does that make sense?

adityasaky:
Sure, I think that's fine.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+
+> > > The version of the Vulnerability DB.
+
+**scanner.db.lastUpdate, required** string (timestamp)
+
+> > > The timestamp of when the vulnerability DB was updated last time.
+
+**scanner.result** list
+
+> > The result contains a list of vulnerabilities.
```
TomHennen, line 73: nit: can you note how readers should interpret the results if list is empty?  "An empty list indicates the scanner found no vulnerabilities."?

Related: should a note somewhere indicate what should happen if the scanner encountered an error and couldn't complete the scan?  Suggestion: don't emit the attestation at all?

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+
+> > > > This is the identifier of the vulnerability, e.g. [GHSA-fxph-q3j8-mv87](https://github.com/advisories/GHSA-fxph-q3j8-mv87) whose CVE id is [CVE-2017-5645](https://nvd.nist.gov/vuln/detail/CVE-2017-5645).
+
+**scanner.result.[*].vulnerability.severity** object
+
+> > > > The severity contains a list to describe the severity of a vulnerability using one or more quantitative scoring method.
+
+**scanner.result.[*].vulnerability.severity.type** string
+
+> > > > > The type describes the quantitative method used to calculate the associated.
```
TomHennen, line 91: nit: I think there are some words missing after 'associated'?

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+
+**scanner.result.[*].vulnerability.id** string
+
+> > > > This is the identifier of the vulnerability, e.g. [GHSA-fxph-q3j8-mv87](https://github.com/advisories/GHSA-fxph-q3j8-mv87) whose CVE id is [CVE-2017-5645](https://nvd.nist.gov/vuln/detail/CVE-2017-5645).
+
+**scanner.result.[*].vulnerability.severity** object
+
+> > > > The severity contains a list to describe the severity of a vulnerability using one or more quantitative scoring method.
+
+**scanner.result.[*].vulnerability.severity.type** string
```
TomHennen, line 89: In the documentation can we list the various known types (e.g. nvd, cvss_score, whatever cvss v3 would use, ...)?

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+    {
+      ...
+    }
+  ],
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/vulns/v0.1",
+  "predicate": {
+    "invocation": {
+      "parameters": [],
+      // [ "--format=json", "--skip-db-update" ]
```
TomHennen, line 124: nit: I think these comments are actually supposed to be in the various parameter fields, right?

spec/predicates/vuln.md
TomHennen, line 89: Elsewhere [there's a thread](https://github.com/secure-systems-lab/dsse/pull/61#discussion_r1271087361) about avoiding the word 'type' since it can often be a keyword.  Maybe `method` instead since that word is being used elsewhere?

TomHennen (changes requested):
Thanks for making this changes, this is looking good!

spec/predicates/vuln.md
hectorj2f: Yes, I'll add a comment to explain what an empty list means. Generally I'd prefer to emit the attestation to certify who is attesting that the scanner.version found no vulnerabilities.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+> > The timestamp of when the scan completed.
+
+## Example
+
+```jsonc
+{
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [
+    {
+      ...
```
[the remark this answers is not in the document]
hectorj2f: Thanks fixed now!

spec/predicates/vuln.md
hectorj2f: Sounds to me!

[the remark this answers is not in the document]
hectorj2f: I made it explicit which are required and which are optional now. Thanks

hectorj2f:
@TomHennen Thanks for the review, I addressed your comments.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+
+**scanner.result.[*].vulnerability.severity.score, required** string
+
+> > > > > This is a string representing the severity score based on the selected method.
+
+**scanner.result.[*].vulnerability.annotations** list, optional
+
+> > > > > This is a list of key/value pairs where scanners can add additional custom information.
+
+**metadata.scanStartedOn, required** string (timestamp)
```
TomHennen, line 101: nit: instead of 'string (timestamp)' perhaps '_Timestamp_' as in https://github.com/in-toto/attestation/blob/main/spec/predicates/runtime-trace.md?

Note that Timestamp is a specific type supported in in-toto. https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#timestamp

(also below)

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+      "digest": {"sha256": "fe4fe40ac7250263c5dbe1cf3138912f3f416140aa248637a60d65fe22c47da4"}
+    }
+  ],
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/vulns/v0.1",
+  "predicate": {
+    "invocation": {
+      "parameters": [],
+      "uri": "",
+      // https://github.com/developer-guy/alpine/actions/runs/1071875574
```
TomHennen, line 126: I think these comments still need to get inlined in the example?

E.g. `"uri": "https://github.com/developer-guy/alpine/actions/runs/1071875574"`

Right?

TomHennen (changes requested):
Thanks for the changes, just a couple minor things to cleanup.

spec/predicates/vuln.md
```diff
@@ -0,0 +1,170 @@
...
+The schema of this predicate type is documented below.
+
+### Fields
+
+The fields that make up this predicate type are:
+
+The `subject` contains whatever software artifacts are to be associated with this vulnerability report document.
+The `predicate` contains a JSON-encoded data with the following fields:
+
+**scanner** object, optional
```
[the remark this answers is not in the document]
hectorj2f: It makes total sense. Fixed!

spec/predicates/vuln.md
```diff
@@ -0,0 +1,160 @@
...
+The schema of this predicate type is documented below.
+
+### Fields
+
+The fields that make up this predicate type are:
+
+The `subject` contains whatever software artifacts are to be associated with this vulnerability report document.
+The `predicate` contains a JSON-encoded data with the following fields:
+
+**scanner, required** object
```
TomHennen, line 45: nit: it seems like sometimes the type comes first and sometimes required/optional comes first.  Would it make sense to standardize that?

TomHennen (approved):
Thanks!  Just one minor comment.

TomHennen:
Thanks for working through this Hector!

hectorj2f:
Thanks to everyone for the reviews 🎉 !

PR #272, opened 2023-08-01
Add Go and Python bindings for SCAI predicate

marcelamelara:
This PR adds basic SCAI predicate wrapper/verification APIs for both Go and Python bindings.

We can discuss if it would be more appropriate to push these APIs to the individual in-toto implementations (in-toto-golang and in-toto) once they've added support for v1 attestations.

### DIFF
--- a/go/predicates/scai/v0/scai.go
+++ b/go/predicates/scai/v0/scai.go
@@ -4,24 +4,24 @@ Wrapper APIs for SCAI AttributeAssertion and AttributeReport protos.
 
 package v0
 
-import "errors"
+import "fmt"
 
 func (a *AttributeAssertion) Validate() error {
 	// at least the attribute field is required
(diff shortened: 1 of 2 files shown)


go/predicates/scai/v0/scai.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	return nil
+}
+
+func (r *AttributeReport) Validate() error {
+	// at least the attributes field is required
+	attrs := r.GetAttributes()
+	if attrs == nil || len(attrs) == 0 {
+		return errors.New("At least one AttributeAssertion required")
```
pxp928, line 35: nit: similar comment to above

pxp928 (commented):
minor nits else looks good

TomHennen (approved):

pxp928 (approved):

Issue #21, opened 2021-04-24
Clarify that any envelope can be used


MarkLodato:
The Envelope layer is not specified in this spec. We recommend a particular one (signing-spec) but the attestation format can be used with any envelope, so long as the producer and consumer agree.

marcelamelara:
A few points of discussion/suggestions from #283:
- Reference ITE-5 in the Envelope spec https://github.com/in-toto/attestation/pull/283#discussion_r1336020524
- Discuss extensibility and support for other existing formats like Sigstore's bundle, COSE https://github.com/in-toto/attestation/pull/283#discussion_r1336180617
- What fields does an Envelope actually require? https://github.com/in-toto/attestation/pull/283#discussion_r1336279318 & https://github.com/in-toto/attestation/pull/283#discussion_r1336280370

PR #283, opened 2023-09-19
Clarify specification on Bundle and Envelope media types

marcelamelara:
This PR updates the specs for Bundles and Envelopes to more formally describe conventions around attestation layer media types when these are stored in arbitrary storage systems, per https://github.com/in-toto/attestation/issues/271#issuecomment-1675008618

Addresses #271


spec/v1/bundle.md
```diff
@@ -2,40 +2,63 @@
...
 supply chain (e.g. Provenance, Code Review, Test Result, vuln scan, ...) to
 be grouped together, allowing users to make decisions based on all available
 information.
 
-**NOTE**: The bundle is not authenticated as a whole. Instead each individual
-attestation is authenticated using [DSSE]. As such, an attacker might be able
-to _delete_ an attestation without being detected. Predicates that follow
-[the monotonic principle] should not have any issues with this behavior.
+**NOTE**: The Bundle is not authenticated as a whole. Instead each individual
+attestation is authenticated using signature schemes like [DSSE]. As such,
```
joshuagl, line 12: Nit: we should avoid using the term "signature schemes" here because that term is typically used to describe a set of key-gen/signing/verification algorithms.

In the rest of this PR we use signature method / signature format. Let's use those here, too? And ensure we're consistent throught.

spec/v1/bundle.md
```diff
@@ -2,40 +2,63 @@
...
 Bundles SHOULD use the suffix `.intoto.jsonl`.
 
-A bundle of attestations relevant for `<filename>` SHOULD be named
-`<filename>.intoto.jsonl`. Attestations in the bundle MAY have different
-subjects, but they SHOULD all be relevant to that file or its dependencies.
-For example, a package named `foo-1.2.3.tar.gz` with hash `abcd` that was
-built from git commit `1234` could have a bundle name
-`foo-1.2.3.tar.gz.intoto.jsonl` with two attestations, one with subject
-`abcd` and one with subject `1234`.
+<!-- @marcelamelara: This likely needs to be updated for ITE-10/11 -->
```
joshuagl, line 35: Can we file an issue for this?

spec/v1/envelope.md
```diff
@@ -19,5 +21,29 @@ standard DSSE fields.
...
+-   The `<predicate>` MUST match the [predicate specification name]
+    without the file extension. Predicate versioning is handled in the
+    [Statement] layer.
+-   The `<sig>` MUST be a succint alias that unambiguously identifies
+    the Envelope signature format.
+-   Consumers SHOULD NOT rely upon the media type for individual attestations
+    as faithful indicators of predicate type because this information is only
+    authenticated at the [Statement] layer.
+-   To obtain predicate information that is authenticated, consumers MUST
+    parse the Envelope `payload`.
```
joshuagl, line 38:
```suggestion
    parse the Envelope's `payload`.
```

joshuagl (commented):
I agree that this PR improves clarity, thanks for working on it.

I'm worried we're not consistent in using capitalised names for components in the attestation model (Bundle, Attestation, Envelope, Statement, Predicate, etc). Should we do a consistency pass and create some contributor documentation explaining conventions like this?

spec/v1/bundle.md
marcelamelara: Good catch, thanks! Yes, I had been trying to avoid signature scheme for exactly the reasons you mention, but this one managed to sneak in. Will definitely change.

spec/v1/bundle.md
marcelamelara: We probably should. The reason I hadn't was because this comment is based on how the [ITE-10/11 prototype](https://github.com/adityasaky/in-toto-attestation-verifier) that @adityasaky built currently works. So, I don't know if the Bundle/Envelope file naming expectations are specific to the prototype or if they will eventually become part of the spec. To my recollection, ITE-10/11 don't currently include conventions for file naming for verification purposes.

marcelamelara:
> I'm worried we're not consistent in using capitalised names for components in the attestation model (Bundle, Attestation, Envelope, Statement, Predicate, etc). Should we do a consistency pass and create some contributor documentation explaining conventions like this?

Agreed, I don't think we're being consistent on this right now. Personally, I find it helpful to capitalize these terms denote when we're talking about attestation model components vs. not, but I wasn't sure if everyone would agree. If we do, then we should definitely follow your suggestion of doing a consistency pass and adding documentation for this.

spec/v1/bundle.md
adityasaky: I'm also a little iffy on saying "signature method like DSSE". Can we say "each individual attestation is authenticated using signatures recorded in an envelope like DSSE"?

spec/v1/bundle.md
adityasaky: The issue in ITE-10 is less with the name of the bundle as a whole but rather with mapping individual attestations to the steps they correspond to. The bundle format is ill suited to associating attestations with their steps. in-toto 1.0 leans on storing each link metadata on disk using the step name, and for good measure, also includes the step name in the predicate.

spec/v1/bundle.md
marcelamelara: I was going to go with "signature format". That's a term that DSSE uses, too. But this discussion has got me thinking that we may want to clarify the distinction between scheme/method/format on the Envelope spec regardless.

spec/v1/bundle.md
marcelamelara: Gotcha. This is probably something we ought to clarify at the Envelope layer then, to also ensure we're in spec with in-toto 1.0.

marcelamelara: @adityasaky @joshuagl if you could PTAL at the updated text!

spec/v1/bundle.md
```diff
@@ -2,40 +2,61 @@
...
 Bundles SHOULD use the suffix `.intoto.jsonl`.
 
-A bundle of attestations relevant for `<filename>` SHOULD be named
-`<filename>.intoto.jsonl`. Attestations in the bundle MAY have different
-subjects, but they SHOULD all be relevant to that file or its dependencies.
-For example, a package named `foo-1.2.3.tar.gz` with hash `abcd` that was
-built from git commit `1234` could have a bundle name
-`foo-1.2.3.tar.gz.intoto.jsonl` with two attestations, one with subject
-`abcd` and one with subject `1234`.
+-   A Bundle of attestations relevant for `<filename>` SHOULD be named
```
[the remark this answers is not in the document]
adityasaky: OTOH, the example shows why we want a file name and not just project + version `foo-1.2.3` as a single release may have different built artifacts for different target architectures, distributions and so on...open to thoughts.

spec/v1/envelope.md
```diff
@@ -19,5 +21,41 @@ standard DSSE fields.
     specifying its schema.
 -   `payload` MUST be a base64-encoded JSON [Statement].
 
+## File naming convention
+
+Envelopes SHOULD use the suffix `.json`.
+
+-   An Envelope containing an attestation about a particular SW supply chain
+    step `<step-name>` SHOULD be named `<step-name>.json`.
+-   If multiple Envelopes are produced for the same step by different
```
adityasaky, line 30: Some thoughts:
1. Currently the KeyID is included even if there's a single envelope for the step. I think it's okay to loosen that requirement but I'm noting it for completeness.
2. The pubkey hash is not well defined. [The spec](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#421-key-formats) says "The KEYID of a key is the hexadecimal encoding of the SHA-256 hash of the canonical JSON form of the public key." This is not the same as "pubkey-hash" as the key ID is calculated across the key type, scheme, etc. fields as well, _including some that are not documented in the spec_.
3. Existing implementations actually use a [shortened version](https://github.com/in-toto/in-toto/blob/develop/in_toto/models/link.py#L30) of the full key ID.
4. I wonder if we also want to consider moving away from the strict key ID requirements, something like TUF's [TAP-12](https://github.com/theupdateframework/taps/blob/master/tap12.md).

spec/v1/envelope.md
```diff
@@ -19,5 +21,41 @@ standard DSSE fields.
...
+    [functionaries] uniquely identified by a public key, an Envelope name
+    SHOULD include the hash of the public key `<pubkey-hash>` of the signing
+    functionary: `<step-name>.<pubkey-hash>.json`.
+
+## Storage convention
+
+The media type `application/vnd.in-toto.<predicate>+<sig>` SHOULD
+be used to denote an individual attestation in arbitrary storage systems.
+
+-   The `<predicate>` MUST match the [predicate specification name]
```
adityasaky, line 40: ITE-9 / documentation here may benefit from adding a field for the predicate name alongside the type field instead of relying on the file names.

spec/v1/envelope.md
```diff
@@ -19,5 +21,41 @@ standard DSSE fields.
...
+    without the file extension. Predicate versioning is handled in the
+    [Statement] layer.
+-   The `<sig>` MUST be a succint alias that unambiguously identifies
```
adityasaky, line 43: Can we use `sigformat` or something like that? `sig` is also the name of the signature field in DSSE so I'd like to avoid confusion...

spec/v1/envelope.md
```diff
@@ -19,5 +21,41 @@ standard DSSE fields.
...
+    the Envelope signature format.
+-   Consumers SHOULD NOT rely upon the media type for individual attestations
+    as faithful indicators of predicate type because this information is only
+    authenticated at the [Statement] layer.
+-   To obtain predicate information that is authenticated, consumers MUST
```
adityasaky, line 48: Clarify that for this to be authenticated, the consumer is also expected to verify the signature themselves to be safe? Otherwise it could be read as "parse the payload and you'll get the authenticated predicate type" which could lead to the wrong idea that this is already authenticated in some way.

spec/v1/bundle.md
```diff
@@ -2,40 +2,61 @@
...
 ## Data structure
 
 Attestation Bundles use [JSON Lines] to store multiple attestations.
 
--   Each attestation within a bundle MAY have a different signing key,
+-   Each attestation within a Bundle MAY have a different signing key,
     `_type`, `subject`, and/or `predicateType`.
--   Each line SHOULD be an [Envelope]. Consumers MUST ignore unrecognized
-    lines.
+-   Each line within a Bundle SHOULD be an [Envelope].
```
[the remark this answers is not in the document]
TomHennen: It's so the bundle can support other envelope types besides DSSE and allow for extensibility in the future.  This may be needed to support other existing formats like Sigstore's bundle, COSE, or any new bundle formats in the future.

@woodruffw and I discussed this some last week.

spec/v1/envelope.md
```diff
@@ -19,5 +21,41 @@ standard DSSE fields.
...
+    parse the Envelope's `payload`.
+
+### Example
```
TomHennen, line 51: optional: can we provide multiple examples?  E.g. what would the SLSA provenance or VSA type be?

TomHennen (changes requested):

spec/v1/bundle.md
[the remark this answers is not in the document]
adityasaky: I think that still calls for bumping this to a `MUST`. The text in this PR for [Envelope] has been updated to say:

> The RECOMMENDED format and protocol for Envelopes are defined per [DSSE v1.0]. Producers MAY use other signature methods and formats so long as they meet the [Bundle] data structure requirements.

So [Envelope] isn't just DSSE. Also, I'm still nervous about storing a sigstore bundle as an entry in the in-toto attestation bundle as that may not actually be a signature envelope if it only includes a raw signature + verification materials...

spec/v1/envelope.md
```diff
@@ -19,5 +21,41 @@ standard DSSE fields.
     specifying its schema.
 -   `payload` MUST be a base64-encoded JSON [Statement].
 
+## File naming convention
+
+Envelopes SHOULD use the suffix `.json`.
```
TomHennen, line 26: I wonder if this section should say something like "if stored in a file by itself, and not in a [Bundle], ..."

spec/v1/envelope.md
```diff
@@ -19,5 +21,41 @@ standard DSSE fields.
...
+
+-   An Envelope containing an attestation about a particular SW supply chain
```
TomHennen, line 28: Would it make any sense to qualify this with something like "If the attestation is being generated for use with the in-toto verifier..." and then have some other guidance for folks that are using them standalone?  Or may not know what step it's a part of?

spec/v1/bundle.md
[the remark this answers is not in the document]
TomHennen: Oh I missed that.  Previously Envelope was always a DSSE.

TBH it now sounds a bit circular?

Bundle says "every line must be an 'Envelope'"
Envelope says "an envelope can be anything as long as Bundle is OK with it"

What fields does an Envelope actually require?

spec/v1/envelope.md
```diff
@@ -7,7 +7,9 @@ authentication and serialization.
 
 ## Schema
 
-The format and protocol are defined per [DSSE v1.0].
+The RECOMMENDED format and protocol for Envelopes are defined per [DSSE v1.0].
+Producers MAY use other signature methods and formats so long as they meet
+the [Bundle] data structure requirements.
 
 ## Fields
```
TomHennen, line 14: Envelopes can now be ~anything, yet this section is just talking about DSSE.

Do we expect all envelopes, DSSE or not, to have these fields?

TomHennen (changes requested):

spec/v1/envelope.md
TomHennen: Yeah that makes sense to me.  Can definitely keep the +dsse advice though!

spec/v1/bundle.md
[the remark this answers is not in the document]
marcelamelara: I think is a good question/discussion, though this also feels out of scope for this PR. Should we open an issue for this?

spec/v1/envelope.md
```diff
@@ -7,7 +7,9 @@ authentication and serialization.
 
 ## Schema
 
-The format and protocol are defined per [DSSE v1.0].
+The RECOMMENDED format and protocol for Envelopes are defined per [DSSE v1.0].
+Producers MAY use other signature methods and formats so long as they meet
```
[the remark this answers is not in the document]
marcelamelara: Great point, we really should be pointing to ITE-5. Based on https://github.com/in-toto/attestation/pull/283#discussion_r1336280370, I think there's a broader need to update the envelope spec, so I would handle all of that in a separate issue/PR.

spec/v1/envelope.md
marcelamelara: When we discussed this at the Aug 11 maintainer's meeting, we felt the predicate spec filename was the most succinct and uniform way of deriving a predicate name without having to change the Statement schema or create some ad-hoc list with TypeURI-to-name mappings. Do you have any particular concerns around using the predicate spec filename?

spec/v1/envelope.md
marcelamelara: Oh I missed that part of the spec, thanks for pointing that out. For the sake of moving this PR along, I would either go with options 3 or 4 where we match the current implementation (KeyID[0:8]) or TAP-12.

spec/v1/envelope.md
marcelamelara: Makes sense to me. Will update, thanks.

spec/v1/envelope.md
marcelamelara: Agreed, the mediaType advice should stay. We can track the discussion on multi-envelope types in #21.

spec/v1/bundle.md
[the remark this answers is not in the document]
marcelamelara: I say we move this discussion as part of #21.

spec/v1/envelope.md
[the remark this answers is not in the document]
marcelamelara: Move to #21.

spec/v1/envelope.md
marcelamelara: done.

spec/v1/envelope.md
marcelamelara: done.

spec/v1/envelope.md
marcelamelara: This bullet point was removed because we're only supporting DSSE for now.

spec/v1/bundle.md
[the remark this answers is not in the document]
adityasaky: Sounds good.

spec/v1/envelope.md
adityasaky: I recommend 3 for consistency until we can sort it out across both specs.

spec/v1/envelope.md
adityasaky: Not per se. Note that I'm not saying we should update the Statement schema, but rather the format ITE-9 specifies for the predicate specification. So alongside the type, you'd also have the name. I think this is functionally equivalent to using the file name, just that all the information can be derived from the same place.

spec/v1/bundle.md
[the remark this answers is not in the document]
adityasaky: Sounds reasonable to move this out. I will also note that perhaps we can just point to ITE-5 and call it a day. It already lists the desired properties for a signature wrapper, and the in-toto 1.0 spec adheres to it.

spec/v1/bundle.md
```diff
@@ -2,40 +2,59 @@
...
+A package named `foo-1.2.3.tar.gz` with hash `abcd` that was built from git
+commit `1234` could have a Bundle name `foo-1.2.3.tar.gz.intoto.jsonl` with
+two attestations, one with subject `abcd` and one with subject `1234`.
+
+## Storage convention
+
+The media type `application/vnd.in-toto.bundle` SHOULD be used to denote
+a Bundle in arbitrary storage systems.
+
+-   The encoding of the Bundle contents MAY be omitted from its media type
```
adityasaky, line 49: Are there any systems that _require_ an encoding in the media type? If not, maybe the `MAY` is too permissive here. Why allow for two media types at all?

adityasaky (commented):
Some minor comments / updates aside, this LGTM.

spec/v1/envelope.md
marcelamelara: Thanks for clarifying, yes, I had misunderstood your previous comment. We can certainly update ITE-9 for this. Who would be responsible for making that change? You as ITE-9 sponsor?

spec/v1/envelope.md
marcelamelara: That was my inclination, too. I'll make those updates

spec/v1/envelope.md
adityasaky: https://github.com/in-toto/ITE/pull/52

Opened a PR.

spec/v1/bundle.md
marcelamelara: I don't know of any systems that require the encoding, but this is a good question. We weren't really thinking of this as allowing to media types given the implicit expectation for jsonl, but I do realize that some systems match string literals as opposed to parsing the media type string. So, I'm fine with erring on the side of being explicit here. 

@TomHennen Any input here?

spec/v1/envelope.md
marcelamelara: BTW, I could also make this change proposal, I just wasn't sure about the expectations are for making edits to ITEs.

marcelamelara: Thank you @adityasaky ! I'll go take a look.

spec/v1/envelope.md
```diff
@@ -19,5 +19,48 @@ standard DSSE fields.
...
+
+If stored in a dedicated file by itself, and not as part of a [Bundle], an
+Envelope SHOULD use the suffix `.json`.
+
+-   For attestations intended for consumption by [in-toto-verify], an
+    Envelope containing an attestation about a particular SW supply chain
+    step `<step-name>` SHOULD be named `<step-name>.json`.
+-   For other verifiers, or cases in which a step name cannot be easily
+    determined, the attestation producer and consumer SHOULD agree on an
+    acceptable choice for filename: `<env-name>.json`.
```
trishankatdatadog, line 32: What's `env-name` and where does it come from?

spec/v1/envelope.md
```diff
@@ -19,5 +19,48 @@ standard DSSE fields.
...
+    [Statement] layer.
+-   Consumers SHOULD NOT rely upon the media type for individual attestations
+    as faithful indicators of predicate type because this information is only
+    authenticated at the [Statement] layer.
+-   To obtain predicate information that is authenticated, consumers MUST
+    parse the Envelope's `payload`, and verify it against its `signatures`.
+
+### Examples
+
+Example media types for single DSSE-signed attestation predicates include:
```
trishankatdatadog, line 54: I think I finally see now. So these media types would be useful for _transporting_ an attestation requested on demand? It's a way for the server to say to clients, "Hey, this is the unathenticated media type for this attestation so you know in advance how to decode it, but please don't take my word for it---make sure you verify the signatures on the Envelope level first." Is that correct?

TomHennen: Right, or conversely, it let's a client request attestations of a specific media type instead of having to get all of them.

spec/v1/envelope.md
```diff
@@ -19,5 +19,48 @@ standard DSSE fields.
...
+
+The media type `application/vnd.in-toto.<predicate>+dsse` SHOULD
+be used to denote an individual attestation in arbitrary storage systems.
+
+-   The `<predicate>` MUST match the [predicate specification name]
+    without the file extension. Predicate versioning is handled in the
+    [Statement] layer.
+-   Consumers SHOULD NOT rely upon the media type for individual attestations
+    as faithful indicators of predicate type because this information is only
+    authenticated at the [Statement] layer.
```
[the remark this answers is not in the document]
TomHennen: I think by 'this information' it means the `predicateType`.

So maybe 

```
Consumers SHOULD NOT rely upon the media type for individual attestations
as faithful indicators of predicate type and instead only used the
`predicateType`field from the [Statement] layer.
```

spec/v1/envelope.md
marcelamelara: `env-name` is supposed to be an arbitrary envelope file name. I'm not a big fan of this choice, so if you have a suggestion for improvement, please let me know!

trishankatdatadog: Oh, that's not clear at all from the text, I'm afraid, which should explain this.

spec/v1/envelope.md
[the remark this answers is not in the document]
trishankatdatadog: Muuuuuch better, thanks (minor typo: should be "use" instead of "used").

spec/v1/envelope.md
marcelamelara: Can update this.

spec/v1/envelope.md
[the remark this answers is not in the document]
marcelamelara: done.

TomHennen (approved):
This looks great!

spec/v1/bundle.md
```diff
@@ -2,40 +2,59 @@
...
 be grouped together, allowing users to make decisions based on all available
 information.
 
-**NOTE**: The bundle is not authenticated as a whole. Instead each individual
-attestation is authenticated using [DSSE]. As such, an attacker might be able
-to _delete_ an attestation without being detected. Predicates that follow
-[the monotonic principle] should not have any issues with this behavior.
+**NOTE**: The Bundle is not authenticated as a whole. Instead each individual
+attestation is authenticated using signature formats like [DSSE]. As such,
+an attacker might be able to _delete_ an attestation without being detected.
```
trishankatdatadog, line 13: Well, not just delete, but also: (1) rollback to obsolete attestations, or (2) mix-and-match attestations that are not supposed to bundled together, no?

spec/v1/bundle.md
```diff
@@ -2,40 +2,59 @@
...
+Predicates that follow [the monotonic principle] should not have any issues
```
trishankatdatadog, line 14: Where's the link to the monotonic principle?

marcelamelara: good catch, thanks! Fixed.

spec/v1/bundle.md
marcelamelara: Added some clarifications on this, thanks!

trishankatdatadog (approved):
LGTM, thanks

spec/v1/bundle.md
```diff
@@ -2,40 +2,63 @@
...
-[the monotonic principle] should not have any issues with this behavior.
+**NOTE**: The Bundle is not authenticated as a whole. Instead each individual
+attestation is authenticated using signature formats like [DSSE]. As such,
+an attacker might be able to delete valid attestations, replay obsolete
+attestations, and/or inject invalid or irrelevant attestations in a Bundle
+without being detected.
+
+Our Bundle specification aims to reduce the risks of such attacks.
+In addition, predicates that follow the [monotonic principle] should not
+have any issues with delettion attacks.
```
joshuagl, line 19:
```suggestion
have any issues with deletion attacks.
```

spec/v1/envelope.md
```diff
@@ -19,5 +19,48 @@ standard DSSE fields.
     specifying its schema.
 -   `payload` MUST be a base64-encoded JSON [Statement].
 
+## File naming convention
+
+If stored in a dedicated file by itself, and not as part of a [Bundle], an
+Envelope SHOULD use the suffix `.json`.
+
+-   For attestations intended for consumption by [in-toto-verify], an
+    Envelope containing an attestation about a particular SW supply chain
```
joshuagl, line 28: can we expand "SW" here?

joshuagl (approved):
This is excellent, thank you!

spec/v1/bundle.md
marcelamelara: Good catch, thank you!

spec/v1/bundle.md
TomHennen: Sorry for the late reply.  I don't really have an opinion and don't know enough about media types to say.

Is the suggestion that we just not allow the encoding of the bundle contents at all?

(I bet we can fix this in a follow on PR if we want)

TomHennen (approved):

spec/v1/bundle.md
adityasaky: Yes. My understanding is that we need the encoding `+jsonl` when there are different encodings to choose from. If the bundle is always `+jsonl`, do we need it? I'll say that I'm not an expert with media types either.

adityasaky: Also, this isn't a blocker for the PR I think.

marcelamelara: If the consensus is that Bundles currently don't support encodings that aren't jsonl, I'm fine with dropping this line in the spec altogether. It's quick enough fix.

PR #288, opened 2023-10-05
Update vsa.md

trishankatdatadog:
Point to SLSA v1


joshuagl (approved):

TomHennen (approved):

marcelamelara:
Presumably we ought to add an updated proto definition for the VSA now as well?

PR #295, opened 2023-11-01
Add an attestation for policy verification

### LINKED ISSUE #277: Attestation for policy verification
[SLSA](https://slsa.dev) has the concept of a Verification Summary Attestation ([VSA](https://slsa.dev/spec/v1.0/verification_summary)). VSA is an in-toto attestation that indicates the attestor verified an artifact's SLSA metadata and found it to meet a certain SLSA level. SLSA users can use VSAs to cache verification decisions and/or delegate such decisions to another trusted party. VSAs are specific to SLSA, but perhaps it would be useful to have an attestation that was agnostic to the type of policy the attestor used. That attestation would mean “<verifier> evaluated <subject> against <policy_uri> and it (passed|failed).” We've discussed creating such an attestation as part of SLSA, but the community agreed that it was outside SLSA's scope. I'm curious whether in-toto would be interested in pursuing the idea.

IIUC, such an attestation is slightly outside of in-toto’s usual scope since it’s a statement about the attestor’s beliefs about the subject rather than metadata about the subject itself, but the framing allows us to move policy decisions further up the supply chain and away from admission control. For example, instead of gathering up a bundle of attestations for SLSA, vuln scans, etc and evaluating a policy about those attestations during admission control, you could check that policy into version control, evaluate it and produce a verification attestation, and then at admission control you have a much simpler policy that checks that there’s a passing verification attestation for the subject that uses the correct policy and comes from a trusted verifier.

As far as I can tell, in-toto doesn’t support inheritance, but I could imagine different sorts of inheritance attestations over time. The essential elements are the identity of the verifier, a policy URI, and whether the verification passed or failed. I’m not sure what to do about divergence between different verification attestations or whether it’s premature to think about now.

kpk47:
This attestation is based on the SLSA VSA, but modified to support arbitrary security policies.

Fixes #277

### DIFF
--- a/spec/predicates/vsa.md
+++ b/spec/predicates/vsa.md
@@ -1,11 +1,279 @@
-# Predicate type: SLSA Verification Summary (VSA)
+# Predicate type: Verification Summary Attestation (VSA)
 
-Type URI: https://slsa.dev/verification_summary/v1
+Type URI: https://in-toto.io/attestation/verification_summary/v1
 
-Version: 1.0
+Version: 1.1
 
 ## Purpose
 
-Records a SLSA verification decision about an artifact.
+Verification summary attestations communicate that an artifact has been verified
+against some policy and details about that verification. Such details may
+include, but are not limited to, what SLSA level the artifact and/or its
+dependencies were verified to meet.
 
-For more details, please see https://slsa.dev/spec/v1.0/verification_summary
+## Use Cases
+
+Software consumers make a decision about the validity of an
+artifact without needing to have access to all of the attestations about the
+artifact or all of its transitive dependencies.  They delegate
+complex policy decisions to some trusted party and then simply trust that
+party's decision regarding the artifact. That decision is represented by a VSA.
+
+VSAs also allow software producers to keep the details of their build pipeline
+confidential while still communicating that some verification has taken place.
+This might be necessary for legal reasons (keeping a software supplier
+confidential) or for security reasons (not revealing that an embargoed patch has
+been included).
+
+## Prerequisites
+
+Understanding of [in-toto specification] and the in-toto attestation framework.
+
+## Model
+
+A Verification Summary Attestation (VSA) is an attestation that some entity
+(`verifier`) verified one or more software artifacts (the `subject` of an
+in-toto attestation [Statement]) by evaluating the artifact and a `bundle`
+of attestations against some `policy`.  Users who trust the `verifier` may
+assume that the artifacts satisfied the indicated security polciy without
+themselves needing to evaluate the artifact or to have access to the
+attestations the `verifier` used to make its determination.
+
+The VSA also allows consumers to determine the verified levels of
+all of an artifact’s _transitive_ dependencies.  The verifier does this by
+either a) verifying the provenance of each non-source dependency listed in
+the [resolvedDependencies](/provenance/v1#resolvedDependencies) of the artifact
+being verified (recursively) or b) matching the non-source dependency
+listed in `resolvedDependencies` (`subject.digest` ==
+`resolvedDependencies.digest` and, ideally, `vsa.resourceUri` ==
+`resolvedDependencies.uri`) to a VSA _for that dependency_ and using
+`vsa.verifiedProperties` and `vsa.dependencyProperties`.  Policy verifiers
+wishing to establish minimum requirements on dependencies may use
+`vsa.dependencyProperties` to do so.
+
+## Schema
+
+```jsonc
+// Standard attestation fields:
+"_type": "https://in-toto.io/Statement/v1",
+"subject": [{
+  "name": <NAME>,
+  "digest": { <digest-in-request> }
+}],
+
+// Predicate
+"predicateType": "https://slsa.dev/verification_summary/v1",
+"predicate": {
+  "verifier": {
+    "id": "<URI>",
+    "version": {
+      "<COMPONENT>": "<VERSION>",
+      ...
+    }
+  },
+  "timeVerified": <TIMESTAMP>,
+  "resourceUri": <artifact-URI-in-request>,
+  "policy": {
+    "uri": "<URI>",
+    "digest": { /* DigestSet */ }
+  }
+  "inputAttestations": [
+    {
+      "uri": "<URI>",
+      "digest": { <digest-of-attestation-data> }
+    },
+    ...
+  ],
+  "verificationResult": "<PASSED|FAILED>",
+  "verifiedProperties": ["<String>"],
+  "dependencyProperties": {
+    "<String>": <Int>,
+    "<String>": <Int>,
+    ...
+  },
+  "slsaVersion": "<MAJOR>.<MINOR>",
+}
+```
+
+### Parsing rules
+
+This predicate follows the in-toto attestation [parsing rules]. Summary:
+
+-   Consumers MUST ignore unrecognized fields.
+-   The `predicateType` URI includes the major version number and will always
+    change whenever there is a backwards incompatible change.
+-   Minor version changes are always backwards compatible and "monotonic." Such
+    changes do not update the `predicateType`.
+-   Producers MAY add extension fields using field names that are URIs.
+
+### Fields
+
+_NOTE: This section describes the fields within `predicate`. For a description
+of the other top-level fields, such as `subject`, see [Statement]._
+
+<a id="verifier"></a>
+`verifier` _object, required_
+
+> Identifies the entity that performed the verification.
+>
+> The identity MUST reflect the trust base that consumers care about. How
+> detailed to be is a judgment call.
+>
+> Consumers MUST accept only specific (signer, verifier) pairs. For example,
+> "GitHub" can sign provenance for the "GitHub Actions" verifier, and "Google"
+> can sign provenance for the "Google Cloud Deploy" verifier, but "GitHub" cannot
+> sign for the "Google Cloud Deploy" verifier.
+>
+> The field is required, even if it is implicit from the signer, to aid readability and
+> debugging. It is an object to allow additional fields in the future, in case one
+> URI is not sufficient.
+
+<a id="verifier.id"></a>
+`verifier.id` _string ([TypeURI]), required_
+
+> URI indicating the verifier’s identity.
+
+<a id="verifier.version"></a>
+`verifier.version` _map (string->string), optional_
+
+> Map of names of components of the verification platform to their version.
+
+<a id="timeVerified"></a>
+`timeVerified` _string ([Timestamp]), optional_
+
+> Timestamp indicating what time the verification occurred.
+
+<a id="resourceUri"></a>
+`resourceUri` _string ([ResourceURI]), required_
+
+> URI that identifies the resource associated with the artifact being verified.
+
+<a id="policy"></a>
+`policy` _object ([ResourceDescriptor]), required_
+
+> Describes the policy that the `subject` was verified against.
+>
+> The entry MUST contain a `uri`.
+
+<a id="inputAttestations"></a>
+`inputAttestations` _array ([ResourceDescriptor]), optional_
+
+> The collection of attestations that were used to perform verification.
+> Conceptually similar to the `resolvedDependencies` field in [SLSA Provenance].
+>
+> This field MAY be absent if the verifier does not support this feature.
+> If non-empty, this field MUST contain information on _all_ the attestations
+> used to perform verification.
+>
+> Each entry MUST contain a `digest` of the attestation and SHOULD contains a
+> `uri` that can be used to fetch the attestation.
+
+<a id="verificationResult"></a>
+`verificationResult` _string, required_
+
+> Either “PASSED” or “FAILED” to indicate if the artifact passed or failed the policy verification.
+
+<a id="verifiedProperties"></a>
+`verifiedProperties` _array (_string), required_
+
+> Indicates the properties verified for the artifact (and not
+> its dependencies), or "FAILED" if policy verification failed.
+>
+> If including SLSA levels, users MUST NOT include more than one level per SLSA
+> track. Note that each SLSA level implies all levels below it (e.g.
+> SLSA_BUILD_LEVEL_3 implies SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so
+> there is no need to include more than one level per track.
+
+<a id="dependencyProperties"></a>
+`dependencyProperties` _object, optional_
+
+> A count of the dependencies verified to have each property.
+>
+> Map from _string to the number of the artifact's _transitive_ dependencies
+> that were verified at the indicated level. Absence of a given value
+>  MUST be interpreted as reporting _0_ dependencies with that value.
+>
+> If including SLSA levels, users MUST count each dependency only once per SLSA
+> track, at the highest level verified. For example, if a dependency meets
+> SLSA_BUILD_LEVEL_2, you include it with the count for SLSA_BUILD_LEVEL_2 but
+> not the count for SLSA_BUILD_LEVEL_1.
+
+<a id="slsaVersion"></a>
+`slsaVersion` _string, optional_
+
+> Indicates the version of the SLSA specification that the verifier used, in the
+> form `<MAJOR>.<MINOR>`. Example: `1.0`. If unset, the default is an
+> unspecified minor version of `1.x`.
+
+## Example
+
+WARNING: This is just for demonstration purposes.
+
+```jsonc
+"_type": "https://in-toto.io/Statement/v1",
+"subject": [{
+  "name": "out/example-1.2.3.tar.gz",
+  "digest": {"sha256": "5678..."}
+}],
+
+// Predicate
+"predicateType": "https://in-toto.io/attestation/verification_summary/v1",
+"predicate": {
+  "verifier": {
+    "id": "https://example.com/publication_verifier",
+    "version": {
+      "slsa-verifier-linux-amd64": "v2.3.0",
+      "slsa-framework/slsa-verifier/actions/installer": "v2.3.0"
+    }
+  },
+  "timeVerified": "1985-04-12T23:20:50.52Z",
+  "resourceUri": "https://example.com/example-1.2.3.tar.gz",
+  "policy": {
+    "uri": "https://example.com/example_tarball.policy",
+    "digest": {"sha256": "1234..."}
+  },
+  "inputAttestations": [
+    {
+      "uri": "https://example.com/provenances/example-1.2.3.tar.gz.intoto.jsonl",
+      "digest": {"sha256": "abcd..."}
+    }
+  ],
+  "verificationResult": "PASSED",
+  "verifiedProperties": ["SLSA_BUILD_LEVEL_3"],
+  "dependencyProperties": {
+    "SLSA_BUILD_LEVEL_3": 5,
+    "SLSA_BUILD_LEVEL_2": 7,
+    "SLSA_BUILD_LEVEL_1": 1,
+  },
+  "slsaVersion": "1.0"
+}
+```
+
+<div id="verificationresult">
+
+## Change history
+
+-   1.1:
+    -   Added optional `verifier.version` field to record verification tools.
+    -   Added Verification section with examples.
+    -   Made `timeVerified` optional.
+-   1.0:
+    -   Replaced `materials` with `resolvedDependencies`.
+    -   Relaxed `VerificationResult` to allow other values.
+    -   Converted to lowerCamelCase for consistency with [SLSA Provenance].
+    -   Added `slsaVersion` field.
+-   0.2:
+    -   Added `resource_uri` field.
+    -   Added optional `input_attestations` field.
+-   0.1: Initial version.
+
+[SLSA Provenance]: /provenance
+[DigestSet]: https://github.com/in-toto/attestation/blob/7aefca35a0f74a6e0cb397a8c4a76558f54de571/spec/v1/digest_set.md
+[ResourceURI]: https://github.com/in-toto/attestation/blob/7aefca35a0f74a6e0cb397a8c4a76558f54de571/spec/v1/field_types.md#resourceuri
+[ResourceDescriptor]: https://github.com/in-toto/attestation/blob/7aefca35a0f74a6e0cb397a8c4a76558f54de571/spec/v1/resource_descriptor.md
+[Statement]: https://github.com/in-toto/attestation/blob/7aefca35a0f74a6e0cb397a8c4a76558f54de571/spec/v1/statement.md
+[Timestamp]: https://github.com/in-toto/attestation/blob/7aefca35a0f74a6e0cb397a8c4a76558f54de571/spec/v1/field_types.md#timestamp
+[TypeURI]: https://github.com/in-toto/attestation/blob/7aefca35a0f74a6e0cb397a8c4a76558f54de571/spec/v1/field_types.md#TypeURI
+[in-toto attestation]: https://github.com/in-toto/attestation
+[parsing rules]: https://github.com/in-toto/attestation/blob/7aefca35a0f74a6e0cb397a8c4a76558f54de571/spec/v1/README.md#parsing-rules
+[in-toto specification]: https://github.com/in-toto/docs/blob/master/in-toto-spec.md
--- a/spec/predicates/vsa.md
+++ b/spec/predicates/vsa.md
@@ -1,4 +1,4 @@
-# Predicate type: Verification Summary Attestation (VSA)
+# Predicate type: Verification Summary
 
 Type URI: https://in-toto.io/attestation/verification_summary/v1
 
@@ -8,8 +8,8 @@ Version: 1.1
 
 Verification summary attestations communicate that an artifact has been verified
 against some policy and details about that verification. Such details may
-include, but are not limited to, what SLSA level the artifact and/or its
-dependencies were verified to meet.
+include, but are not limited to, what [SLSA](https://slsa.dev) level the
+artifact and/or its dependencies were verified to meet.
 
 ## Use Cases
 
@@ -62,7 +62,7 @@ wishing to establish minimum requirements on dependencies may use
 }],
 
 // Predicate
-"predicateType": "https://slsa.dev/verification_summary/v1",
+"predicateType": "https://in-toto.io/attestation/verification_summary/v1",
 "predicate": {
   "verifier": {
     "id": "<URI>",
@@ -72,7 +72,7 @@ wishing to establish minimum requirements on dependencies may use
     }
   },
   "timeVerified": <TIMESTAMP>,
-  "resourceUri": <artifact-URI-in-request>,
+  "verifiedUri": <artifact-URI-in-request>,
   "policy": {
     "uri": "<URI>",
     "digest": { /* DigestSet */ }
@@ -91,7 +91,6 @@ wishing to establish minimum requirements on dependencies may use
     "<String>": <Int>,
     ...
   },
-  "slsaVersion": "<MAJOR>.<MINOR>",
 }
 ```
 
@@ -177,33 +176,16 @@ of the other top-level fields, such as `subject`, see [Statement]._
 `verifiedProperties` _array (_string), required_
 
 > Indicates the properties verified for the artifact (and not
-> its dependencies), or "FAILED" if policy verification failed.
->
-> If including SLSA levels, users MUST NOT include more than one level per SLSA
-> track. Note that each SLSA level implies all levels below it (e.g.
-> SLSA_BUILD_LEVEL_3 implies SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so
-> there is no need to include more than one level per track.
+> its dependencies).
 
 <a id="dependencyProperties"></a>
 `dependencyProperties` _object, optional_
 
 > A count of the dependencies verified to have each property.
 >
-> Map from _string to the number of the artifact's _transitive_ dependencies
+> Map from _string_ to the number of the artifact's _transitive_ dependencies
 > that were verified at the indicated level. Absence of a given value
 >  MUST be interpreted as reporting _0_ dependencies with that value.
->
-> If including SLSA levels, users MUST count each dependency only once per SLSA
-> track, at the highest level verified. For example, if a dependency meets
-> SLSA_BUILD_LEVEL_2, you include it with the count for SLSA_BUILD_LEVEL_2 but
-> not the count for SLSA_BUILD_LEVEL_1.
-
-<a id="slsaVersion"></a>
-`slsaVersion` _string, optional_
-
-> Indicates the version of the SLSA specification that the verifier used, in the
-> form `<MAJOR>.<MINOR>`. Example: `1.0`. If unset, the default is an
-> unspecified minor version of `1.x`.
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+    ...
+  ],
+  "verificationResult": "<PASSED|FAILED>",
+  "verifiedProperties": ["<String>"],
+  "dependencyProperties": {
+    "<String>": <Int>,
+    "<String>": <Int>,
+    ...
+  },
+  "slsaVersion": "<MAJOR>.<MINOR>",
```
TomHennen, line 94: Since this is no longer a slsa attestation should slsaVersion be dropped?  If necessary (and I'm not sure it is) it could just be included in the names of the various verifiedProperties.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+"predicate": {
+  "verifier": {
+    "id": "<URI>",
+    "version": {
+      "<COMPONENT>": "<VERSION>",
+      ...
+    }
+  },
+  "timeVerified": <TIMESTAMP>,
+  "resourceUri": <artifact-URI-in-request>,
```
TomHennen, line 75: optional: Would this be a good time to consider if we should change this name?

It's a bit confusing that this _field name_ also happens to match [this _type name_](https://github.com/in-toto/attestation/blob/99d851228fe284c66b2cde353a6693c5eff69db1/spec/v1/field_types.md?plain=1#L12).

Would `verifiedUri` be better??

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+<a id="verificationResult"></a>
+`verificationResult` _string, required_
+
+> Either “PASSED” or “FAILED” to indicate if the artifact passed or failed the policy verification.
+
+<a id="verifiedProperties"></a>
+`verifiedProperties` _array (_string), required_
+
+> Indicates the properties verified for the artifact (and not
+> its dependencies), or "FAILED" if policy verification failed.
```
TomHennen, line 180: I seem to remember @AdamZWu had some difficulties when implementing this.  Adam is it fine to keep this or would it be better to say the array should be empty?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+>
+> If including SLSA levels, users MUST NOT include more than one level per SLSA
```
TomHennen, line 182: Should this section be removed here and go with the SLSA docs?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+> track. Note that each SLSA level implies all levels below it (e.g.
+> SLSA_BUILD_LEVEL_3 implies SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so
+> there is no need to include more than one level per track.
+
+<a id="dependencyProperties"></a>
+`dependencyProperties` _object, optional_
+
+> A count of the dependencies verified to have each property.
+>
+> Map from _string to the number of the artifact's _transitive_ dependencies
```
TomHennen, line 192: nit: Should this be `_string_`

spec/predicates/vsa.md
TomHennen: Maybe it's easier to leave it as-is for now so that semantic changes are kept to a minimum?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+> that were verified at the indicated level. Absence of a given value
+>  MUST be interpreted as reporting _0_ dependencies with that value.
+>
+> If including SLSA levels, users MUST count each dependency only once per SLSA
```
TomHennen, line 196: I wonder if this text can be generalized.

> If including 'leveled' properties (e.g. SLSA), users MUST count each dependency once per property, at the highest level verified.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+    "SLSA_BUILD_LEVEL_2": 7,
+    "SLSA_BUILD_LEVEL_1": 1,
+  },
+  "slsaVersion": "1.0"
+}
+```
+
+<div id="verificationresult">
+
+## Change history
```
TomHennen, line 254: Include something about it moving here from slsa?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+    -   Replaced `materials` with `resolvedDependencies`.
+    -   Relaxed `VerificationResult` to allow other values.
+    -   Converted to lowerCamelCase for consistency with [SLSA Provenance].
+    -   Added `slsaVersion` field.
+-   0.2:
+    -   Added `resource_uri` field.
+    -   Added optional `input_attestations` field.
+-   0.1: Initial version.
+
+[SLSA Provenance]: /provenance
```
TomHennen, line 270: I think this link needs to be updated.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
 
-Version: 1.0
+Version: 1.1
 
 ## Purpose
 
-Records a SLSA verification decision about an artifact.
+Verification summary attestations communicate that an artifact has been verified
+against some policy and details about that verification. Such details may
+include, but are not limited to, what SLSA level the artifact and/or its
```
TomHennen, line 11: optional: link to SLSA?

spec/predicates/vsa.md
TomHennen, line 1: Can you bring over 'How to verify' too? https://slsa.dev/spec/v1.1/verification_summary#how-to-verify

spec/predicates/vsa.md
adityasaky: +1, moving this to verifiedProperties makes sense to me.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+> Indicates the properties verified for the artifact (and not
+> its dependencies), or "FAILED" if policy verification failed.
+>
+> If including SLSA levels, users MUST NOT include more than one level per SLSA
+> track. Note that each SLSA level implies all levels below it (e.g.
+> SLSA_BUILD_LEVEL_3 implies SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so
+> there is no need to include more than one level per track.
+
+<a id="dependencyProperties"></a>
+`dependencyProperties` _object, optional_
```
adityasaky, line 188: I think it'd be nice to have more guidance for this one too. I think it'll be needed as this generalizes beyond the SLSA-specific schema.

adityasaky: And possibly we'd want to define "dependency" in this spec as well?

adityasaky (commented):
Thanks for working on moving this over to in-toto!

spec/predicates/vsa.md
TomHennen: That's what I originally thought when we were developing this internally, but @AdamZWu made some compelling arguments for separating it.  At least one of those arguments was that it seemed to violate layering.

I think we also recently ran across a case and thought "oh wow, it's a good thing we have this separate field and don't have it in the subject".  Maybe something to do with having VSAs that have _lots_ of subjects but really only one resourceUri applies.  If it were in the subject it needs to be repeated for each digest even if it's the same thing and that can result in very large attestations.   @AdamZWu should more more details.

spec/predicates/vsa.md
AdamZWu: My main reason against having the "FAILED" property is described here: https://github.com/slsa-framework/slsa/issues/968#issuecomment-1733818744

But I am fine for leaving it as a future improvement rather than cram everything is a big change.

> Is this based entirely on the producer<->consumer to reconcile?

My understanding is, sort of. In the context of in-toto specs, yes, in-toto simply provides this field for the producers and consumers, there is no motivation to further regulate its uses. However, under the bigger picture of building a network of secure software supply chains, these are needs for standardization -- If organization A produces "apple" in `verifiedProperties` and organization B expects to see "orange", they cannot effectively exchange the trusted build data.

This, I think, is where SLSA comes in. SLSA defines a set of standardized trusted build properties and levels, which everyone agrees upon *and* mandates that in the VSA, so that, in addition to "apple", organization A must put "SLSA_..." properties in `verifiedProperties`; this makes it possible for organization B to consume that VSA -- they don't know what "apple" means, but they do know what "SLSA_..." means. :D

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+    "digest": { /* DigestSet */ }
+  }
+  "inputAttestations": [
+    {
+      "uri": "<URI>",
+      "digest": { <digest-of-attestation-data> }
+    },
+    ...
+  ],
+  "verificationResult": "<PASSED|FAILED>",
```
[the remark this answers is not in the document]
TomHennen: We do have some use cases for them internally like 1) having a secure record of the decision we made, 2) being a convenient way to propagate the error message back to users (though I don't think that field is in here).

As with most things VSA, @AdamZWu is the expert.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+  "verifier": {
+    "id": "<URI>",
+    "version": {
+      "<COMPONENT>": "<VERSION>",
+      ...
+    }
+  },
+  "timeVerified": <TIMESTAMP>,
+  "resourceUri": <artifact-URI-in-request>,
+  "policy": {
```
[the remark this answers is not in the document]
TomHennen: It is already a ResourceDescriptor (but not a list) according to the doc below.  I guess we just need to update this summary?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+    "uri": "<URI>",
+    "digest": { /* DigestSet */ }
+  }
+  "inputAttestations": [
+    {
```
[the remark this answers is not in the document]
TomHennen: It is, but maybe we just need to update the summary?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
-# Predicate type: SLSA Verification Summary (VSA)
+# Predicate type: Verification Summary Attestation (VSA)
```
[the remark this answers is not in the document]
TomHennen: Would it still be abbreviated VSA or would it be just VS?  (or maybe we don't abbreviate at all)

spec/predicates/vsa.md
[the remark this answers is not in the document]
AdamZWu: Agree with Tom. IMO having failed VSA allows more homogeneous processing of artifacts along the supply chain. 

One benefit of VSA is that it allows pruning of attestations -- when a VSA is produced the supply chain stage can simply forward the VSA downstream instead of hauling a big bundle of attestations for all transitive dependencies.

Even if verification failed, VSA still contains useful information, such as what specific checks failed in `verifiedProperties`; and producers can also put in some internal technical data in the custom fields.

Imagine if we don't produce VSA for failed verification, then the only way to reproduce the failure details is to re-perform the exact verification, which would require hauling the big bundle of attestations for all transitive dependencies downstream...

spec/predicates/vsa.md
TomHennen: Suggestion: if no one objects lets remove 'FAILED'.  If anyone does object we keep it?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+> SLSA_BUILD_LEVEL_3 implies SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so
+> there is no need to include more than one level per track.
+
+<a id="dependencyProperties"></a>
+`dependencyProperties` _object, optional_
+
+> A count of the dependencies verified to have each property.
+>
+> Map from _string to the number of the artifact's _transitive_ dependencies
+> that were verified at the indicated level. Absence of a given value
```
jkjell, line 193: If the dependencies are `verified` shouldn't that generate a VSA or some other evidence of its own to reference?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+>  MUST be interpreted as reporting _0_ dependencies with that value.
```
jkjell, line 194: Is this MUST specific to a particular verifier? Is there an expectations that all dependency properties are known in advance?

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
...
+in-toto attestation [Statement]) by evaluating the artifact and a `bundle`
+of attestations against some `policy`.  Users who trust the `verifier` may
+assume that the artifacts satisfied the indicated security polciy without
+themselves needing to evaluate the artifact or to have access to the
+attestations the `verifier` used to make its determination.
+
+The VSA also allows consumers to determine the verified levels of
+all of an artifact’s _transitive_ dependencies.  The verifier does this by
+either a) verifying the provenance of each non-source dependency listed in
+the [resolvedDependencies](/provenance/v1#resolvedDependencies) of the artifact
```
jkjell, line 45: I think this a link to SLSA Build provenance? There may be value for the generic VSA to define an internal version of a resolved dependency for dependencies without a provenance document. I'm currently thinking that along the lines that a dependency could be almost anything that is the subject of an attestation.

spec/predicates/vsa.md
TomHennen: Are you asking why we need this map when we could just include all the VSAs for the transitive dependencies?

spec/predicates/vsa.md
TomHennen: Hmm, that seems like it would be pretty hard to define and reason about.

But it does seem odd for this spec to depend on SLSA build provenance.  Maybe this is an example of why VSAs might stay better with SLSA?

spec/predicates/vsa.md
AdamZWu: I guess it would be quite hard to require every verifier to know all dependency properties in advance. :P

So yeah "Absence of a given value MUST be interpreted as reporting _0_ dependencies..." seems to be too strong of a requirement.

One possible fix is to introduce SLSA's "track" concept here (e.g. SLSA_BUILD_{0,1,2,3}):
- The presence of any value in a particular track means that the track has been verified;
- And within a verified track, we can safely have: "Absence of a given value MUST be interpreted as reporting _0_ dependencies..."

spec/predicates/vsa.md
kpk47: @jkjell You can include VSAs for each transitive dependency in the attestation bundle, but doing so for a nontrivial piece of software will mean creating a large attestation bundle that may be impractical to evaluate. Avoiding that giant bundle is kind of the point of a VSA.

That being said, I'm in favor of removing this field. It isn't clear to me why you'd want dependencies' security properties summarized in the VSA instead of implicitly through the evaluated policy. Alternatively, I could imagine gathering this data on a dashboard if you had a database of VSAs for the subject's transitive dependencies. Such a dashboard could be updated in real time based on whatever data source the VSA producer used to fill these fields. WDYT @TomHennen ?

spec/predicates/vsa.md
kpk47: As I commented above, my proposed solution is to delete the dependencyProperties field.

spec/predicates/vsa.md
kpk47: That could work, but I prefer to delete this text (even if we keep the field) and incorporate it into the SLSA spec's description of how to use a VSA for SLSA.

spec/predicates/vsa.md
kpk47: Done

spec/predicates/vsa.md
kpk47: Done

spec/predicates/vsa.md
kpk47: I moved the "How to verify" section from SLSA and updated it to use the new field names. I'm not sure I could write in much detail about how to verify using in-toto layouts. Could that be follow-on work?

spec/predicates/vsa.md
kpk47: I left a longer comment down below, but I'm wondering if we should just remove dependencies from VSAs. They have started to feel like a false start towards recursive SLSA, which I'd prefer to address in a dependency track or external monitoring.

spec/predicates/vsa.md
kpk47: Done.

We can define how to express SLSA version in the SLSA spec. We're going to need a page on how to use VSAs with SLSA, so we can work out the details there.

spec/predicates/vsa.md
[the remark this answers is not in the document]
TomHennen: If we wanted to update this it could say

Maybe it should match what's done here? https://slsa.dev/spec/v1.0/provenance#schema

`"inputAttestations": [ ...#ResourceDescriptor ],`

spec/predicates/vsa.md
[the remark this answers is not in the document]
TomHennen: Or `"policy": #ResourceDescriptor,` to match https://slsa.dev/spec/v1.0/provenance#schema

spec/predicates/vsa.md
TomHennen: Since VSAs are often used to cross organization boundaries I don't think it's always policy to know the actual policy used across all orgs with all dependencies, likewise it may not be possible to have a real-time dashboard that could provide these feedback cross-org.

Certainly in some cases it may work (seems like a good use case for GUAC), but leaving it in VSAs would solve that problem.

We certainly still have this use case internally so we'd need to support it, maybe by adding an org specific field.  My preference would be to keep it, perhaps as a SLSA specific field addition, but we can probably work around it (I'd just be sad).

> @jkjell You can include VSAs for each transitive dependency in the attestation bundle, but doing so for a nontrivial piece of software will mean creating a large attestation bundle that may be impractical to evaluate. Avoiding that giant bundle is kind of the point of a VSA.

When consuming a piece of software via its VSA you shouldn't need any additional attestations.  So just one VSA.

If you're verifying the provenance of some piece of software though, then you may want VSAs for each of the _direct_ dependencies (which will cover their transitive dependencies).  You don't need them all.  So it shouldn't wind up being that big.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,358 @@
...
+
+<a id="resourceUri"></a>
+`resourceUri` _string ([ResourceURI]), required_
+
+> URI that identifies the resource associated with the artifact being verified.
+
+<a id="policy"></a>
+`policy` _object ([ResourceDescriptor]), required_
+
+> Describes the policy that the `subject` was verified against.
```
jkjell, line 153: Nit: We should refer to the `resource` of the `resourceUri` or change `resourceUri` to `subjectUri` just to be painfully obvious about the relationship between these two.

spec/predicates/vsa.md
jkjell: > VSAs are often used to cross organization boundaries I don't think it's always policy to know the actual policy used across all orgs with all dependencies

@TomHennen do you mean the policy of the dependency that is verified or the policy of the VSA itself? If policy is unknown for the VSA, then the VSA is nothing more than a signature for external parties. I suppose that might be ok in some situations but, it's a reversion to implicit trust.

spec/predicates/vsa.md
jkjell: I agree with removing dependencyProperties. Perhaps it would be worth considering a more generic policy result? We could allow verifiedProperties to be an object which encapsulates verifier specific information. This could also improve trust without knowing the contents of a policy.

spec/predicates/vsa.md
TomHennen: @kpk47 and I spoke offline, it sounds like a reasonable approach is:

1. Remove `dependencyProperties` from this predicate type (to avoid in-toto needing to define the concept of dependencies and other confusing problems)
2. Have SLSA define something like the `dependencyProperties` field _when using_ the in-toto VSA following [the existing in-toto guidance on extension fields](https://github.com/in-toto/attestation/blob/main/spec/v1/README.md#parsing-rules:~:text=considered%20major%20versions.-,Extension%20fields%3A,-Producers%20MAY%20add).

Does this sound reasonable to everyone?

jkjell:
@mikhailswift has also been working on our implementation of VSAs in `go-witness`

spec/predicates/vsa.md
TomHennen: > > VSAs are often used to cross organization boundaries I don't think it's always policy to know the actual policy used across all orgs with all dependencies
> 
> @TomHennen do you mean the policy of the dependency that is verified or the policy of the VSA itself? If policy is unknown for the VSA, then the VSA is nothing more than a signature for external parties. I suppose that might be ok in some situations but, it's a reversion to implicit trust.

I think maybe removing the thing resolves this, but I don't think it's totally true to say that it's "nothing more than a signature".  I agree that it would be ideal if people can see the original evidence and policies that were used throughout the entire supply chain (and the VSA, provenance, and other attestations let people do that), but it won't always be possible (some orgs may not be willing or able to make that data public because of proprietary information).

However, even if they can't make the policy and other attestations public, the VSA still includes the `verifiedProperties` fields and the `verifier.id` field, which means that the VSA is better than a plain signature in one important respect, it makes it clear that the verifier (`verified.id`) is asserting that the software covered by the VSA has some specific but high level properties (`verifiedProperties`).  That can be pretty useful when orgs are exchanging software?

So, the way I see it, plain signature < VSA without public evidence or policy < VSA with public evidence and policy?

But maybe this isn't the place to have that discussion?

jkjell: I think the `verifiedProperties` was the piece I was missing. Establishing trust in a verifier AND that it asserts a given set of properties were confirmed by a policy is a great way of abstracting the value of the policy from the policy itself.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,358 @@
...
+    "digest": { /* DigestSet */ }
+  }
+  "inputAttestations": [
+    {
+      "uri": "<URI>",
+      "digest": { <digest-of-attestation-data> }
+    },
+    ...
+  ],
+  "verificationResult": "<PASSED|FAILED>",
```
MarkLodato, line 87: I suggest we remove this field. If there is a notion of "pass" or "fail", one can create an entry in `verifiedProperties` to represent that with more specific semantics.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,279 @@
-# Predicate type: SLSA Verification Summary (VSA)
+# Predicate type: Verification Summary Attestation (VSA)
 
-Type URI: https://slsa.dev/verification_summary/v1
+Type URI: https://in-toto.io/attestation/verification_summary/v1
 
-Version: 1.0
+Version: 1.1
```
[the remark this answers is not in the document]
arewm: +1, I would expect the 1.0 SLSA VSA attestation to be deprecated in favor of a 0.1 in-toto VSA since there is a fair amount of incompatible changes between these two.

spec/predicates/vsa.md
arewm: Transitive dependency verification can also be achieved by having a policy engine which performs the tree walking. It can be abstracted to be represented by the policy's URI.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,358 @@
...
+"predicate": {
+  "verifier": {
+    "id": "<URI>",
+    "version": {
+      "<COMPONENT>": "<VERSION>",
+      ...
+    }
+  },
+  "timeVerified": <TIMESTAMP>,
+  "verifiedUri": <artifact-URI-in-request>,
```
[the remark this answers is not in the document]
arewm: Is this different from the subject in the statement above? @jkjell suggests changing to `subjectUri` below: https://github.com/in-toto/attestation/pull/295/files#r1393385717 -- I was thinking something similar.

spec/predicates/vsa.md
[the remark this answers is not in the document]
arewm: @adityasaky , I agree that it makes sense to support lists of policies. Even in the SLSA use case, there may be specific individual policies for each track. It would be beneficial to be able to aggregate all policies into a single VSA. There could additionally be policies that are owned/maintained by various organizations -- being able to differentiate between them (and their results) separately seems like a good direction for transparency.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,358 @@
...
+    "digest": {"sha256": "1234..."}
+  },
+  "inputAttestations": [
+    {
+      "uri": "https://example.com/provenances/example-1.2.3.tar.gz.intoto.jsonl",
+      "digest": {"sha256": "abcd..."}
+    }
+  ],
+  "verificationResult": "PASSED",
+  "verifiedProperties": ["SLSA_BUILD_LEVEL_3"],
```
arewm, line 224: The relationship between the policy and verification results that I find myself expecting is:

* `policy` is a list of policies that are considered in the VSA
* `verificationResult` is the net result -- whether all policies pass or fail
* `verificationProperties` is some indication about the result from each policy taken into account (i.e. whether it is pass or fail)

What might be better, however, is to have keys following the pattern:
*  `policies` - the list of all policies that have been applied to create this VSA
*  `result` - the ultimate result of all policies combined
*  `policyResults` - the individual result of each policy applied. There would have to be easy mapping from each policy to its result, but this could potentially be done by assigning a unique name for each policy so that the full uri/digest don't have to be repeated.

Outside of the SLSA specification, it might be harder to realize value from a field like `verifiedProperties`.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,358 @@
...
+
+Understanding of [in-toto specification] and the in-toto attestation framework.
+
+## Model
+
+A Verification Summary Attestation (VSA) is an attestation that some entity
+(`verifier`) verified one or more software artifacts (the `subject` of an
+in-toto attestation [Statement]) by evaluating the artifact and a `bundle`
+of attestations against some `policy`.  Users who trust the `verifier` may
+assume that the artifacts satisfied the indicated security polciy without
```
lcarva, line 38: Minor typo: "polciy"

spec/predicates/vsa.md
```diff
@@ -1,11 +1,358 @@
...
+artifact or all of its transitive dependencies.  They delegate
+complex policy decisions to some trusted party and then simply trust that
+party's decision regarding the artifact. That decision is represented by a VSA.
+
+VSAs also allow software producers to keep the details of their build pipeline
+confidential while still communicating that some verification has taken place.
+This might be necessary for legal reasons (keeping a software supplier
+confidential) or for security reasons (not revealing that an embargoed patch has
+been included).
+
```
lcarva, line 27: Another use case is that verification sometimes can be costly and not suitable for certain scenarios. 

Something like [Enterprise Contract](https://enterprisecontract.dev/) can evaluate policies based on an image's labels or even one of its file's contents. This wouldn't really be possible to perform as a Kubernetes admission web hook.

spec/predicates/vsa.md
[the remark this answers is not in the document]
lcarva: A list would be nice, but the same could also be achieved by creating multiple VSAs. We should explore which option makes the most sense.

spec/predicates/vsa.md
```diff
@@ -1,11 +1,358 @@
...
+    "id": "<URI>",
+    "version": {
+      "<COMPONENT>": "<VERSION>",
+      ...
+    }
+  },
+  "timeVerified": <TIMESTAMP>,
+  "verifiedUri": <artifact-URI-in-request>,
+  "policy": {
+    "uri": "<URI>",
```
lcarva, line 77: I would like the ability to provide an inline policy. This would account for certain scenarios where the policy is adjusted dynamically and may not have a published URI. I think this could be achieved with a [data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs). Thoughts? If that's the intention, it would be great to clarify somewhere in this doc.

spec/predicates/vsa.md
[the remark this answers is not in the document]
jkjell: @kpk47 In the case of Witness policies (and I think in-toto layouts), you would have keys to sign the policy itself, and potential keys for each functionary that performed an operation. The later could be embedded in the policy but, the signing key for the policy itself is an out-of-band communication (needs a separate resource description).

As far as multiple policies go... I think it's going to get complicated fast. What logic is applied between multiple policies? You could defer that decision making process external to the VSA if you're only capturing discrete results per policy but, for me, that's counter to the purpose of capturing a policy decision in a VSA. I think that @lcarva is on the write track with additional VSAs. Each could be inputs to a subsequent policy or future decision.

spec/predicates/vsa.md
jkjell: > the ultimate result of all policies combined

It's not clear to me how this should actually be calculated.

For my intended consumption of this, trust in a VSA will be based on a 1:1 mapping of trust in a policy. If I revoke trust in a policy (i.e. just to update the policy). I will revoke trust in all VSA's generated from that policy. Not arguing against the use case, I don't think we would ever use multiple policies in a single VSA. I'm curious to hear from others.

AdamZWu: Could we consider `policy`+`verificationResult` vs. `verifiedProperties` as two sibling structures to convey trust? i.e.:

- One could decide to delegate trust on a specific policy (if they know its content meets their needs), and just verify that: 1) the `policy` that they are interested has been verified; and 2) the overall `verificationResult` is a PASS.

- Alternatively, one could decide they accept any artifact that meets a certain set of `verifiedProperties`, e.g. SLSA_BUILD_3 and SLSA_SOURCE_2 (imaginary), and not caring what exact policy triggered the verification of those properties. In that case, they will only be looking up their expected keys in `verifiedProperties`.

Of course, the latter requires the keys in `verifiedProperties` to be arbitrated (i.e. having a higher-level protocol like SLSA). Otherwise any verifier can put any random text in it that can be misunderstood by other verifiers.

---
OTOH, if we remove the `verificationResult` as @MarkLodato suggests [here](#discussion_r1394481717), then `verifiedProperties` becomes the only facility to convey trust.

spec/predicates/vsa.md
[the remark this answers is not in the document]
arewm: > What logic is applied between multiple policies?

That would depend on the policy engine responsible for creating the VSA. If multiple policies are given as input to a decision, the VSA-generating tool would be responsible for reconciling any conflicts/inconsistencies and generating the ultimate pass/fail. Even if a policy engine supports multiple policies as input, it might elect to create multiple VSAs. 

> You could defer that decision making process external to the VSA if you're only capturing discrete results per policy but, for me, that's counter to the purpose of capturing a policy decision in a VSA.

Depending on the cross-relationship between different policies for the policy engine, this might not be straightforward. Another potential approach could be for an engine to create pseudo policies that are composed of multiple interrelated sub policies and just report _that_. One potential use case for multiple policies would be if there were policy modifications on-the-fly, i.e. using inline policies as mentioned by @lcarva: https://github.com/in-toto/attestation/pull/295#discussion_r1414559935.

spec/predicates/vsa.md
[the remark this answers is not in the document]
adityasaky: Since the URI's domain is changing, can we drop the predicate version to v0.x?

spec/predicates/vsa.md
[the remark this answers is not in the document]
adityasaky: Apologies for the delay, my notifications have been a mess the last 3 weeks or so.

@jkjell covered my point about policy keys above. I'm actually less convinced a VSA needs to cover multiple actual policies. I think a verification workflow should be scoped per policy artifact (and applicable attestations), so if multiple policies are verified, we should generate multiple VSAs. This, to me, makes sense given the expected use cases for VSAs: record of verification elsewhere in the supply chain or for policy delegations like [in-toto's sublayouts](https://github.com/in-toto/docs/blob/master/in-toto-spec.md#45-specifying-sublayouts).

spec/predicates/vsa.md
adityasaky: The resource descriptor allows for that using `content`: https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md.

spec/predicates/vsa.md
adityasaky: > For my intended consumption of this, trust in a VSA will be based on a 1:1 mapping of trust in a policy. If I revoke trust in a policy (i.e. just to update the policy). I will revoke trust in all VSA's generated from that policy. Not arguing against the use case, I don't think we would ever use multiple policies in a single VSA. I'm curious to hear from others.

@jkjell if policy A delegates to policy B, and therefore verification of policy A results in VSA A and VSA B, I don't think we can claim that revoking policy A automatically revokes policy B, at least in the in-toto sublayout context. Policy B may be delegated to by C, D, E, etc as well, so IMO VSA B is fine to continue to trust in isolation. Instead, as your verification starts at policy A (the root policy), you'd not consider VSA B in that instance of the workflow, rather than _revoking_ VSA B.

marcelamelara (changes requested):
@kpk47 Thanks for getting this started. As others have noted, since we're looking to define this as a predicate separate from the existing SLSA-defined VSA, I'd recommend this PR introduce a brand new spec doc for the VS predicate that is separate from the VSA, rather than editing on top of the existing spec.

marcelamelara:
Closing until we decide next steps for a resolution here. Let's pick up the discussion back in #277.

spec/predicates/vsa.md
arewm: Looking back at the unresolved comments in this PR where we previously tried to move the VSA predicate. Would it not just be reasonable to have one VSA per predicate?

trishankatdatadog: Oof, yeah, this PR ran too long. Since it's closed, is it OK if we start a fresh discussion (not an issue or PR yet)?

Issue #298, opened 2023-11-26
VSA is at v0.2 instead of v1


trishankatdatadog:
As the title says, the VSA [protobuf](https://github.com/in-toto/attestation/blob/f73239c75066685e954052ac74bb4918bf9e6233/protos/in_toto_attestation/predicates/vsa/v0/vsa.proto) (and generated [Go](https://github.com/in-toto/attestation/blob/f73239c75066685e954052ac74bb4918bf9e6233/go/predicates/vsa/v0/vsa.pb.go) code) is at v0.2, although the [documentation](https://github.com/in-toto/attestation/blob/f73239c75066685e954052ac74bb4918bf9e6233/spec/predicates/vsa.md) refers to v1.

PR #302, opened 2023-12-08
update vsa proto to v1 and update generated code

pxp928:
fixes https://github.com/in-toto/attestation/issues/298


protos/in_toto_attestation/predicates/vsa/v0/vsa.proto
mikhailswift, line 1: This may be my ignorance, and this isn't something you touched but something I noticed while looking over this, but it seems that VSA v1.0 uses camel case names, whereas this protobuf is using snake casing.

Example `inputAttestations` is what https://slsa.dev/spec/v1.0/verification_summary defines, where as this protobuf defines the json name as `input_attestations`

protos/in_toto_attestation/predicates/vsa/v0/vsa.proto
```diff
@@ -8,7 +8,7 @@ option go_package = "github.com/in-toto/attestation/go/predicates/vsa/v0";
 option java_package = "io.github.intoto.attestation.predicates.vsa.v0";
 
 
-// Proto representation of predicate type https://slsa.dev/verification_summary/v0.2
+// Proto representation of predicate type https://slsa.dev/verification_summary/v1
```
marcelamelara, line 11: I think this only changes the documented version, but what I think @trishankatdatadog is pointing out is that this whole VSA proto definition still follows the v0.2 spec. So I think we might need to create separate v1 proto.

marcelamelara (commented):
I think the

protos/in_toto_attestation/predicates/vsa/v0/vsa.proto
pxp928: I will work to get a new v1 proto created

protos/in_toto_attestation/predicates/vsa/v0/vsa.proto
pxp928: fixed.

protos/in_toto_attestation/predicates/vsa/v0/vsa.proto
pxp928: created new v1

marcelamelara (approved):
LGTM, thanks @pxp928 !

trishankatdatadog (approved):

PR #314, opened 2023-12-19
Add python venv dir to gitignore

adityasaky:
https://github.com/in-toto/attestation/pull/308 checks in a full virtual environment, which we ought to avoid...

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -30,7 +30,7 @@ all:
 dev: $(VENV_STAMP)
 
 $(VENV_STAMP): pyproject.toml
-	python -m venv env
+	python -m venv $(VENV)
 	$(VENV_BIN)/python -m pip install --upgrade pip
 	$(VENV_BIN)/python -m pip install -e .[$(INSTALL_EXTRA)]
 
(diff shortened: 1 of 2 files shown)


marcelamelara:
I think this is a needed change, but the more robust solution is to have the Python tests clean up the venv when they're done, I think.

marcelamelara (approved):
Thanks @adityasaky !

python/Makefile
```diff
@@ -61,3 +61,7 @@ package: $(VENV_STAMP)
 .PHONY: edit
 edit:
 	$(EDITOR) $(ALL_PY_SRCS)
+
+.PHONY: clean
```
pxp928, line 65: should this not be called after the test has completed?

marcelamelara: @adityasaky and I were talking about this earlier, too. I think it's good practice to clean up after the tests, but for local development it could make sense to leave the virtual environment.

pxp928: ah ok sounds good.

pxp928 (approved):

PR #319, opened 2024-01-11
Add Release Attestation

steiza:
This pull request proposes a Release Attestation, following the process outlined by [ITE-9].

The purpose of this attestation is to authoritatively link from a specific release name and version string in a registry, to the artifact names and hashes that make up that release. This allows consumers of that release version to ensure the artifacts they are consuming have not been tampered with.

This is a formalization of what npm does today with what it calls the [publish attestation].

Feedback welcome!

[ITE-9]: https://github.com/in-toto/ITE/tree/master/ITE/9
[publish attestation]: https://github.com/npm/attestation/tree/main/specs/publish/v0.1


spec/predicates/release.md
```diff
@@ -0,0 +1,95 @@
...
+
+This predicate depends on the [in-toto attestation framework], as well as the [purl-spec] for identifying packaging ecosystems.
+
+Perhaps surprisingly, this predicate does not depend on [SLSA Provenance], but they are better together. In an ecosystem where some (but not all) packages have SLSA Provenance and you query by artifact hash, you can't tell the difference between an artifact that has been tampered with and one that does not yet have SLSA provenance. If the ecosystem has release attestations, you can give authoritative answers to what artifacts make up a given release version, and what hashes those artifacts should have.
+
+## Model
+
+This predicate is for the final stages of the software supply chain, where consumers are looking for attestations that the software has not been tampered with during distribution.
+
+## Schema
```
kommendorkapten, line 29: Should we be prescriptive about what goes into the `name` field of the `subject`? I.e file on disk, or the native name for a package in the registry. In the examples below, there are two versions, one with a package name and one with filenames. To simplify and avoid ambiguity I would lean towards a requirement of the `name` being the file on disk. Thoughts?

kommendorkapten:
During another discussion with @phillmv, for certain ecosystems such as Go there may not be an artifact, the release is the source tree. [Git commits](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag) can have their digest computed, so that may be a good candidate for the digest in the subject. What would a good name be, the name of the release (commonly a tag name)?

spec/predicates/release.md
```diff
@@ -0,0 +1,95 @@
...
+  "subject": [
+    {
+      "name": "@scope/my-package",
+      "digest": {
+        "sha256": "deadbeef0123456789..."
+      }
+    }
+  ],
+  "predicateType": "https://in-toto.io/attestation/release/v0.1",
+  "predicate": {
```
jkjell, line 54: Is there any metadata about the `authorized user` that could be added to the attestation? This could allow consumers to revoke trust for this attestation in the future, based on a policy.

spec/predicates/release.md
```diff
@@ -0,0 +1,95 @@
...
+## Model
+
+This predicate is for the final stages of the software supply chain, where consumers are looking for attestations that the software has not been tampered with during distribution.
+
+## Schema
+
+### Fields
+
+- **purl, required** string (ResourceURI)
+  - A purl uniquely identifying a specific release name and version from a package registry.
```
trishankatdatadog, line 34: Should this be as specific as a complete URL with a domain name then? I would imagine so.

spec/predicates/release.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+If a release has multiple artifacts that might be consumed separately, the attestation SHOULD have a subject per artifact:
+
+```json
+{
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [
+    {
+      "name": "urllib3-2.1.0.tar.gz",
```
trishankatdatadog, line 67: I might also recommend a file size here to prevent [endless data attacks](https://theupdateframework.github.io/specification/v1.0.33/index.html#goals-to-protect-against-specific-attacks).

spec/predicates/release.md
```diff
@@ -0,0 +1,95 @@
...
+
+To authoritatively link from a specific release name and version string in a registry, to the artifact names and hashes that make up that release.
+
+## Use Cases
+
+When receiving a new release version, package registries can publish a release attestation covering the artifact names and hashes that make up that release. This allows consumers of that release version to ensure the artifacts they are consuming have not been tampered with.
+
+If these release attestations are optionally published to a transparency log, package authors (or other interested parties) can monitor when a new version of a package is released.
+
+These use cases are not hypothetical; both of them are the case today for npm's [build provenance feature], which includes a [publish attestation]. Note that while npm calls this a publish attestation, calling it a release attestation better reflects that it's coming from the package registry. Publishing often refers to an author sending content to the registry, as in [PyPI's trusted publishers feature].
```
laurentsimon, line 17: A publish attestation seems like a better term to me, since a registry publishes artifacts, whereas users and organization release software, eg on GitHub.

spec/predicates/release.md
MarkLodato: I agree with file on disk.

spec/predicates/release.md
steiza: For the npm publish attestation (and npm build provenance) we deliberately omit account identity for several reasons:

- Strongly identifying and uniquely mapping humans to digital identities is a very hard problem at global scale
- Account behavior changes over time: either because people change or because their account is compromised
- Even if you had account identity here, how would generate (and distribute in a trustworthy way) what accounts should be trusted over what period of time, in order to write policy for a given release?

That's not to say that these aren't problems worth addressing in the future. But I feel strongly that they are out of scope of this iteration of release attestations.

TomHennen:
This is one of the use cases I've imagined for [VSAs](https://slsa.dev/spec/v1.0/verification_summary).

[`resourceUri`](https://slsa.dev/spec/v1.0/verification_summary#resourceUri) would contain the purl, [`verificationResult`](https://slsa.dev/spec/v1.0/verification_summary#verificationResult) would be 'PASSED' and [`verifiedLevels`](https://slsa.dev/spec/v1.0/verification_summary#verifiedLevels) could be empty.

I've heard feedback that one reason folks might not want to use VSAs for this purpose is that they don't want be in position of evaluating SLSA policy themselves (fair).

The VSA format is also currently probably more complex than it needs to be (I'm working on how it can be significantly simplified).

Is there any other reason you wouldn't use a VSA for this purpose?

spec/predicates/release.md
```diff
@@ -0,0 +1,95 @@
...
+
+Version 0.1
+
+## Purpose
+
+To authoritatively link from a specific release name and version string in a registry, to the artifact names and hashes that make up that release.
+
+## Use Cases
+
+When receiving a new release version, package registries can publish a release attestation covering the artifact names and hashes that make up that release. This allows consumers of that release version to ensure the artifacts they are consuming have not been tampered with.
```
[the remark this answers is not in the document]
steiza: This is essentially the same as the npm publish attestation, but generalized to work for many / most ecosystems.

npm only ever has one artifact for a release version, the `dist.tarball` that consumers don't usually interact with directly.

But many ecosystems (like PyPI) support multiple artifacts for a single release (e.g. `.tar.gz`, `.whl`) and other ecosystems might have pre-build binaries per platform, so we wanted to have the option to attest to more than one artifact.

I'd suggest you attest to whatever unit the end-user is consuming from the repository, so more like supporting multiple tarballs a user would choose between when downloading, and *not* attesting to each file in a tarball.

spec/predicates/release.md
laurentsimon: yes naming is hard. We plan a work stream for Q1 in SLSA (or supply-chain) WG to align on naming conventions across the supply-chain. I think it'd be beneficial to work together on it and circle back on this PR after we've reached a broad consensus. Wdut?

@MarkLodato and @kpk47 will have more info

spec/predicates/release.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Model
+
+This predicate is for the final stages of the software supply chain, where consumers are looking for attestations that the software has not been tampered with during distribution.
+
+## Schema
+
+### Fields
+
+- **name, required** string
```
MarkLodato, line 33: nit: clarify that this is in the subject but the other field is in the predicate. Also would be good to use code font.

```suggestion
- **`subject.name`, required** string
```

spec/predicates/release.md
```diff
@@ -0,0 +1,98 @@
...
+  - The filename of the artifact as it would appear on disk.
+
+- **purl, required** string (ResourceURI)
```
MarkLodato, line 36:
```suggestion
- **`predicate.purl`, required** string (ResourceURI)
```

spec/predicates/release.md
MarkLodato, line 33: Sanity check: Is this truly required? Is there ever a legitimate situation where a filename is meaningless?

spec/predicates/release.md
steiza: Ideally this field would uniquely identify a release name and version. purl isn't exactly this, but if you make the version string required (which is optional in the purl spec) and leave off the qualifiers / subpath, I think it can uniquely identify a release name and version.

URLs are a bit too free-form: should you point to the README? The download page? It's very easy to construct multiple distinct URLs that point to the same resource.

The downside of using purls is that ecosystems would need to be accepted as a [known purl type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst) which could be seen as a barrier to entry. But I think there's also value in this formalization.

spec/predicates/release.md
steiza: This is a really interesting suggestion!

Technically this is part of the [statement layer](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md#statement-layer-specification) (specifically the [resource description](https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md)) and not the proposed release attestation predicate. Maybe an in-toto maintainer can help us figure out how to propose this change to the resource description (or if it could just be part of the existing `annotations`)?

However, I think statement layer proposals are probably separate from predicate proposals, even if they inform one another.

spec/predicates/release.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Prerequisites
+
+This predicate depends on the [in-toto attestation framework], as well as the [purl-spec] for identifying packaging ecosystems.
+
+Perhaps surprisingly, this predicate does not depend on [SLSA Provenance], but they are better together. In an ecosystem where some (but not all) packages have SLSA Provenance and you query by artifact hash, you can't tell the difference between an artifact that has been tampered with and one that does not yet have SLSA provenance. If the ecosystem has release attestations, you can give authoritative answers to what artifacts make up a given release version, and what hashes those artifacts should have.
+
+## Model
+
+This predicate is for the final stages of the software supply chain, where consumers are looking for attestations that the software has not been tampered with during distribution.
```
MarkLodato, line 27: Could you clarify the semantics around multiple attestations for a single release?

- Is a registry allowed to issue multiple attestations for a given `{purl, name}` pair? (I'm guessing that is legal if and only if the registry's semantics have mutable releases?)
- Are you allowed to issue separate attestations per file (rather than one attestation with all files)? If so, is one form recommended over the other?

spec/predicates/release.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Schema
+
+### Fields
```
MarkLodato, line 31: suggestion: Should we add an optional timestamp indicating when the package was released?

steiza:
> Is there any other reason you wouldn't use a VSA for this purpose?

Yes, so from the proposal we said: "Perhaps surprisingly, this predicate does not depend on SLSA Provenance, but they are better together." We want ecosystems to be able to use release attestations completely independently from if they support SLSA provenance or not.

This isn't because we don't like SLSA provenance (quite the opposite!) Rather, we're looking for security capabilities that are easier for ecosystems to make use of on their way towards SLSA provenance.

spec/predicates/release.md
MarkLodato: Should you state here that this MUST include a version, SHOULD include a qualifier if and only if it's necessary to scope a package name (as is the case for `oci`, for example), SHOULD NOT include a subpath?

spec/predicates/release.md
steiza: Ah yes, I should have mentioned immutability before! I've added clarification to structuring the subject for mutable and immutable releases below.

spec/predicates/release.md
steiza: I'm on the fence, but leaning towards no. It's unlikely a bare in-toto attestation will be hanging around; they should be sent to some transparency log or otherwise made tamper-resistant.

In the case of npm publish attestations they are published to the public-good Sigstore Rekor, which of course includes a witnessing timestamp. Does the in-toto attestation itself need a timestamp? Can you describe the use-case where the attestation itself needs the timestamp?

spec/predicates/release.md
steiza: Good suggestion - I've updated the Parsing Rules below to be more explicit.

spec/predicates/release.md
steiza: I thought it was required, but https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md#fields says it's optional, so I have updated this.

spec/predicates/release.md
laurentsimon: Here's an example of another publish attestation proposal https://github.com/trailofbits/homebrew-attestation/tree/main/specs/publish/v0.1, which would also benefit from broader alignment on terminology and goals for each attestation

spec/predicates/release.md
```diff
@@ -0,0 +1,102 @@
...
+
+Type URI: https://in-toto.io/attestation/release
+
+Version 0.1
+
+## Purpose
+
+To authoritatively link from a specific release name and version string in a registry, to the artifact names and hashes that make up that release.
+
+## Use Cases
```
MarkLodato, line 11: Could you go into slightly more detail about what this is and is not intended to protect against? IIUC:

- Protects against a mirror substituting a malicious artifact (i.e. authenticity / integrity).
- Does not protect against a mirror pretending that a release, or an artifact within a release, does not exist (i.e. availability).

Not sure if there's more nuance to it than that.

MarkLodato (commented):
Thanks again

spec/predicates/release.md
```diff
@@ -26,16 +26,23 @@ Perhaps surprisingly, this predicate does not depend on [SLSA Provenance], but t
 
 This predicate is for the final stages of the software supply chain, where consumers are looking for attestations that the software has not been tampered with during distribution.
 
+If a registry supports immutable releases, there MUST be one release attestation for a given `predicate.purl`.
+
+Even if the registry does not support immutable releases, the attestation subject SHOULD include all the artifacts associated with the release; otherwise it will be unclear if an artifact was later removed from a release.
```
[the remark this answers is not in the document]
steiza: I added clarification on how each attestation should include all known artifacts at that time.

spec/predicates/release.md
steiza: I added language around protecting integrity and not protecting availability.

spec/predicates/release.md
steiza: > If the intent is to always put this on a transparency log, should we instead be looking toward generalizing the go checksum database rather than creating an attestation type

The way I view this is the release attestation allows registries the benefit / possibility of using a shared transparency log (like Sigstore public-good Rekor) without each registry having to maintain their own transparency log.

If registries want to maintain a transparency log - great! If we want to make it easier for registries to maintain their own transparency log - also great! But today it's a heavy lift for a registry to maintain their own transparency log.

In a similar vein, https://github.com/repository-service-tuf/repository-service-tuf is working towards making it easier for repositories to maintain their own TUF root that signs the repository index metadata. That project has a similar goal to release attestations - securely distributing release artifact hashes - but goes about it in a very different way.

steiza:
Do note that in the update I pushed today I added an option predicate, `releaseId` to help prevent confusion when a release is renamed, or when a release is deleted and a new release is created with the same name.

steiza:
> I'm not asking about the SLSA provenance, but rather the verification summary attestation which doesn't have to imply any SLSA evaluation has taken place (there's some discussion of making it more abstract and making it a more generic attestation)

Making VSA more abstract is an interesting idea. But today, the top line of https://slsa.dev/spec/v1.0/verification_summary says:

> Verification summary attestations communicate that an artifact has been verified at a specific SLSA level and details about that verification.

I'm happy to follow along as the VSA conversation continues, but I don't think we should block this proposal of release attestations as an in-toto predicate on those VSA conversations.

spec/predicates/release.md
```diff
@@ -0,0 +1,110 @@
...
+
+This predicate depends on the [in-toto attestation framework], as well as the [purl-spec] for identifying packaging ecosystems.
+
+Perhaps surprisingly, this predicate does not depend on [SLSA Provenance], but they are better together. In an ecosystem where some (but not all) packages have SLSA Provenance and you query by artifact hash, you can't tell the difference between an artifact that has been tampered with and one that does not yet have SLSA provenance. If the ecosystem has release attestations, you can give authoritative answers to what artifacts make up a given release version, and what hashes those artifacts should have.
+
+## Model
+
+This predicate is for the final stages of the software supply chain, where consumers are looking for attestations that the software has not been tampered with during distribution.
+
+If a previously attested release is later determined to be untrustworthy, a registry MAY stop serving the artifacts associated with that release.
```
marcelamelara, line 31: I know this says "MAY", but I'm still not sure this assumption belongs in the predicate spec, at least not as currently worded, since this is really registry policy. Is the intent here really to express the availability assumptions also stated above in the Use Cases section?

spec/predicates/release.md
```diff
@@ -0,0 +1,110 @@
...
+
+If a registry supports immutable releases, there SHOULD be one release attestation for a given `predicate.purl`.
+
+Even if the registry does not support immutable releases, the attestation subject SHOULD include all the artifacts associated with the release at that time; otherwise it will be unclear if an artifact was later removed from a release. For example, if a release version has an initial release attestation with artifact A, and then later has a release attestation with only artifact B, that SHOULD be interpreted as the release version now only containing artifact B.
```
marcelamelara, line 35: Not a necessary requirement, but I wonder how SBOM may fit here. Would a reference to an SBOM enumerating all artifacts in the release be a valid subject?

spec/predicates/release.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Schema
+
+### Fields
+
+- **`subject.name`** string
+  - The filename of the artifact as it would appear on disk.
+
+- **`predicate.purl`, required** string (ResourceURI)
```
marcelamelara, line 44: Can you please link to the ResourceURI field type spec for completeness?

marcelamelara (changes requested):
This first iteration looks quite good, thanks @steiza ! 

A couple high level comments. Can you please wrap lines at 80 chars? Also, per #320, once you're ready, can you please also add a proto definition and add this predicate to the in-toto.io namespace?

spec/predicates/release.md
marcelamelara: Per the field description of `subject.name`, the answer is probably no. But this may be an interesting consideration in the future.

spec/predicates/release.md
```diff
@@ -0,0 +1,110 @@
...
+
+This predicate is for the final stages of the software supply chain, where consumers are looking for attestations that the software has not been tampered with during distribution.
+
+If a previously attested release is later determined to be untrustworthy, a registry MAY stop serving the artifacts associated with that release.
+
+If a registry supports immutable releases, there SHOULD be one release attestation for a given `predicate.purl`.
+
+Even if the registry does not support immutable releases, the attestation subject SHOULD include all the artifacts associated with the release at that time; otherwise it will be unclear if an artifact was later removed from a release. For example, if a release version has an initial release attestation with artifact A, and then later has a release attestation with only artifact B, that SHOULD be interpreted as the release version now only containing artifact B.
+
+## Schema
```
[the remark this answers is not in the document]
steiza: I have added the full schema

spec/predicates/release.md
```diff
@@ -0,0 +1,110 @@
...
+
+If these release attestations are optionally published to a transparency log, package authors (or other interested parties) can monitor when a new version of a package is released.
+
+The release attestation provides integrity for the release artifacts, but it does not provide availability as it does not require the registry to serve an attested release artifact.
+
+These use cases are not hypothetical; both of them are the case today for npm's [build provenance feature], which includes a [publish attestation]. Note that while npm calls this a publish attestation, calling it a release attestation better reflects that it's coming from the package registry. Publishing often refers to an author sending content to the registry, as in [PyPI's trusted publishers feature].
+
+## Prerequisites
+
+This predicate depends on the [in-toto attestation framework], as well as the [purl-spec] for identifying packaging ecosystems.
```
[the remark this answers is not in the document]
steiza: Aha, thanks!

spec/predicates/release.md
```diff
@@ -0,0 +1,110 @@
...
+
+Perhaps surprisingly, this predicate does not depend on [SLSA Provenance], but they are better together. In an ecosystem where some (but not all) packages have SLSA Provenance and you query by artifact hash, you can't tell the difference between an artifact that has been tampered with and one that does not yet have SLSA provenance. If the ecosystem has release attestations, you can give authoritative answers to what artifacts make up a given release version, and what hashes those artifacts should have.
```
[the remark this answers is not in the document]
steiza: Sounds good - I have moved it up to "Use Cases".

spec/predicates/release.md
steiza: Yes, since we added the language above in the Use Case section I'll just strike this sentence.

spec/predicates/release.md
steiza: Done!

spec/predicates/release.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Schema
+
+### Fields
+
+- **`subject.name`** string
+  - The filename of the artifact as it would appear on disk.
+
+- **`predicate.purl`, required** string (ResourceURI)
+  - A purl uniquely identifying a specific release name and version from a package registry.
```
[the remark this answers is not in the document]
steiza: I haven't made any changes here yet, but I'm still open to it.

I'm not sure what we mean, exactly, by not resolvable. Do we mean that purls aren't understood by tools like `curl` or a web browser? Or do we mean that just because you have a purl the registry might not serve you that content (if the registry is down, if it determines the content is malicious, or for any other reason)? Or is there another meaning I'm missing?

steiza:
> Also, per https://github.com/in-toto/attestation/pull/320, once you're ready, can you please also add a proto definition

I've added the proto definition (but not the generated code) to this pull request - let me know if you'd like the generated code as well.

> and add this predicate to the in-toto.io namespace?

I'm not sure I understand this process, but I took a stab at this with https://github.com/in-toto/in-toto.io/pull/23

spec/predicates/release.md
marcelamelara: interesting point! My initial reaction is to suggest using the `annotations` field for this sort of extra metadata, because file size is a resource-specific piece of info. But I can also see the benefits of adding an optional `size` field to the object. The parsing rules for this field may change depending on the type of resource being described, so that may get a little tricky. 

Do feel free to open an issue on this, because yes, changing the RD spec should be handled separately from this PR.

spec/predicates/release.md
marcelamelara: Yes! And keeping the SBOM separate from the release attestation, I think, also makes verification of each a little more straightforward.

spec/predicates/release.md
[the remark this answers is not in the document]
marcelamelara: Good point. I was hoping to find more clarity in the spec, but it only states that the "URI SHOULD resolve to the artifact". So I take that to mean that "unresolvable" does cover both a purl that isn't understood by a web tool and/or may point to a non-existent entry in a registry. We really ought to update the spec to clarify this, though, since you're right to point out that "unresolvable" is imprecise.

marcelamelara:
> I've added the proto definition (but not the generated code) to this pull request - let me know if you'd like the generated code as well.

We have a workflow for generating the language bindings from the protos, so you're all set!

> I'm not sure I understand this process, but I took a stab at this with https://github.com/in-toto/in-toto.io/pull/23

This is a new part of the process for us, so we're still finalizing some of the details, but the PR you opened in that repo is definitely in line with how we want this to work. Thanks!!

spec/predicates/release.md
trishankatdatadog: It's also the sort of stuff that TUF was designed for, so maybe it's not necessary here, but it's something worth thinking about upstream with the RD spec.

spec/predicates/release.md
```diff
@@ -2,54 +2,105 @@
 
 Type URI: https://in-toto.io/attestation/release
 
-Version 0.1
+Version 1
```
marcelamelara, line 5: Any particular reason for the change from v0.1 to v1?

marcelamelara (commented):
Thanks for the updates @steiza ! I think we're almost there. Can you please also fix the linter errors?

spec/predicates/release.md
steiza: At first I didn't understand how to specify a protobuf for v0.1, but I think I figured it out, so I put the version back at 0.1 here and in https://github.com/in-toto/in-toto.io/pull/23.

spec/predicates/release.md
```diff
@@ -156,8 +156,55 @@ attestation SHOULD have a subject per artifact:
 ## Changelog and Migrations
 
 As this is the initial version, no changes or migrations to previous versions.
-This proposal is a subset of the information in the existing npm
-[publish attestation], so npm could easily migrate to this specification.
+The required predicates in this specification are a subset of the information
```
marcelamelara, line 159: typo?
```suggestion
The required fields in this specification are a subset of the information
```

spec/predicates/release.md
```diff
@@ -156,8 +156,55 @@ attestation SHOULD have a subject per artifact:
...
+in the existing npm [publish attestation], so npm could easily migrate to this
+specification.
+
+For example, here is an existing npm publish attestation:
+
+```json
+{
+  "_type": "https://in-toto.io/Statement/v0.1",
```
marcelamelara, line 167: This should probably be Statement/v1. Same below.

marcelamelara (commented):
Thanks again!

spec/predicates/release.md
steiza: Oops! I got carried away setting things to v0.1 - I have now fixed this.

spec/predicates/release.md
steiza: Based on the discussion we are going to leave out account identity.

spec/predicates/release.md
steiza: We have determined this is out of scope for this ITE-9 process.

spec/predicates/release.md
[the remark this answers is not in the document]
steiza: After we update the spec, we can evaluate if we need to clarify here.

trishankatdatadog (approved):
Let's go 🚀

spec/predicates/release.md
steiza: Yeah, I'm not sure we'll be able to come up with a one-size-fits-all term here, but based on current input I think it makes sense to stick with "release attestation".

spec/predicates/release.md
```diff
@@ -0,0 +1,217 @@
...
+versions (e.g. it's associated with urllib3, not urllib3 v2.1.0). This will
+allow users to confirm that a release has moved to a new name, and prevent
+confusion if the old name is re-used. This could be an automatically
+incrementing database key or a randomly generated UUID.
+
+### Parsing Rules
+
+The purl field MUST be parsed using the [purl-spec]. It MUST include a purl
+`version` (which is OPTIONAL in the [purl-spec]). It SHOULD NOT include purl
+`qualifiers` or `subpath`, unless the `type` requires them to uniquely identify
```
pxp928, line 105: Question: `qualifiers` would be needed for different architectures?

spec/predicates/release.md
```diff
@@ -0,0 +1,217 @@
...
+A purl uniquely identifying a specific release name and version from a package
+registry.
+
+**`predicate.releaseId`** string
+
+Stable identifier for a release; this should remain unchanged between release
+versions (e.g. it's associated with urllib3, not urllib3 v2.1.0). This will
+allow users to confirm that a release has moved to a new name, and prevent
+confusion if the old name is re-used. This could be an automatically
+incrementing database key or a randomly generated UUID.
```
pxp928, line 99: the UUID would have to be generated by the releaser. In that case, we would need to keep track of how the UUIDs map to each artifact. How would we know if the `releaseID` does not match?

spec/predicates/release.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Parsing Rules
+
+The purl field MUST be parsed using the [purl-spec]. It MUST include a purl
+`version` (which is OPTIONAL in the [purl-spec]). It SHOULD NOT include purl
+`qualifiers` or `subpath`, unless the `type` requires them to uniquely identify
+a release (as a counter-example `type:oci` would include `qualifiers`).
```
pxp928, line 106: Does this also include images? or just artifacts? If so, can we add an example of this for reference?

pxp928 (commented):
couple of comments but looks good overall!

spec/predicates/release.md
```diff
@@ -153,6 +154,26 @@ attestation SHOULD have a subject per artifact:
 }
 ```
 
+Here's what a release with a container images looks like:
```
marcelamelara, line 157: Tiny typo.

```suggestion
Here's what a release with a container image looks like:
```

marcelamelara (approved):
Thanks again!

pxp928 (approved):
Thanks @steiza!

PR #320, opened 2024-01-22
Add common action items for the new predicate vetting process

### LINKED ISSUE #313: Should in-toto managed vetted predicates type URIs point to GitHub repo?
Many predicateType URIs are non-existing. The TypeURI spec reads "SHOULD resolve to a human-readable description, but MAY be unresolvable.", so it is allowed to not to be there. But the non-existence of a predicateType URI seems to be causing confusion for some implementers when deciding when to pick a predicateType for generated attestations.

A possible solution may be to use the GitHub repo location for vetted predicates that are managed by in-toto as the predicateType URI. This would require an ITE-9 change, but pointing to the GH repo still signals that a predicate spec is managed by in-toto _and_ is resolvable.

_Originally posted by @fred-jonkhart in https://github.com/in-toto/attestation/issues/256#issuecomment-1829597525_

CC @adityasaky

marcelamelara:
This PR explicitly adds three main common action items that are part of the vetting process for new predicate types:
- updating the list of existing predicates
- adding a proto definition to generate language bindings
- updating the URL redirects for predicateType URIs that are defined within the in-toto.io/attestation namespace

fixes #313

### DIFF
--- a/docs/new_predicate_guidelines.md
+++ b/docs/new_predicate_guidelines.md
@@ -55,12 +55,14 @@ Our vetting process is simple.
     -   Add the new predicate to the list of [existing predicates].
     -   To generate Go/Python/Java language bindings for the new predicate,
         include a [protobuf definition].
-    -   If the [predicateType] URI will be defined under the
-        https://in-toto.io/attestation namespace, submit a PR to add a line
-        to the [URL redirects list] for the in-toto.io domain.
 2.  The in-toto Attestation Framework maintainers will review the PR at the
     next maintainers meeting.
 3.  If accepted, the new predicate type will be included in our directory.
+4.  Finally, if the new [predicateType] URI is defined under the
+    https://in-toto.io/attestation namespace, submit a PR to add the following
+    line to the [URL redirects list] for the in-toto.io domain:
+
+`attestation/<predicate name>/<version> https://github.com/in-toto/attestation/tree/main/spec/predicates/<predicate name>.md`
 
 [ITE-9]: https://github.com/in-toto/ITE/tree/master/ITE/9#document-format
 [Predicates]: ../spec/v1/predicate.md
--- a/docs/new_predicate_guidelines.md
+++ b/docs/new_predicate_guidelines.md
@@ -52,6 +52,12 @@ specification for you.
 Our vetting process is simple.
 
 1.  Open a PR following [ITE-9] formatting guidelines.
+    -   Add a the new predicate to the list of [existing predicates].
+    -   To generate Go/Python/Java language bindings for the new predicate,
+        include a [protobuf definition].
+    -   If the [predicateType] URI will be defined under the
+        https://in-toto.io/attestation namespace, submit a PR to add a line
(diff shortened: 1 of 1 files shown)


docs/new_predicate_guidelines.md
```diff
@@ -52,14 +52,23 @@ specification for you.
...
     next maintainers meeting.
 3.  If accepted, the new predicate type will be included in our directory.
 
 [ITE-9]: https://github.com/in-toto/ITE/tree/master/ITE/9#document-format
 [Predicates]: ../spec/v1/predicate.md
 [RFC 3339]: https://tools.ietf.org/html/rfc3339
+[URL redirects list]: https://github.com/in-toto/in-toto.io/blob/master/public/_redirects
 [field types]: ../spec/v1/field_types.md
-[existing predicates]: ../spec/predicates
+[existing predicates]: ../spec/predicates/README/md
```
pxp928, line 70: typo -> `../spec/predicates/README.md`

pxp928 (approved):
LGTM

docs/new_predicate_guidelines.md
marcelamelara: thanks for catching this. fixed.

adityasaky:
Possibly out of scope for this PR but can this guide be updated to include inline ITE-9 recommendations so we don't turn ITE-9 into a living doc?

adityasaky:
> 1. Open a PR following [ITE-9] formatting guidelines.

Probably just inlining this so the ITE-9 reference is just for where it came from. Any future formatting updates then don't need to happen in ITE-9.

TomHennen (approved):

marcelamelara:
> Probably just inlining this so the ITE-9 reference is just for where it came from. Any future formatting updates then don't need to happen in ITE-9.

@adityasaky we figured that ITE-9 _is_ the authoritative source for predicate spec formatting, are you suggesting that any future format updates should be specified here instead?

adityasaky:
>@adityasaky we figured that ITE-9 is the authoritative source for predicate spec formatting, are you suggesting that any future format updates should be specified here instead?

yes. That'd enable ITE-9 to be accepted and then move into "final" status, for example (https://github.com/in-toto/ITE/blob/master/ITE/1/README.adoc#ite-workflow). My understanding of the ITE process is that they're not intended to be living documents, even if changes are expected to be far and few in between.

marcelamelara:
> My understanding of the ITE process is that they're not intended to be living documents, even if changes are expected to be far and few in between.
>
> So re-specifying the format in the attestation docs + stating they originated from ITE-9 is probably fine?

Thanks for the clarification. Now this makes sense. My take is to handle this in a dedicated PR, though.

marcelamelara:
@pxp928 @TomHennen Given the discussion with @adityasaky , are you with merging this PR as-is now?

adityasaky:
>Thanks for the clarification. Now this makes sense. My take is to handle this in a dedicated PR, though.

Sounds good to me!

PR #328, opened 2024-02-23
Add ITE-9 predicate template

### LINKED ISSUE #326: Incorporate ITE-9 predicate formatting into main spec
To enable ITE-9 to be accepted and then move into ["final" status](https://github.com/in-toto/ITE/blob/master/ITE/1/README.adoc#ite-workflow), the main attestation spec should become the authoritative source for predicate spec formatting. The reasoning for this is that ITEs are not intended to be living documents, even if changes are expected to be far and few in between.

So we should be re-specifying the format in the attestation spec + stating they originated from ITE-9 instread.

_Originally posted by @adityasaky in https://github.com/in-toto/attestation/issues/320#issuecomment-1928104090_

adityasaky:
Closes #326.

I've amended some of the original text to say predicate where ITE-9 says attestation as the full model is fleshed out here already.

### DIFF
--- a/docs/new_predicate_guidelines.md
+++ b/docs/new_predicate_guidelines.md
@@ -49,9 +49,10 @@ specification for you.
 
 ## Vetting process
 
-Our vetting process is simple.
+Our vetting process is simple, and is based on the process specified in [ITE-9].
 
-1.  Open a PR following [ITE-9] formatting guidelines.
+1.  Open a PR following the formatting guidelines specified in the
+    [predicate template].
     -   Add the new predicate to the list of [existing predicates].
     -   To generate Go/Python/Java language bindings for the new predicate,
         include a [protobuf definition].
@@ -74,3 +75,4 @@ Our vetting process is simple.
 [predicateType]: ../spec/v1/predicate.md#fields
 [protobuf definition]: ../protos/README.md
 [vetting process]: #vetting-process
+[predicate template]: /spec/predicates/template/template.md
--- a/spec/predicates/template/template.md
+++ b/spec/predicates/template/template.md
@@ -21,8 +21,8 @@ this predicate and how the current predicate types (if any) do not solve this.
 
 New predicate types may be very specific to a particular technology or
 specification, which must be detailed in the prerequisites section. The in-toto
-Attestation Framework is likely a necessary prerequisite for all predicate
-types, but it SHOULD be explicitly mentioned to err on the side of caution.
+Attestation Framework may be a necessary prerequisite for all predicate types,
+if so it SHOULD be explicitly mentioned to err on the side of caution.
 
 ## Model
 
--- a/spec/predicates/template/template.md
+++ b/spec/predicates/template/template.md
@@ -1,0 +1,87 @@
+# Predicate type: <name>
+
+Type URI:
+
+Version:
+
+Predicate Name:
+
+## Purpose
+
+This section must discuss at a high level the reason for creating the new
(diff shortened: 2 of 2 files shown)


spec/predicates/template/template.md
```diff
@@ -0,0 +1,87 @@
...
+## Use Cases
+
+This section must discuss at least one concrete use case that will be covered by
+this predicate and how the current predicate types (if any) do not solve this.
+
+## Prerequisites
+
+New predicate types may be very specific to a particular technology or
+specification, which must be detailed in the prerequisites section. The in-toto
+Attestation Framework is likely a necessary prerequisite for all predicate
```
TomHennen, line 24: I'm not sure if this is true anymore.

optional: "The in-toto Attestation Framework may be a necessary prerequisite... if so it SHOULD be explicitly..."

If you don't want to mess with it though that's fine.

docs/new_predicate_guidelines.md
```diff
@@ -74,3 +75,4 @@ Our vetting process is simple.
 [predicateType]: ../spec/v1/predicate.md#fields
 [protobuf definition]: ../protos/README.md
 [vetting process]: #vetting-process
+[predicate template]: /spec/predicates/template/template.md
```
marcelamelara, line 78: Nit: Do we expect other templates in the future? I'm curious about the choice to place the template file in a sub-directory.

marcelamelara (approved):
Thanks so much @adityasaky !

docs/new_predicate_guidelines.md
adityasaky: I don't expect more templates, I just wanted to separate it from the other predicate definitions. I also wonder if we ever foresee a template predicate but I suspect not. Should I move it up a directory?

marcelamelara: Yes, in fact, since it's in the same directory as the predicates, it might be a bit confusing as currently stored. In fact, I'm comfortable having this become `/spec/v1/predicate_template.md` since this may change in future versions. Wdyt?

adityasaky: I'm not sure if this belongs in the v1 directory. The new predicate guidelines aren't part of the spec, they're in docs/ and predicates are also independently versioned. I could just rename this as _TEMPLATE.md or something and put it in the same directory as all the other predicate specs.

marcelamelara: I tend to think of the predicate template as part of the actual spec because all predicates defined under v1 follow the same format, regardless of their version, whereas the new predicate guidelines are in docs/ because those are more fluid guidelines. In other words, I might expect the predicate format to evolve between different versions of the spec, while the guidelines shouldn't have an impact on the format of predicates. That said, I don't want to block this PR on this particular point either; we can always move files around the repo later, if need be.

docs/new_predicate_guidelines.md
```diff
@@ -49,9 +49,10 @@ specification for you.
 
 ## Vetting process
 
-Our vetting process is simple.
+Our vetting process is simple, and is based on the process specified in [ITE-9].
 
-1.  Open a PR following [ITE-9] formatting guidelines.
+1.  Open a PR following the formatting guidelines specified in the
```
pxp928, line 54: also linking to the existing predicates that people can look at as examples would be helpful for first time users.

pxp928 (approved):
LGTM. Quick suggestion

TomHennen (approved):

PR #330, opened 2024-03-01
Add DigestSet hex encoding validation

marcelamelara:
Since the spec requires digests in a DigestSet to be hex-encoded, this PR enhances the `Validate()` function for the ResourceDescriptor and adds a corresponding test.

### DIFF
--- a/go/v1/resource_descriptor.go
+++ b/go/v1/resource_descriptor.go
@@ -35,11 +35,12 @@ func (d *ResourceDescriptor) Validate() error {
 	if len(d.GetDigest()) > 0 {
 		for alg, digest := range d.GetDigest() {
 
-			// check encoding and length for supported algorithms
+			// Per https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
+			// check encoding and length for supported algorithms;
+			// use of custom, unsupported algorithms is allowed and does not not generate validation errors.
 			supported, size := isSupportedAlgorithm(alg)
 			if supported {
 				// the in-toto spec expects a hex-encoded string in DigestSets for supported algorithms
-				// https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
 				hashBytes, err := hex.DecodeString(digest)
 
 				if err != nil {
--- a/go/v1/resource_descriptor.go
+++ b/go/v1/resource_descriptor.go
@@ -5,14 +5,26 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
 package v1
 
 import (
+	"crypto/md5"
+	"crypto/sha1"
+	"crypto/sha512"
 	"encoding/hex"
 	"errors"
 )
 
 var (
+	ErrIncorrectDigestLength = errors.New("digest is not correct length")
 	ErrInvalidDigestEncoding = errors.New("digest is not valid hex-encoded string")
 	ErrRDRequiredField       = errors.New("at least one of name, URI, or digest are required")
 )
+
+// Supported standard hash algorithms
+func isSupportedAlgorithm(alg string) (bool, int) {
+	algos := map[string]int{"md5": md5.Size, "sha1": sha1.Size, "shake128": md5.Size, "sha224": sha512.Size224, "sha3_224": sha512.Size224, "sha512_224": sha512.Size224, "sha256": sha512.Size256, "sha3_256": sha512.Size256, "sha512_256": sha512.Size256, "shake256": sha512.Size256, "sha384": sha512.Size384, "sha3_384": sha512.Size384, "sha512_384": sha512.Size384, "sha512": sha512.Size, "sha3_512": sha512.Size, "dirHash": sha512.Size256, "gitCommit": sha1.Size}
+
+	size, ok := algos[alg]
+	return ok, size
+}
 
 func (d *ResourceDescriptor) Validate() error {
 	// at least one of name, URI or digest are required
@@ -20,14 +32,24 @@ func (d *ResourceDescriptor) Validate() error {
 		return ErrRDRequiredField
 	}
 
-	// the in-toto spec expects a hex-encoded string in DigestSets
-	// https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
 	if len(d.GetDigest()) > 0 {
-		for _, digest := range d.GetDigest() {
-			_, err := hex.DecodeString(digest)
+		for alg, digest := range d.GetDigest() {
 
-			if err != nil {
-				return ErrInvalidDigestEncoding
+			// check encoding and length for supported algorithms
+			supported, size := isSupportedAlgorithm(alg)
+			if supported {
+				// the in-toto spec expects a hex-encoded string in DigestSets for supported algorithms
+				// https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
+				hashBytes, err := hex.DecodeString(digest)
+
+				if err != nil {
+					return ErrInvalidDigestEncoding
+				}
+
+				// check the length of the digest
+				if len(hashBytes) != size {
+					return ErrIncorrectDigestLength
+				}
 			}
 		}
 	}
--- a/go/v1/resource_descriptor.go
+++ b/go/v1/resource_descriptor.go
@@ -5,22 +5,24 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
 package v1
 
 import (
-	"crypto/md5"
-	"crypto/sha1"
-	"crypto/sha512"
 	"encoding/hex"
 	"errors"
+	"fmt"
 )
 
 var (
-	ErrIncorrectDigestLength = errors.New("digest is not correct length")
+	ErrIncorrectDigestLength = errors.New("digest has incorrect length")
 	ErrInvalidDigestEncoding = errors.New("digest is not valid hex-encoded string")
 	ErrRDRequiredField       = errors.New("at least one of name, URI, or digest are required")
 )
 
-// Supported standard hash algorithms
-func isSupportedAlgorithm(alg string) (bool, int) {
-	algos := map[string]int{"md5": md5.Size, "sha1": sha1.Size, "shake128": md5.Size, "sha224": sha512.Size224, "sha3_224": sha512.Size224, "sha512_224": sha512.Size224, "sha256": sha512.Size256, "sha3_256": sha512.Size256, "sha512_256": sha512.Size256, "shake256": sha512.Size256, "sha384": sha512.Size384, "sha3_384": sha512.Size384, "sha512_384": sha512.Size384, "sha512": sha512.Size, "sha3_512": sha512.Size, "dirHash": sha512.Size256, "gitCommit": sha1.Size}
+// Indicates if a given fixed-size hash algorithm is supported by default and returns the algorithm's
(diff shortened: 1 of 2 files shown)


go/v1/resource_descriptor.go
```diff
@@ -4,15 +4,33 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
...
 	// at least one of name, URI or digest are required
 	if d.GetName() == "" && d.GetUri() == "" && len(d.GetDigest()) == 0 {
 		return ErrRDRequiredField
 	}
 
+	// the in-toto spec expects a hex-encoded string in DigestSets
+	// https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
+	if len(d.GetDigest()) > 0 {
+		for _, digest := range d.GetDigest() {
+			_, err := hex.DecodeString(digest)
```
adityasaky, line 27: This could also potentially do other checks like length checks for a known hash algorithm :thinking:

marcelamelara: We probably should, yes.

go/v1/resource_descriptor.go
```diff
@@ -4,15 +4,33 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
...
+	ErrRDRequiredField       = errors.New("at least one of name, URI, or digest are required")
+)
 
 func (d *ResourceDescriptor) Validate() error {
 	// at least one of name, URI or digest are required
 	if d.GetName() == "" && d.GetUri() == "" && len(d.GetDigest()) == 0 {
 		return ErrRDRequiredField
 	}
 
+	// the in-toto spec expects a hex-encoded string in DigestSets
```
TomHennen, line 23: Note that this is only expected for the named standard algorithms listed on the page: https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#:~:text=The%20named%20standard%20algorithms.

For other algorithm types there's no restriction.

FWIW we are using some other types and their values wind up being non-hex-encoded.

marcelamelara: ah, interesting. So even this check should really be made conditional on the algorithm, making Aditya's suggestion to check length an easier addition, still.

adityasaky: > For other algorithm types there's no restriction.

Is this by design or is it more of a loophole? :thinking:

TomHennen: By design!  For example, we have some immutable IDs that start with `1-`.  We might also have some IDs which are immutable based on underlying implementation that are just digits.

We even tell folks "New algorithms MUST document how the value is encoded, e.g. URL-safe base64, lowercase hex, etc..." which indicates that new algorithms have the choice.

go/v1/resource_descriptor.go
```diff
@@ -5,29 +5,51 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
...
 
 var (
+	ErrIncorrectDigestLength = errors.New("digest is not correct length")
 	ErrInvalidDigestEncoding = errors.New("digest is not valid hex-encoded string")
 	ErrRDRequiredField       = errors.New("at least one of name, URI, or digest are required")
 )
 
+// Supported standard hash algorithms
+func isSupportedAlgorithm(alg string) (bool, int) {
+	algos := map[string]int{"md5": md5.Size, "sha1": sha1.Size, "shake128": md5.Size, "sha224": sha512.Size224, "sha3_224": sha512.Size224, "sha512_224": sha512.Size224, "sha256": sha512.Size256, "sha3_256": sha512.Size256, "sha512_256": sha512.Size256, "shake256": sha512.Size256, "sha384": sha512.Size384, "sha3_384": sha512.Size384, "sha512_384": sha512.Size384, "sha512": sha512.Size, "sha3_512": sha512.Size, "dirHash": sha512.Size256, "gitCommit": sha1.Size}
```
TomHennen, line 23: should 'shake128' be using md5.Size?  (I don't know much about that hash algorithm).

I'd probably want someone else to take a look at the rest of these too.

BTW can you say a bit about how you came up with these values?

go/v1/resource_descriptor.go
```diff
@@ -5,29 +5,51 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
...
 import (
+	"crypto/md5"
+	"crypto/sha1"
+	"crypto/sha512"
 	"encoding/hex"
 	"errors"
 )
 
 var (
+	ErrIncorrectDigestLength = errors.New("digest is not correct length")
```
TomHennen, line 16: Since attestations can have lots of digests in them would it make sense to put more metadata in the errors that get returned?  Ideally we could return the digest type and value?  That way folks could more easily find the part of the attestation that's generating the error?

(This probably applies to both `ErrIncorrectDigestLength` and `ErrInvalidDigestEncoding`)

go/v1/resource_descriptor.go
```diff
@@ -5,29 +5,51 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
...
 
-	// the in-toto spec expects a hex-encoded string in DigestSets
-	// https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
 	if len(d.GetDigest()) > 0 {
-		for _, digest := range d.GetDigest() {
-			_, err := hex.DecodeString(digest)
+		for alg, digest := range d.GetDigest() {
+
+			// check encoding and length for supported algorithms
+			supported, size := isSupportedAlgorithm(alg)
```
TomHennen, line 39: Can you add a comment like "Users are allowed to use unsupported algorithms and their use should not generate a validation error."

go/v1/resource_descriptor_test.go
```diff
@@ -68,12 +70,22 @@ func TestBadResourceDescriptor(t *testing.T) {
 	assert.ErrorIs(t, err, ErrRDRequiredField, "created malformed ResourceDescriptor")
 }
 
-func TestBadResourceDescriptorDigest(t *testing.T) {
+func TestBadResourceDescriptorDigestEncoding(t *testing.T) {
```
TomHennen, line 73: Perhaps I missed them, but are there any tests for 'good' digest encodings?

TomHennen (changes requested):

go/v1/resource_descriptor.go
marcelamelara: I specifically looked up shake128 because I'm also not familiar, and found that it's 16 bytes (size of MD5). But yes, in general, these are based on the hash size constants specified in the respective Go crypto packages. I didn't want to have to import 6 different packages just for the size constants, which is why I use the sha512 constants everywhere possible.

go/v1/resource_descriptor.go
marcelamelara: We could definitely do this for the length error. `ErrInvalidDigestEncoding` is currently only returned if any of the standard supported algorithms isn't hex encoded, so not sure what other metadata we could add here other than have a more detailed error message?

go/v1/resource_descriptor_test.go
marcelamelara: Ah, good catch. We should add an explicit test for a valid supported digest (encoding and length).

go/v1/resource_descriptor.go
marcelamelara: Oh, I just realized, you probably mean to add specific algorithm information to `ErrInvalidDigestEncoding`. Yes, that makes sense too!

go/v1/resource_descriptor.go
TomHennen: Gotcha.  Can you add a comment to that effect?  (Or alternatively use literals)

marcelamelara: If literals are a bit clearer, I can switch to that.

go/v1/resource_descriptor.go
```diff
@@ -4,15 +4,56 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
...
+)
 
-var ErrRDRequiredField = errors.New("at least one of name, URI, or digest are required")
+var (
+	ErrIncorrectDigestLength = errors.New("digest is not correct length")
+	ErrInvalidDigestEncoding = errors.New("digest is not valid hex-encoded string")
+	ErrRDRequiredField       = errors.New("at least one of name, URI, or digest are required")
+)
+
+// Supported standard hash algorithms
```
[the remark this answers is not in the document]
marcelamelara: To clarify, those are the raw unencoded sizes in bytes. Will specify that in the comment.

go/v1/resource_descriptor.go
marcelamelara: I ended up leaving these errors intact and instead parameterized the error messages in the `Validate()` function, as parametrizing these errors broke error checking in the tests.

go/v1/resource_descriptor.go
marcelamelara: done.

go/v1/resource_descriptor.go
```diff
@@ -4,15 +4,58 @@ Wrapper APIs for in-toto attestation ResourceDescriptor protos.
...
+			// Per https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
+			// check encoding and length for supported algorithms;
+			// use of custom, unsupported algorithms is allowed and does not not generate validation errors.
+			supported, size := isSupportedFixedSizeAlgorithm(alg)
+			if supported {
+				// the in-toto spec expects a hex-encoded string in DigestSets for supported algorithms
+				hashBytes, err := hex.DecodeString(digest)
+
+				if err != nil {
+					return fmt.Errorf("%w: %s", ErrInvalidDigestEncoding, alg)
```
TomHennen, line 49: optional: I suppose I was thinking it would be most helpful to include `digest` in the error message.

I'm imagining a statement could have thousands of ResourceDescriptors with sha256 and maybe only one of them is missing a single character or doesn't have the right encoding.

go/v1/resource_descriptor_test.go
```diff
@@ -65,3 +81,23 @@ func TestBadResourceDescriptor(t *testing.T) {
...
 }
+
+func TestBadResourceDescriptorDigestEncoding(t *testing.T) {
+	got := &ResourceDescriptor{}
+	err := protojson.Unmarshal([]byte(badRdDigestEncoding), got)
+
+	assert.NoError(t, err, "Error during JSON unmarshalling")
+
+	err = got.Validate()
+	assert.ErrorIs(t, err, ErrInvalidDigestEncoding, "created ResourceDescriptor with invalid digest encoding")
```
TomHennen, line 92: nit/optional: should the message, if this check fails be something like "Did not get expected error when using invalid digest encoding"?

(Similar for below)

TomHennen (approved):

go/v1/resource_descriptor.go
marcelamelara: Good point. It's relatively straightforward to add, happy to do it.

go/v1/resource_descriptor_test.go
marcelamelara: done.

adityasaky (approved):

TomHennen (approved):

pxp928 (approved):
LGTM

kamealrabeai (approved):

PR #338, opened 2024-03-13
Clarify that digests don't have to be cryptographic ones.

TomHennen:
There are uses for DigestSets that don't involve cryptographic hashes and I don't think we ever meant to preclude those uses?

(This PR motivated by [this discussion](https://github.com/in-toto/attestation/pull/330/commits/e8c06cf62050b96daf4ef356bf382f4680a07be3#r1511872856) and another I've had with @laurentsimon).

Sometimes users have a need to refer to something by some other immutable identifier.  Either because the content can't be hashed traditionally, because it's impractical to hash traditionally, or because they interact with the content through an interface that doesn't expose them to the entirety of the content.

In these situations users may wish to use other identifiers in a DigestSet.  Those users should be careful to understand the trust that they're placing in the identifier to be sure that it meets their needs.

One concrete example of where a non-cryptographic hash can be useful is when referring to Virtual Machine images.  Often these images are very large (so impractical to run a cryptographic hash over) and users often interact with them via APIs that the platform provides that don't involve the user having complete custody of the content.  Platforms like AWS and GCP provide 'ids' for users to use when referring to these images.  A user may say something like "create an instance with image 123".  In that case the user doesn't actually have the bits that correspond to 'image 123' so they cannot digest it themselves.  And by the time the image has started it can be difficult, if not impossible, to digest the original content that was used to boot the instance (this is probably a good use case for secure boot, but is out of scope here :)).

These IDs can often be treated as immutable and may be perfectly suited to users threat profiles.  Allowing DigestSets to use these types of identifiers would allow providers to make statements about the content of these VM images using the identifiers their users have ready access to.

In addition, using an ID like this does not preclude including a cryptographic hash in the DigestSet _as well_. If possible including both may provide the most flexibility for the user's various use cases.

### DIFF
--- a/spec/v1/CHANGELOG.md
+++ b/spec/v1/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+
+## v1.1
+
+* Clarified that subjects are assumed to be immuatble and that it is
+acceptable to use a non-cryptographic digest (though cryptographic
+digests are still preferred).
+
+## v1
+
+Initial release.
--- a/spec/v1/CHANGELOG.md
+++ b/spec/v1/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v1.1
 
-* Clarified that subjects are assumed to be immuatble and that it is
+-   Clarified that subjects are assumed to be immuatble and that it is
 acceptable to use a non-cryptographic digest (though cryptographic
 digests are still preferred).
 
--- a/spec/v1/CHANGELOG.md
+++ b/spec/v1/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## v1.1
 
--   Clarified that subjects are assumed to be immuatble and that it is
+-   Clarified that subjects are assumed to be immutable and that it is
 acceptable to use a non-cryptographic digest (though cryptographic
-digests are still preferred).
+digests are still strongly recommended).
 
 ## v1
 
--- a/spec/v1/digest_set.md
+++ b/spec/v1/digest_set.md
@@ -23,7 +23,7 @@ algorithms below use lowercase hex encoding. Usually there is just a
 single key/value pair, but multiple entries MAY be used for algorithm
 agility.
 
-Users SHOULD use a _cryptographic_ digest, but MAY use another identifier 
+Users SHOULD use a _cryptographic_ digest, but MAY use another identifier
 if the underlying implementation ensures immutability via other means.
 
 ### Supported algorithms
--- a/spec/v1/digest_set.md
+++ b/spec/v1/digest_set.md
@@ -2,8 +2,8 @@
 
 Version: v1.1
 
-Set of one or more immutable digests for a single software artifact or
-metadata object.
+Set of one or more immutable digests for a single software artifact or metadata
+object.
 
 ## Schema
 
@@ -11,20 +11,19 @@ metadata object.
 {
   "<ALGORITHM_1>": "<VALUE>",
   "<ALGORITHM_2>": "<VALUE>",
-  ... 
+  ...
 }
 ```
 
 ## Fields
 
-A DigestSet is represented as a _JSON object_ mapping algorithm name to
-a string encoding of the digest using that algorithm. The named standard
-algorithms below use lowercase hex encoding. Usually there is just a
-single key/value pair, but multiple entries MAY be used for algorithm
-agility.
+A DigestSet is represented as a *JSON object* mapping algorithm name to a string
+encoding of the digest using that algorithm. The named standard algorithms below
+use lowercase hex encoding. Usually there is just a single key/value pair, but
+multiple entries MAY be used for algorithm agility.
 
-Users SHOULD use a _cryptographic_ digest, but MAY use another identifier
-if the underlying implementation ensures immutability via other means.
+Users SHOULD use a *cryptographic* digest, but MAY use another identifier if the
+underlying implementation ensures immutability via other means.
 
 ### Supported algorithms
 
@@ -36,11 +35,11 @@ for cases when the method of serialization is obvious or well known.
 
 #### `dirHash`
 
-The [directory Hash1][] function, omitting the "h1:" prefix
-and output in lowercase hexadecimal instead of base64. This algorithm was
-designed for go modules but can be used to digest the _contents_ of an
-arbitrary archive or file tree. Equivalent to extracting the archive to an
-empty directory and running the following command in that directory:
+The [directory Hash1][] function, omitting the "h1:" prefix and output in
+lowercase hexadecimal instead of base64. This algorithm was designed for go
+modules but can be used to digest the *contents* of an arbitrary archive or file
+tree. Equivalent to extracting the archive to an empty directory and running the
+following command in that directory:
 
 ```bash
 find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut -f1 -d' '
@@ -48,13 +47,14 @@ find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum | sha256sum | cut
 
 For example, the module dirhash
 `h1:Khu2En+0gcYPZ2kuIihfswbzxv/mIHXgzPZ018Oty48=` would be encoded as
-`{"dirHash1": "2a1bb6127fb481c60f67692e22285fb306f3c6ffe62075e0ccf674d7c3adcb8f"}`.
+`{"dirHash1":
+"2a1bb6127fb481c60f67692e22285fb306f3c6ffe62075e0ccf674d7c3adcb8f"}`.
 
 <details>
 <summary>Detailed example</summary>
 
-The go module `github.com/marklodato/go-hello-world@v0.0.1` has module
-dirhash `h1:Khu2En+0gcYPZ2kuIihfswbzxv/mIHXgzPZ018Oty48=`:
+The go module `github.com/marklodato/go-hello-world@v0.0.1` has module dirhash
+`h1:Khu2En+0gcYPZ2kuIihfswbzxv/mIHXgzPZ018Oty48=`:
 
 ```bash
 $ curl https://sum.golang.org/lookup/github.com/marklodato/go-hello-world@v0.0.1
@@ -63,8 +63,8 @@ github.com/marklodato/go-hello-world v0.0.1 h1:Khu2En+0gcYPZ2kuIihfswbzxv/mIHXgz
 ...
 ```
 
-To compute the dirhash by hand, first fetch the module archive and extract
-it to an empty directory:
+To compute the dirhash by hand, first fetch the module archive and extract it to
+an empty directory:
 
 ```bash
 curl -O https://proxy.golang.org/github.com/marklodato/go-hello-world/@v/v0.0.1.zip
@@ -73,8 +73,8 @@ cd tmp
 unzip ../v0.0.1.zip
 ```
 
-We can see all of the files in the directory using the first part of the
-command above:
+We can see all of the files in the directory using the first part of the command
+above:
 
 ```bash
 $ find . -type f | cut -c3- | LC_ALL=C sort | xargs -r sha256sum
@@ -114,9 +114,9 @@ This hash is computed over `<type> SP <size> NUL <content>`, where:
 -   `NUL` is the ASCII NUL character, 0x00
 -   `<content>` is git representation of the object:
     -   For `commit`, the raw commit object ([more info][so-commit][^git-docs])
-    -   For `tree`,  the raw tree object, which is a series of
-        `<unix-octal-mode> <name> NUL <binary-digest>` entries, sorted by
-        `<name>` in the C locale ([more info][so-tree][^git-docs])
+    -   For `tree`, the raw tree object, which is a series of `<unix-octal-mode>
+        <name> NUL <binary-digest>` entries, sorted by `<name>` in the C locale
+        ([more info][so-tree][^git-docs])
     -   For `blob`, the raw file contents
     -   For `tag`, the raw tag object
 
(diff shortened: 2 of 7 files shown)


JustinCappos:
How do you enforce immutability in this case?  Or, to ask another way, why is immutability important if you're relaxing the constraint on cryptographically verification and trusting a third party?

TomHennen:
> How do you enforce immutability in this case? Or, to ask another way, why is immutability important if you're relaxing the constraint on cryptographically verification and trusting a third party?

Both good questions.

I'd expect the underlying implementation to ensure immutability to the extent required by users.  It could be something like "there's a monotonic system that assigns new IDs for each change (like subversion does)" to something more complicated.

> why is immutability important

Maybe it's not?  Maybe it would be enough to say "with custom types the type must document what the properties are (immutable, cryptographic digest, ...)"?  Mostly I'm trying to keep this change as small as needed but I don't think it's actually needed.

adityasaky:
I think some examples and discussion in-spec about the change in properties would be helpful. I'm reminded a little of [this TAP](https://github.com/theupdateframework/taps/blob/master/tap19.md) but I think the proposal here is looser still.

TomHennen:
> I think some examples and discussion in-spec about the change in properties would be helpful. I'm reminded a little of [this TAP](https://github.com/theupdateframework/taps/blob/master/tap19.md) but I think the proposal here is looser still.

How do you think that would be best handled here?  Inline in `spec/v1/digest_set.md` or in some new place?  (I don't think in-toto attestations has such a place yet?)

TomHennen:
Friendly ping?

SantiagoTorres:
My understanding is that this can be handled by ITE-4 and addressed by the type of "transport" that's being used. We've been leaving how to handle these to the specification, given that there were not a lot of "non file-hash" based identifiers. I assume this is something we could start standardizing/settling along on attestation or somewhere else.

All in all, each "hashable" would need its own security analysis. I understand that some cases use identifiers which, essentially, transfer trust to a third party identifier (e.g., using an SVN checkout id). I think we could taxonomize this, as this will probably fall within just a handful of cases.

SantiagoTorres:
I was wondering if that's in order, perhaps something that overrides ITE-4 with a better security notion about hashers/algorithms. 

Technically, a collission resistant hash can be replaced with something else (e.g., an artifact directory) under a different, more centralized threat model. In that sense, then the hash objects become just raw identifiers from a particular location.
It'd warrant a bit more exploration but I can see it opening doors in some cases

TomHennen:
> I was wondering if that's in order, perhaps something that overrides ITE-4 with a better security notion about hashers/algorithms.
> 
> Technically, a collission resistant hash can be replaced with something else (e.g., an artifact directory) under a different, more centralized threat model. In that sense, then the hash objects become just raw identifiers from a particular location. It'd warrant a bit more exploration but I can see it opening doors in some cases

Can you say what you're looking in this PR more specifically?

TomHennen:
> is it because you still expect this ID to have the immutability semantics of a cryptographic digest?

Yes I do.  That's definitely the case for GCE images and I suspect it's also the case for the Amazon ones.

FWIW the current PR does use 'immutable' and I'd be happy to keep it that way.

I'm realizing this was discussed in https://github.com/in-toto/attestation/issues/5.

I have a preference to use 'digest' for this purpose because:

1. The spec currently [requires digest to be set](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md#:~:text=Each%20element%20MUST%20have%20digest%20set.) (which could be changed)
2. The spec says things should matched [purely by digest](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md#:~:text=Subject%20artifacts%20are%20matched%20purely%20by%20digest) (which could be changed)
3. Infrastructure and tooling that is just looking for something that matches can no longer query just by digest type/value it must now conditionally support looking in other fields too.

I'm not really sure what the downsides would be apart from the name being somewhat confusing.  Users still need to provide the digests they're planning to match against so I don't think people would wind up accidentally using something?

I'm curious to get @MarkLodato's thoughts though as he had a lot of input on #5.

MarkLodato:
I'm OK abusing `digest` here. Basically we'd be saying that a `subject` is an immutable reference to a collection of artifacts, identified via `digest`. The strongly recommended, foolproof way to do this is via a cryptographic digest, which is why we call the field `digest`. However, if you know what you're doing, you can put some other non-cryptographic-digest identifier in that field, so long as it has the semantics of being an immutable reference and you accept the risks of the immutability breaking.

For example, GCE image IDs are effectively immutable unless something goes horribly wrong within GCE. That seems like an acceptable risk, and something I'd personally accept. You might not, so don't accept it, just like you might not accept SHA-1 or even SHA-256.

I don't like the idea of using another field because it makes the semantics so much more complicated, as @TomHennen said. Always matching by `digest` is straightforward. Let's keep it simple for the common case.

Regarding immutability, I think it's an inherent property of `subject`. Statement is an assertion that `subject` has property `predicate` without any sort of time bound, and that can only logically work if `subject` is immutable. (The predicate itself can have a time bound, but that is not a concept at the Statement layer.)

If we accept this, then I think we can close this issue and #5 (which we'd say it's always `digest` but you can abuse `digest` if you know what you're doing.)

marcelamelara:
@TomHennen @MarkLodato Thanks for your comments. I'm convinced that the `digest` field makes the most sense for any immutable identifier. 

Per Mark's suggestion, this PR should also update the Statement level spec to make the immutability assumption for the subject crystal clear, closing out #5.

TomHennen:
Thanks!

Updated statement, added rationale to the DigestSet docs per @adityasaky (but maybe he was suggesting something else?)

spec/v1/digest_set.md
MarkLodato, line 1: Could you revert the unrelated changes to formatting? Looks like you used a tool that formatted the entire file instead of just modified lines.

spec/v1/digest_set.md
```diff
@@ -1,27 +1,29 @@
 # DigestSet field type specification
 
-Version: v1.0
+Version: v1.1
 
-Set of one or more cryptographic digests for a single software artifact or
-metadata object.
+Set of one or more immutable digests for a single software artifact or metadata
+object.
```
MarkLodato, line 6: nit: a digest is always immutable since it's just a string. I think "immutable reference" is more accurate. Also I think it may be worth calling out that the overwhelming majority use case is cryptographic digest, to avoid confusion.

```suggestion
Set of one or more cryptographic digests, or other immutable references, for a
single software artifact or metadata object.
```

(Also ok to just say "immutable references" - your choice)

spec/v1/digest_set.md
```diff
@@ -1,27 +1,29 @@
...
-algorithms below use lowercase hex encoding. Usually there is just a
-single key/value pair, but multiple entries MAY be used for algorithm
-agility.
+A DigestSet is represented as a *JSON object* mapping algorithm name to a string
+encoding of the digest using that algorithm. The named standard algorithms below
+use lowercase hex encoding. Usually there is just a single key/value pair, but
+multiple entries MAY be used for algorithm agility.
+
+Users SHOULD use a *cryptographic* digest, but MAY use another identifier if the
+underlying implementation ensures immutability via other means.
```
MarkLodato, line 26: You might want to spell this out a bit more, since it's a bit nuanced. Maybe:

```suggestion
Each entry in a DigestSet MUST be an immutable reference to an artifact. It is
STRONGLY RECOMMENDED to use a commonly accepted, cryptographically secure digest
algorithm to achieve this immutability. See [Use cases for non-cryptographic,
immutable, digests](...) for further guidance.
```

spec/v1/digest_set.md
```diff
@@ -130,20 +133,51 @@ $ printf 'Hello' | git hash-object -t blob --stdin
...
+
+Sometimes users have a need to refer to something by some other immutable
+identifier. Either because the content can't be hashed traditionally, because
+it's impractical to hash traditionally, or because they interact with the
+content through an interface that doesn't expose them to the entirety of the
+content.
+
+In these situations users may wish to use other identifiers in a DigestSet.
+Those users should be careful to understand the trust that they're placing in
+the identifier to be sure that it meets their needs.
```
MarkLodato, line 159: Some editing suggestions to improve readability. Feel free to accept partially or to edit further.

```suggestion
While cryptographic digests are the strongly recommended immutable identifier,
users might have need to refer to an artifact by some other means. For example,
it might be technically infeasible to compute a digest over the content, or
because the user might interact with the content through an interface that
doesn't expose them to the entirety of the content.

In these situations, users MAY use a non-cryptographic identifier in a DigestSet so
long as the risk of the object being mutated is acceptable for the application.
```

MarkLodato (changes requested):
Thanks, Tom. I approve of the overall change. I just have some minor suggestions to improve clarity of the PR and content.

spec/v1/digest_set.md
TomHennen: Done. At some point I'd like to hear about your workflow here.  I'm just podging things together. :/

spec/v1/digest_set.md
TomHennen: Done

spec/v1/digest_set.md
TomHennen: Done

spec/v1/digest_set.md
TomHennen: Done.

TomHennen:
> Thanks, Tom. I approve of the overall change. I just have some minor suggestions to improve clarity of the PR and content.

Thanks for the comments and suggestions.  Done.

spec/v1/statement.md
```diff
@@ -38,6 +38,9 @@ Additional [parsing rules] apply.
 > Set of software artifacts that the attestation applies to. Each element
 > represents a single software artifact. Each element MUST have `digest` set.
 >
+> Subjects are assumed to me _immutable_, i.e. the artifacts identified by the
```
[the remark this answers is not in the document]
TomHennen: Fixed.

spec/v1/digest_set.md
```diff
@@ -1,16 +1,16 @@
 # DigestSet field type specification
 
-Version: v1.0
+Version: v1.1
```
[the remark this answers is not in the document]
TomHennen: Added

marcelamelara (commented):
LGTM @TomHennen . Before I approve, can you please double-check that we don't need to update the spec version to 1.1 anywhere else? I want to make sure we don't end up with version inconsistencies across the docs.

spec/v1/statement.md
```diff
@@ -67,6 +67,15 @@ Additional [parsing rules] apply.
 > set-but-empty. MAY be omitted if `predicateType` fully describes the
 > predicate.
 
+## Change History
+
+### v1.1
+Clarified that subjects are assumed to be immutable.
```
MarkLodato, line 74:
```suggestion

Clarified that subjects are assumed to be immutable.
```

MarkLodato (approved):

TomHennen:
> LGTM @TomHennen . Before I approve, can you please double-check that we don't need to update the spec version to 1.1 anywhere else? I want to make sure we don't end up with version inconsistencies across the spec docs.

Ah, so do we want to rev everything under spec/v1 to 1.1?  It's actually not clear to me if the version at the top is for _that document_ or for the attestation spec.  I could see it going either way.

If it is for the whole spec perhaps I should remove the 'change logs' from each doc and have a centralized changelog.md to document those things?

spec/v1/statement.md
TomHennen: I'm a bit confused by this suggestion.  In the most recent commit there is a blank line after `### v1.1` and before this sentence.

I don't think you're suggesting it needs to blanks.  So perhaps this was on an old commit?  Maybe the suggestion is something else?

MarkLodato: There's no blank line in this file (statement.md) while there is one in digest_set.md.

MarkLodato: Oh, weird, in my screen that wasn't present. Nevermind.

TomHennen: Looks like GitHub was serving you an old commit.   An extra line was added in https://github.com/in-toto/attestation/pull/338/commits/efef96f0b83b682f34b91bb7ab1e08fa86a82cc9

TomHennen:
> > Ah, so do we want to rev everything under spec/v1 to 1.1? It's actually not clear to me if the version at the top is for that document or for the attestation spec. I could see it going either way.
> 
> Yeah, I'm actually not sure why we decided to have a version at the top for each document, my interpretation was that it matched the attestation spec, though maybe I forgot other reasons we had. But I had imagined that any semantic changes to the Statement layer, at least, would rev the entire attestation spec. And when I thought about versioning each layer individually, I wasn't sure we wanted to get into the territory of having mismatched versions of fields within a Statement. As in, what would it mean to have a v1.1 subject alongside a v1.0 (predicate layer)[https://github.com/in-toto/attestation/blob/main/spec/v1/predicate.md] that then contains v1.1 DigestSets? It's a little trickier with the bundle spec, since that isn't associated with a schema, per se. In any case, imo versioning each document within spec/v1 separately introduces more complexity than we need.

Ok, added CHANGELOG.md, updated version in README.md, removed 'version' tags from the other docs.

Leaving the folder as just v1 because that seems to be what we decided before?

marcelamelara:
> Leaving the folder as just v1 because that seems to be what we decided before?

Right. I think we decided we wanted to keep everything within the major version backwards compatible.

spec/v1/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+
+## v1.1
+
+-   Clarified that subjects are assumed to be immuatble and that it is
```
marcelamelara, line 5:
```suggestion
-   Clarified that subjects are assumed to be immutable and that it is
```

spec/v1/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
...
+acceptable to use a non-cryptographic digest (though cryptographic
+digests are still preferred).
```
marcelamelara, line 7: Nit: wonder if this should match the language in the spec a bit more closely.
```suggestion
digests are still strongly recommended).
```

marcelamelara (approved):
Thanks @TomHennen ! I only have a couple nits.

pxp928 (approved):
LGTM, I agree with the changes based on the discussion that have already taken place in this PR.

spec/v1/CHANGELOG.md
TomHennen: Well, that's embarrassing.  I need to find a better setup when working with GitHub (one that provides spell-check :p).

Fixed!

spec/v1/CHANGELOG.md
TomHennen: Makes sense to me.

TomHennen:
Thanks all!

spec/v1/digest_set.md
```diff
@@ -23,6 +23,15 @@ algorithms below use lowercase hex encoding. Usually there is just a
...
 agility.
 
+Each entry in a DigestSet MUST be an immutable reference to an artifact. It is
+STRONGLY RECOMMENDED to use a commonly accepted, cryptographically secure digest
+algorithm to achieve this immutability. See [Use cases for non-cryptographic,
+immutable, digests](#use-cases-for-non-cryptographic-immutable-digests) for
+further guidance.
+
+Users SHOULD use a _cryptographic_ digest, but MAY use another identifier
+if the underlying implementation ensures immutability via other means.
```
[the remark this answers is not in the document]
kamealrabeai: []()

PR #344, opened 2024-04-05
Clarify that the storage specification uses the _file_ name.

### LINKED ISSUE #343: Media Type directions in envelope.md aren't clear
It says [MUST match the predicate specification name without the file extension. Predicate versioning is handled in the Statement layer](https://github.com/in-toto/attestation/blob/ad3ec5f1e6c9f56a10b7be2e366772b224a99ff8/spec/v1/envelope.md#storage-convention:~:text=MUST%20match%20the%20predicate%20specification%20name%20without%20the%20file%20extension.%20Predicate%20versioning%20is%20handled%20in%20the%20Statement%20layer.)

When we wrote this I'm fairly certain we mean the predicate specification _file_ name.  It's the filename that would have an extension to omit and the filename within the linked destination (spec/predicates) is guaranteed to be unique.  There isn't other name I'm aware of that would make sense here, but it's still not clear.

If we're agreed I'd be happy to send a PR.

TomHennen:
The examples already show the usage of the filename.

fixes #343

### DIFF
--- a/spec/v1/envelope.md
+++ b/spec/v1/envelope.md
@@ -63,4 +63,4 @@ Example media types for single DSSE-signed attestation predicates include:
 [Statement]: statement.md
 [in-toto-verify]: https://github.com/in-toto/in-toto#verification
 [functionaries]: https://github.com/in-toto/docs/blob/v1.0/in-toto-spec.md#212-functionaries
-[predicate specification name]: ../predicates
+[predicate specification filename]: ../predicates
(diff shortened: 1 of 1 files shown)


marcelamelara (approved):
Thanks for this update @TomHennen ! Per https://github.com/in-toto/attestation/issues/343#issuecomment-2040704256, I'm ok with this change.

pxp928 (approved):
LGTM

Issue #342, opened 2024-04-02
Conflict in Vulnerabilities predicate definition


sherzberg-1:
> scanner.result.[*].vulnerability.severity, required **object** 

says that the severity value should be an object, but then the description says

> The severity contains a **list** to describe the severity of a vulnerability using one or more quantitative scoring method.

And the example has 

> **"severity": [**

PR #345, opened 2024-04-08
Add proto definition for vuln predicate type

hectorj2f:
This PR adds a proto definition for the vuln predicate type. This was pending since we added the vuln predicate type and it is needed to update the cosign vuln predicate types.

### DIFF
--- a/protos/in_toto_attestation/predicates/vuln/v0.1/vuln.proto
+++ b/protos/in_toto_attestation/predicates/vuln/v0.1/vuln.proto
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package in_toto_attestation.predicates.vuln.v1;
+package in_toto_attestation.predicates.vuln.v01;
 
 import "google/protobuf/struct.proto";
 import "google/protobuf/timestamp.proto";
@@ -16,9 +16,9 @@ message Vuln {
 
 message Scanner {
   string uri = 1;
-  string version = 2;
+  optional string version = 2;
   VulnDatabase database = 3;
-  Result result = 4;
+  repeated Result result = 4;
 }
 
 message VulnDatabase {
@@ -28,10 +28,6 @@ message VulnDatabase {
 }
 
 message Result {
- repeated Vulnerability vulnerabilities = 3;
-}
-
-message Vulnerability {
   string id = 1;
   repeated Severity severity = 2;
 
--- a/protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
+++ b/protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
@@ -1,0 +1,50 @@
+syntax = "proto3";
+
+package in_toto_attestation.predicates.vuln.v1;
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/in-toto/attestation/go/predicates/vuln/v1";
+option java_package = "io.github.intoto.attestation.predicates.vuln.v1";
+
+// Validation of all fields is left to the users of this proto.
+message Vuln {
+  Scanner scanner = 1;
+  ScanMetadata scan_metadata = 2;
+}
+
+message Scanner {
+  string uri = 1;
+  string version = 2;
+  VulnDatabase database = 3;
+  Result result = 4;
+}
+
+message VulnDatabase {
+  optional string uri = 1;
+  optional string version = 2;
+  google.protobuf.Timestamp last_update = 3;
+}
+
+message Result {
+ repeated Vulnerability vulnerabilities = 3;
+}
+
+message Vulnerability {
+  string id = 1;
+  Severity severity = 2;
+
+  message Severity {
+    string method = 1;
+    string score = 2;
+  }
+
+  google.protobuf.Struct annotations = 3;
+}
+
+message ScanMetadata {
+  google.protobuf.Timestamp scan_started_on = 1;
+  google.protobuf.Timestamp scan_finished_on = 2;
+}
+
--- a/protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
+++ b/protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
@@ -33,14 +33,14 @@ message Result {
 
 message Vulnerability {
   string id = 1;
-  Severity severity = 2;
+  repeated Severity severity = 2;
 
   message Severity {
     string method = 1;
     string score = 2;
   }
 
-  google.protobuf.Struct annotations = 3;
+  repeated google.protobuf.Struct annotations = 3;
 }
 
 message ScanMetadata {
--- a/spec/predicates/README.md
+++ b/spec/predicates/README.md
@@ -34,7 +34,7 @@ our [vetting process], and may be of general interest:
 [Release]: release.md
 [Runtime Traces]: runtime-trace.md
 [SCAI Report]: scai.md
-[Vulnerability]: vuln.md
+[Vulnerability]: vulns.md
 [SLSA Provenance]: https://slsa.dev/provenance
 [SLSA Verification Summary]: vsa.md
 [SPDX]: spdx.md
--- a/spec/predicates/README.md
+++ b/spec/predicates/README.md
@@ -23,7 +23,7 @@ our [vetting process], and may be of general interest:
     artifact.
 -   [SPDX]: SPDX-formatted BOM for software artifacts.
 -   [CycloneDX]: CycloneDX BOM for software artifacts.
--   [Vulnerability]: Defines the metadata to share the results of vulnerability scanning on software artifacts.
+-   [VULNS]: Defines the metadata to share the results of vulnerability scanning on software artifacts.
 -   [Release]: Details an artifact that is part of a given release version.
 -   [Test Result]: A generic schema to express results of any type of tests.
 -   [Reference]: References documents that are relevant to some resource.
@@ -34,7 +34,7 @@ our [vetting process], and may be of general interest:
 [Release]: release.md
 [Runtime Traces]: runtime-trace.md
 [SCAI Report]: scai.md
-[Vulnerability]: vulns.md
+[VULNS]: vulns.md
 [SLSA Provenance]: https://slsa.dev/provenance
 [SLSA Verification Summary]: vsa.md
 [SPDX]: spdx.md
--- a/spec/predicates/README.md
+++ b/spec/predicates/README.md
@@ -34,7 +34,7 @@ our [vetting process], and may be of general interest:
 [Release]: release.md
 [Runtime Traces]: runtime-trace.md
 [SCAI Report]: scai.md
-[VULNS]: vulns.md
+[VULNS]: vulns_02
 [SLSA Provenance]: https://slsa.dev/provenance
 [SLSA Verification Summary]: vsa.md
 [SPDX]: spdx.md
--- a/protos/README.md
+++ b/protos/README.md
@@ -20,6 +20,8 @@ predicates have protobuf definitions:
     artifact.
 -   [SCAI]: Evidence-based assertions about software artifact and supply
     chain attributes.
+-   [VULN]: Describes how to store the results of scanners when detecting vulnerabilities in a software artifact.
+    chain attributes.
 -   [Test Result]: Expresses the result of a test run in the software supply
     chain.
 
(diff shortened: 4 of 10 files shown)


go/predicates/vuln/v1/vuln.pb.go
TomHennen, line 1: I believe you don't need (and shouldn't!) include these files in your change.  Instead they'll be autogenerated by this workflow https://github.com/in-toto/attestation/blob/main/.github/workflows/make-protos.yml

Can you remove this these files from this PR?

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
```diff
@@ -0,0 +1,50 @@
...
+  google.protobuf.Timestamp last_update = 3;
+}
+
+message Result {
+ repeated Vulnerability vulnerabilities = 3;
+}
+
+message Vulnerability {
+  string id = 1;
+  Severity severity = 2;
```
TomHennen, line 36: I think this is where we need resolution to https://github.com/in-toto/attestation/issues/342

As defined _here_ a given vulnerability can only have a single severity and method.  In practice there are many ways to measure severity and users may need to express both in the vuln attestation.

We can see this in practice in [the example](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md#:~:text=CVE%2D123%22%2C-,%22severity%22%3A%20%5B,-%7B%20%22method%22).

So perhaps this should be repeated?

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
```diff
@@ -0,0 +1,50 @@
...
+
+  message Severity {
+    string method = 1;
+    string score = 2;
+  }
+
+  google.protobuf.Struct annotations = 3;
```
TomHennen, line 43: This is [also a list](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md#:~:text=scanner.result.%5B*%5D.vulnerability.annotations%2C%20optional%20list) so perhaps it should be repeated too?

TomHennen (changes requested):
Thanks for sending this.

Besides the other comments, have you by any chance generated any json with this?  Could you provide a sample?  it would be nice to compare it against the example https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md#example

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
hectorj2f: Good catch.

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
hectorj2f: I'll make all these changes. 👍🏻

AungkoMCM (approved):
#348 
https://github.com/in-toto/attestation/issues/348#issue-2250696904

marcelamelara:
Friendly ping on this PR. @hectorj2f Could you please remove the generated Go/Python/Java files and add an entry for this proto [here](https://github.com/in-toto/attestation/blob/main/protos/README.md)? then I think your PR will be pretty much ready to go.

hectorj2f:
@marcelamelara Yes, I'll do it.

TomHennen:
I think something might have gotten a bit screwy with the git commits (I've done this myself for sure).  When I look at 'Files Changed' it doesn't look like anything has changed since our earlier comments?

hectorj2f:
I'm confused now. I followed what @marcelamelara commented above remove all the generated go/python and java code and leave the entry in protos.

marcelamelara:
If I understand correctly, having only the .proto file is what Tom had originally requested. But there may have been some edits to that proto file that were lost when the auto-generated files were removed? I only see 1 commit in this PR right now. 

Separately, I also had requested a minor documentation update to the /protos/README.md file. This way this new proto is included in that list.

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
```diff
@@ -0,0 +1,50 @@
...
+
+// Validation of all fields is left to the users of this proto.
+message Vuln {
+  Scanner scanner = 1;
+  ScanMetadata scan_metadata = 2;
+}
+
+message Scanner {
+  string uri = 1;
+  string version = 2;
```
TomHennen, line 19: nit: version should be marked `optional`?

https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md#:~:text=scanner.version%2C-,optional,-string%20(ResourceURI)

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
```diff
@@ -0,0 +1,50 @@
...
+  VulnDatabase database = 3;
+  Result result = 4;
```
TomHennen, line 21: If I'm reading this correctly I think `result` is supposed to be a list, but `vulnerability` is supposed to be singular.

https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md#:~:text=scanner.result%2C%20required-,list,-The%20result%20contains

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
```diff
@@ -0,0 +1,50 @@
...
+}
+
+message VulnDatabase {
+  optional string uri = 1;
+  optional string version = 2;
+  google.protobuf.Timestamp last_update = 3;
+}
+
+message Result {
+ repeated Vulnerability vulnerabilities = 3;
```
TomHennen, line 31: As mentioned above I think this should just be singular and optional. https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md#:~:text=result.%5B*%5D.vulnerability%2C%20optional-,object,-The%20vulnerability%20object

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
```diff
@@ -0,0 +1,50 @@
...
+}
+
+message Vulnerability {
+  string id = 1;
+  repeated Severity severity = 2;
```
TomHennen, line 36: How would you feel about also updating https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md to mark `severity` as a list? https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md#:~:text=vulnerability.severity%2C%20required-,object,-The%20severity%20contains  (which would close #342)

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
+
+package in_toto_attestation.predicates.vuln.v1;
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/in-toto/attestation/go/predicates/vuln/v1";
```
TomHennen, line 8: nit: currently this is version v0.1 so maybe `v01`?

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
TomHennen: This might entail bumping the version number to v0.2?

marcelamelara: +1 to sync'ing the vuln predicate spec to close that issue as part of this PR. I'm fine with bumping the version to v0.2.

protos/in_toto_attestation/predicates/vuln/v1/vuln.proto
hectorj2f: I am thinking the spec shouldn't consider `result` as a list. It looks logical to me to have results as an object where there is a list of vulnerabilities. What do you think ?

hectorj2f: So I am proposing to change the spec so the result is singular and the vulnerability is a list.

TomHennen: I think that's fine with me (especially if we're updating to v0.2).  if that's what you want can you update the docs in https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md  and the example there too?

marcelamelara (approved):
Thanks @hectorj2f ! LGTM.

marcelamelara:
One final question (which can be addressed in a separate PR, I think) is about the predicate type URI for this predicate. Right now, in the [vuln predicate spec](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md) it's listed as `https"//in-toto.io/attestation/vuln_s_` rather than `/vuln` which is the name we use everywhere else. Does the spec have a typo? We usually want the name to match everywhere, but I'm also worried that tooling might already be using the pluralized type URL. 

@hectorj2f What do you think? Is this a concern?

protos/in_toto_attestation/predicates/vuln/v0.1/vuln.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
+
+package in_toto_attestation.predicates.vuln.v01;
```
lumjjb, line 3: Should this be `vulns` instead? 

[Spec](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md) seems to be `vulns`, i do know at least 1 implementation already doing that, so my plead would be to go with vulns for `v0.1` :)

protos/in_toto_attestation/predicates/vuln/v0.1/vuln.proto
```diff
@@ -0,0 +1,46 @@
...
+message Result {
+  string id = 1;
+  repeated Severity severity = 2;
+
+  message Severity {
+    string method = 1;
+    string score = 2;
+  }
+
+  repeated google.protobuf.Struct annotations = 3;
```
lumjjb, line 39: i think the spec needs some clarification on this struct (i created https://github.com/in-toto/attestation/issues/390 to ask for some clarification)

protos/in_toto_attestation/predicates/vuln/v0.1/vuln.proto
```diff
@@ -0,0 +1,46 @@
...
+  repeated Result result = 4;
+}
+
+message VulnDatabase {
+  optional string uri = 1;
+  optional string version = 2;
+  google.protobuf.Timestamp last_update = 3;
+}
+
+message Result {
```
lumjjb, line 30: I think there's something weird going on with the result struct in the spec, i opened https://github.com/in-toto/attestation/issues/391 to discuss.

hectorj2f:
@marcelamelara Let's change vuln to vulns wherever we use it as part of a different PR.

protos/in_toto_attestation/predicates/vuln/v0.1/vuln.proto
hectorj2f: @lumjjb I've made some changes about it.

hectorj2f:
@marcelamelara @lumjjb I've applied some of the changes you proposed.

spec/predicates/README.md
```diff
@@ -34,7 +34,7 @@ our [vetting process], and may be of general interest:
 [Release]: release.md
 [Runtime Traces]: runtime-trace.md
 [SCAI Report]: scai.md
-[Vulnerability]: vuln.md
+[Vulnerability]: vulns.md
```
marcelamelara, line 37: The other README uses the shorter name "VULNS", so it would be good to name these consistently. I do think that "Vulnerability" or even "Vulnerability Scan" would be clearer for the long name ("vulns" is still ok for the short predicate name).

marcelamelara (changes requested):
Thanks so much for the updates @hectorj2f ! At this point the changes I'm requesting are just some small documentation fixes.

marcelamelara (approved):
Thanks @hectorj2f ! LGTM

TomHennen:
I appreciate these changes, but the one big problem I see is that it erases the old 'vuln v0.1' predicate and replaces it with this new one.

There are folks that are using the old predicate (however flawed it might be me), we should keep that copy around for them to reference.

So a couple requests:

1. Leave the old vuln.md there (folks probably have links to it).  Maybe add a note pointing them at the newer version.
2. Update the version in the new vulns to 0.2
3. Update the proto to match 0.2. (including it's path in the Protos directory)
4. Rename vulns.md to vulns_02.md

hectorj2f:
@TomHennen I believe I've made all your requested changes.

TomHennen:
Thank you!  And I'm so sorry it took me so long to get back to this.  I missed the email about the update. :(

TomHennen (approved):

PR #359, opened 2024-05-07
Clarify versioning and tagged release process

marcelamelara:
This PR adds a short description clarifying _when_ the attestation spec version is rev'd and how that relates to tagged releases.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -24,6 +24,16 @@ The source of this diagram can be found [here](../images/envelope_relationships.
 The [validation model] provides pseudocode showing how these layers fit
 together. See the [documentation] for more background and examples.
 
+## Tagged Releases
+
+The latest [tagged release] version matches the [SemVer](https://semver.org)
+MAJ.MIN version of the Attestation Framework spec.
+
+Backwards-compatible semantic updates to the spec (except predicates) are
+indicated through new tagged minor version releases.
+We use new tagged patch version releases to indicate updates to predicate
+specifications and/or backwards-compatible changes to the language bindings.
+
 ## Keywords
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
@@ -41,5 +51,6 @@ this specification are to be interpreted as described in [RFC 2119].
 [Statement]: v1/statement.md
 [documentation]: ../docs
 [in-toto-verify]: https://github.com/in-toto/in-toto#verification
+[tagged release]: https://github.com/in-toto/attestation/releases
 [v1.1]: v1/README.md
 [validation model]: ../docs/validation.md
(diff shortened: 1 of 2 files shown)


spec/README.md
```diff
@@ -24,6 +24,16 @@ The source of this diagram can be found [here](../images/envelope_relationships.
...
 together. See the [documentation] for more background and examples.
 
+## Tagged Releases
+
+The latest [tagged release] version matches the [SemVer](https://semver.org)
+MAJ.MIN version of the Attestation Framework spec.
+
+Backwards-compatible semantic updates to the spec (except predicates) are
+indicated through new tagged minor version releases.
+We use new tagged patch version releases to indicate updates to predicate
```
TomHennen, line 34:
```suggestion

Updates to predicate specifications (backwards-compatible or not), and _backwards-compatible_ changes to the language bindings are also indicated via minor version releases.
```

TomHennen: Apparently I don't know how to make the suggestion span multiple lines because it should take line 35 as well. :/

marcelamelara: Ah so what you're suggesting is different from what we had been doing so far. So far. we've been bumping the patch version to indicate changes in everything but the statement layer spec to keep the major/minor version in line with the core spec. What you're suggesting seems to be that we might release a `v1.2` release even though we haven't touched the statement layer? I don't have a strong preference, though my sense is that the tagged release shouldn't diverge from the core spec version, mostly as a signal to users.

TomHennen: Oh!  Then I just misunderstood, I didn't mean to change your suggestion.  Perhaps to avoid confusion user UPPERCASE for 'PATCH' and provide some examples?

marcelamelara: Can do! Examples would definitely be a good addition.

marcelamelara: Done.

TomHennen (approved):
Thanks, this is great!

spec/README.md
```diff
@@ -24,6 +24,36 @@ The source of this diagram can be found [here](../images/envelope_relationships.
...
+
+Backwards-compatible semantic updates to the spec (except predicates) are
+indicated through new tagged MINOR version releases.
+We use new tagged PATCH version releases to indicate updates to predicate
+specifications and/or backwards-compatible changes to the language bindings.
+
+### Examples
+
+-   Attestation Framework tagged release v1.0.2 (PATCH version) incorporates
+    refinements to the predicate specification process, a new predicate type,
```
adityasaky, line 40: IMO it is a bit confusing for semantic versioning of the spec to _not_ apply to predicate changes (which I agree with) while also saying a spec version ships with a new predicate. Maybe the version of the spec should entirely omit predicates and it's just a matter of convenience that predicates live alongside the spec?

This does get a bit confusing with versioning of the code gen libraries though, as you highlight here.

marcelamelara: I see what you mean. The intent here isn't to say that none of the things under the PATCH version category really affect the core spec. Maybe we can take out the mention of predicate types altogether and say that PATCH versions are only for doc edits and language binding changes?

adityasaky: I think that might be better!

pxp928 (approved):
LGTM!

Issue #179, opened 2023-03-28
Add a ResourceDescriptor predicate


marcelamelara:
There are some use cases that need to communicate that a document or artifact exists and don't necessarily need to reference it from within some pre-existing in-toto statement (per https://github.com/in-toto/attestation/issues/117#issuecomment-1333815831). 

The [ResourceDescriptor](https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md) is designed for describing and pointing to other artifacts within an in-toto attestation, so it can be used to implement this attestation predicate. The main usage of this predicate type is to have a signature over the referenced document or artifact without having to sign the entire document directly.

Tasks:
- [ ] Write a resource descriptor attestation predicate definition (following the [new predicate guidelines](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md))
- [ ] Update the spec/predicates/README.md
- [ ] Write a protobuf definition for the predicate
- [ ] Add a test for this predicate to the test suite for at least one of the supported language bindings (Go, Python, Java, Rust)

adityasaky:
I think this may overlap with #124. I _am_ a bit worried about fragmentation though.

marcelamelara:
> I think this may overlap with #124. I _am_ a bit worried about fragmentation though.

Can you elaborate what you mean? I'm not immediately seeing the overlap.

adityasaky:
So the other thread is about a "source" attestation recording artifact sources. Each source is then going to be a resource descriptor. As it's proposed, I think we're essentially talking there about the link predicate. But if we do want to carve out a subset predicate that describes sources, cutting other information like `environment` and so on out, I think we end up with 1+ resource descriptors. This is for just the one RD instance. The question is how different are all of these predicate options and should be we collect them as one predicate type? If we have a single RD type and a source predicate type, I imagine either would be usable in single-source scenarios.

marcelamelara:
Ah, thanks for clarifying. I guess there are two schemas in that issue, and you're right that the schema proposed by @trishankatdatadog and this RD predicate do have significant overlap. I was looking at the schema that Cole had posted, and that one contains a lot more info than just a list of RDs. I tend to agree that if source and RD predicates will mostly be the same, we should develop a single predicate that can meet the needs of both use cases.

[...]

PR #375, opened 2024-08-01
Add the reference predicate

mdeicas:
Adds the specification of a predicate to address #179.

One question I had was how to express what content the reference contains. I think the current use cases are met by the MIME type, and adding another field to specify this more explicitly would make the predicate more complex. If needed, it can be added in the future in a new version.

Open to feedback!

### DIFF
--- a/spec/predicates/reference.md
+++ b/spec/predicates/reference.md
@@ -1,0 +1,143 @@
+# Predicate type: Reference
+
+Type URI: https://in-toto.io/attestation/reference/v0.1
+
+Version: v0.1.0
+
+## Purpose
+
+The reference attestation references a set of documents that are relevant to
+some resource. It is used to point to and sign metadata that is not
+provided in-band.
+
+## Use Cases
+
+One use case is to provide an SBOM of an artifact without needing to sign
+the entire SBOM and transmit it in-band. A consumer, such as
+[GUAC](https://guac.sh/), retrieves the SBOM, verifies the digest, and uses it
+for vulnerability management, compliance, analysis, or any other purpose.
+
+Another use case is to provide evidence for a policy such as "deny unless a
+reference to an SBOM is provided".
+
+There is encodability overlap with some other predicates, mainly SCAI, but
+the use case is distinct.
+
+## Prerequisites
+The
+[in-toto Attestation Framework](https://github.com/in-toto/attestation/blob/main/spec/README.md)
+and an SBOM specification such as [SPDX](https://spdx.dev/).
+
+## Model
+This predicate is intended to be generated and consumed throughout the software
+supply chain. In addition, it is intended to be used in the analysis of it as a
+whole.
+
+## Schema
+
+```jsonc
+{
+  // Standard attestation fields:
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [{ ... }],
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/reference/v0.1",
+  "predicate": {
+    "attester": {
+      "id": "<TypeUri>"
+    },
+    "references": ["<ResourceDescriptor>", ...]
+  }
+}
+```
+
+### Parsing Rules
+This predicate follows the
+[in-toto Attestation Framework's parsing rules](../v1/README.md#parsing-rules).
+
+The `references` field is repeated as an optimization: multiple references
+SHOULD be treated as if each reference was represented in its own attestation
+with all other content unchanged.
+
+The reference(s) are associated with the set of subjects as a whole, instead of
+being independently associated with each subject. See the
+[example](#reference-to-an-sbom-for-multiple-artifacts) with two subjects.
+
+### Fields
+`attester.id`: string ([TypeUri](../v1/field_types.md#typeuri)), *required*
+
+An identifier for the system that provides the document.
+
+`references` array of [ResourceDescriptor](../v1/field_types.md#resource_descriptor) objects, *required*
+
+The referred documents. The `mediaType` and `downloadLocation` of each SHOULD
+be provided.
+
+## Examples
+### Reference to an SBOM for an image
+```json
+{
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [{
+    "name": "registry.example.com/my-project/my-image",
+    "digest": {
+      "sha256": "886a77e9b9c993e221d1843e3af5b1fbeea26d8962995e8562174a6aba0c7cc9"
+    }
+  }],
+  "predicateType": "https://in-toto.io/attestation/reference/v0.1",
+  "predicate": {
+    "attester": {
+      "id": "https://my-builder.com/v1"
+    },
+    "references": [{
+      "downloadLocation": "https://cloud.com/my-sboms/sbom.spdx.json",
+      "digest": {
+        "sha256": "3b0d19b348f1e46a571c6e9df32897637c20dd9803261c2fc9cbe38b0c8422c4"
+      },
+      "mediaType": "application/spdx+json"
+    }]
+  }
+}
+```
+### Reference to an SBOM for multiple artifacts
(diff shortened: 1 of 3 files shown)


pxp928 (approved):
This looks good to me for a first revision Thank You!

spec/predicates/reference.md
```diff
@@ -0,0 +1,143 @@
...
+[example](#reference-to-an-sbom-for-multiple-artifacts) with two subjects.
+
+### Fields
+`attester.id`: string ([TypeUri](../v1/field_types.md#typeuri)), *required*
+
+An identifier for the system that provides the document.
+
+`references` array of [ResourceDescriptor](../v1/field_types.md#resource_descriptor) objects, *required*
+
+The referred documents. The `mediaType` and `downloadLocation` of each SHOULD
```
TomHennen, line 74: Should `downloadLocation` be "MUST"?  How useful would a reference attestation that doesn't point to something be?

TomHennen, line 74: optional: should `mediaType` be required with a suggested 'default' type if the user doesn't know what it is?

spec/predicates/reference.md
```diff
@@ -0,0 +1,143 @@
...
+One use case is to provide an SBOM of an artifact without needing to sign
+the entire SBOM and transmit it in-band. A consumer, such as
+[GUAC](https://guac.sh/), retrieves the SBOM, verifies the digest, and uses it
+for vulnerability management, compliance, analysis, or any other purpose.
+
+Another use case is to provide evidence for a policy such as "deny unless a
+reference to an SBOM is provided".
+
+There is encodability overlap with some other predicates, mainly SCAI, but
+the use case is distinct.
```
TomHennen, line 24: Can you say more about why SCAI isn't quite right for this case?  (E.g. can we answer [this question](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md#:~:text=Why%20don%E2%80%99t%20existing%20predicates%20cover%20this%20use%20case%3F))

TomHennen:
Thanks for sending this!

Would it be possible to create a [protobuf definition](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md#:~:text=To%20generate%20Go/Python/Java%20language%20bindings%20for%20the%20new%20predicate%2C%20include%20a%20protobuf%20definition.)

spec/predicates/reference.md
mdeicas: Makes sense to me, changed to MUST.

mdeicas: `application/octet-stream` [is specified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types#:~:text=contain%20binary%20data.-,application/octet%2Dstream,-is%20the%20default) as the default MIME type for unknown file types. Having a required field is more usable, so this sounds good to me.

spec/predicates/reference.md
mdeicas: Done.

marcelamelara:
I don't see how SCAI can't cover this use case. In fact, we have a [few](https://github.com/in-toto/scai-demos/tree/main/examples/sbom%2Bslsa) [examples]((https://github.com/in-toto/attestation/blob/main/spec/predicates/scai.md#attestation-for-evidence-collection) that show how SCAI can be used to transmit SBOM (or any other type of reference to another doc) without having to embed an entire SBOM in a Statement. 

Rather than re-inventing the wheel for one specific use case, I'd much rather we do attempt to standardize around specific attribute values (e.g., our example in scai-demos uses the GUAC verbs `HasSBOM` and `HasSLSA`), and declare certain fields in the SCAI predicate as required when used in certain ways, not unlike how we do for [DigestSets](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#supported-algorithms), where the spec does in fact standardize certain field values.

marcelamelara:
To add, I'm absolutely happy to work with you to explore different fields/values that could be standardized for this common use case (i.e. exchanging SBOM, having it consumed by GUAC). For instance, for this use case, the `target` field of the SCAI predicate may be required if `attribute: HasSBOM` and `mediaType: application/spdx+json`. Again, there's nothing in the existing in-toto attestation spec that precludes these sorts of restrictions for specific use cases.

lumjjb:
(copying my response from the original issue since it may be relevant to discussion)

> 
> I think this is slightly different framing from the witness/provenance materials, and resembling more of an intoto link predicate (although the link predicate is a bit generic, so having a defined predicate will help the consumption/adoption by being opinionated on how fields should be used).
> 
> i can see where folks are coming from and indeed there is "encodability" overlap with the different predicates, but I agree with sentiment that the intent here is different. The fields required for supporting documents vs source code attestations vs security statements are different, and i believe it is helpful to be intentional with defining each of them. Else, we may see issues around there being too many optional fields (which also have different use based on context), and that can cause confusion. I'll use SPDX as an example (since i am a contributor and thus an offender), where a "Package" can refer to many things (sources, application bundles) and has many optional fields depending on what it represents (which 3.0 does try and solve by providing a bit more specificity in subclassing, but still suffers from legacy convention).

Agreed on the overall goal of having a "predicate dictionary" type construct where we define here are all the predicates or properties and having guidance on how to consume them. It is difficult for me (who is perhaps uninformed) to understand the delineation around predicates and when something is a separate enough usecase for its own predicate. e.g. SCAI vs vulnerability, intoto link vs SLSA.

mdeicas:
Thanks for the responses. IMO a new predicate type has a few benefits over SCAI, but I agree that SCAI does work for the current use case at a technical level, and I'm happy to go with the option that the maintainers think is best. Is the SCAI predicate type spec a good place to define the attribute?

The benefits I see for this dedicated predicate type is that it can evolve more easily (e.g. adding new fields) and is simpler to use for consumers (possibly fewer changes to the spec and the data "type" is encoded at the statement layer instead of the predicate layer).

marcelamelara:
Sorry for my delay.

> It is difficult for me (who is perhaps uninformed) to understand the delineation around predicates and when something is a separate enough usecase for its own predicate.

This is totally fair, and the more general feedback for us here is that we should set clearer guidelines for users of in-toto attestations to choose the right predicate(s), or describe more clearly when it really makes sense to create new ones. We have an upcoming maintainer's meeting, this would be a good topic to raise. Thanks!!

> Is the SCAI predicate type spec a good place to define the attribute?

Yes, what I'm imagining is adding a new section to the spec titled something like "Supported Assertions" where we would catalogue the GUAC attributes and expected other field requirements/values for those assertions. This would also entail adding the appropriate verification logic to the SCAI APIs, much how we validate [supported hash algorithms](https://github.com/in-toto/attestation/blob/main/go/v1/resource_descriptor.go) in resource descriptors.

>The benefits I see for this dedicated predicate type is that it can evolve more easily (e.g. adding new fields) and is simpler to use for consumers (possibly fewer changes to the spec and the data "type" is encoded at the statement layer instead of the predicate layer).

I definitely understand this rationale. And the predicates are meant to be living specs that evolve as the community gathers feedback from users/consumers about use cases or usability. This is meant to be a community-driven process, so subsequent versions of the SCAI predicate (or any other) can definitely be amended to add/change fields, and better suit the broad use cases they were designed for.

Could you please explain what you mean by "the type is encoded at the statement layer" in the case of the dedicated reference predicate? If a consumer just sees `predicateType: "https://in-toto.io/attestation/reference/v0.1"`, it still needs to step into the predicate layer and iterate over the list of references in order to understand what types of documents/resources are being referenced. I could just be missing something.

TomHennen (approved):

marcelamelara:
Given that we have the needed approvals for this PR, I won't stand in the way of merging.

spec/predicates/reference.md
mdeicas: I think if the the `uri` is left empty and `digest` is provided, then wouldn't be a conflict.

Issue #371, opened 2024-07-03
Making predicateTypes consistent with predicate names


marcelamelara:
We recently introduced the notion of the predicate name, which we use in a couple ways: as a hint in the [envelope mediaType](https://github.com/in-toto/attestation/blob/main/spec/v1/envelope.md#storage-convention), and in the `predicateType` URI for [predicates in the in-toto/attestation namespace](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md#vetting-process) (see step 4). 

Most predicates in the in-toto/attestation namespace already follow this convention for the predicateTypes, but we have two that don't:

* scai: https://in-toto/attestatoin/scai/attribute-report/v* (adds extra dir in path)
* vuln: https://in-toto/attestation/vulns/v* (adds S to predicate name).

How should we resolve these two cases? 

My recommendation for SCAI is to remove the "attribute-report" piece since it somehow implies that there may be other subtypes of SCAI, and we don't currently support predicate subtypes. This may break existing tooling (mostly in-toto/scai-demos), and we may need to bump the predicate version number.

My suggestion for vuln may be to update the predicate name to `vulns.md` since that won't break current tools.

Any other thoughts? If this looks good, I'll open a PR to make the relevant changes.

marcelamelara:
Update: We may probably also move the expected naming convention to the predicate template in `spec/predicates/templates`.

adityasaky:
> My suggestion for vuln may be to update the predicate name to vulns.md since that won't break current tools.

This seems reasonable to me.

PR #380, opened 2024-08-30
Simplify SCAI predicate TypeURI

marcelamelara:
Unlike the [recommended naming convention](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md#vetting-process), the SCAI predicate TypeURI has an extra element `/attribute-report`. This may confusingly signal other types of SCAI predicates, which don't exist AFAIK, and is therefore redundant. This PR makes the type URI conform with the expected convention, and bumps the minor version.

Related to #371 .

### DIFF
--- a/python/in_toto_attestation/predicates/scai/v0/scai.py
+++ b/python/in_toto_attestation/predicates/scai/v0/scai.py
@@ -6,6 +6,7 @@ from in_toto_attestation.v1.resource_descriptor import ResourceDescriptor
 SCAI_PREDICATE_TYPE = "https://in-toto.io/attestation/scai/"
 SCAI_PREDICATE_VERSION = "v0.3"
 
+
 class AttributeAssertion:
     def __init__(self, attribute, target=None, conditions=None, evidence=None) -> None:
         self.pb = scaipb.AttributeAssertion()  # type: ignore[attr-defined]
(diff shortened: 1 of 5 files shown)


python/in_toto_attestation/predicates/scai/v0/scai.py
```diff
@@ -3,9 +3,8 @@
 import in_toto_attestation.predicates.scai.v0.scai_pb2 as scaipb
 from in_toto_attestation.v1.resource_descriptor import ResourceDescriptor
 
-SCAI_PREDICATE_TYPE = "https://in-toto.io/attestation/scai/attribute-report/"
-SCAI_PREDICATE_VERSION = "v0.2"
-
+SCAI_PREDICATE_TYPE = "https://in-toto.io/attestation/scai/"
+SCAI_PREDICATE_VERSION = "v0.3"
```
adityasaky, line 7: I wonder if the removed newline here is the issue with the linter?

TomHennen (approved):

python/in_toto_attestation/predicates/scai/v0/scai.py
marcelamelara: But there still is a newline after the second constant. Let me see if I can get the linter to actually give us a useful error message.

adityasaky: I'm not sure if this is configured for PEP8 / a derivative, but top level function and class defs need to be surrounded by two blank lines.

https://peps.python.org/pep-0008/#blank-lines

marcelamelara: Yeah, you're right. This was the problem. In the future the linter will output the diff for what needs to change.

pxp928 (approved):

Issue #391, opened 2024-10-11
Vulns attestation "result" struct discrepancy


lumjjb:
There may be a discrepancy with the [intoto vuln predicate](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md) (if i interpret it right).

It looks like the spec specifies `scanner.result.[*].vulnerability, optional object` indicates a nested `vulnerability` object, but within the example, it shows no intermediary "vulnerability" object.

It seems like the intent is to have `scanner.result.[*] optional object` instead of `scanner.result.[*].vulnerability, optional object`?

lumjjb:
FYI @edonadei

lumjjb:
Decision was to be fixed in v0.2 which is merged in  #345.  However, the `invocation` part of it is still a problem, i will create a PR to fix it directly on v0.2 since it is reasonable that the impact of spec implementors is minimal due to it being recent.

puerco:
Following up to this one, I've opened https://github.com/in-toto/attestation/pull/434 to address two inconsistencies between the protos and the spec with the `metadata` and `scanner.db` fields.

marcelamelara:
@lumjjb @puerco if I'm reading this issue correctly, there is still a possible discrepancy related to optional subfields of the `scanner.result[*]` field that needs to be fixed?

Issue #390, opened 2024-10-11
Ambiguity for annotations in vuln attestation


lumjjb:
In the [vuln attestation](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md), annotations are specified as an optional list, but the list is not typed. 

Would this be a list of strings `["foo=bar"]`, or a struct
`[ {{"key":"foo"}, {"value": "bar"}}]`, or a list of single maps? `[{"foo": "bar"}]`

```
scanner.result.[*].vulnerability.annotations, optional list

This is a list of key/value pairs where scanners can add additional custom information.
```

hectorj2f:
@lumjjb This would be more of a map <string, value> in my opinion.

lumjjb:
Resolved by #345 through introducing new version 0.2

PR #392, opened 2024-10-11
Changes to vuln 0.1 spec and documentation

lumjjb:
Addressing #391 and #390, and making several changes which help with implementing the spec. Comments welcome! Am happy to split up the changes if some of them may be potentially controversial. Tagging original author: @hectorj2f 

#390
- List types updated to disambiguate list types.

#391 
- Remove invocation from example (which is not in the spec).
- Removed the additional "vulnerability" object indirection
  from the spec since it is the only field within it (as reflected in
  the example).
- Severity changed to a list to contain multiple scores (as reflected in
  the example).

Make several fields which may not always be available optional. 
- ScanStartedOn changed from required to optional as it may not always be
  available.
- scanner.db.lastUpdate changed from required to optional since this information
  may not always be available when a DB uri/version is known.

### DIFF
--- a/spec/predicates/vuln.md
+++ b/spec/predicates/vuln.md
@@ -64,41 +64,39 @@ The `predicate` contains a JSON-encoded data with the following fields:
 
 > > > The version of the Vulnerability DB.
 
-**scanner.db.lastUpdate, required** string (timestamp)
+**scanner.db.lastUpdate, optional** string (timestamp)
 
 > > > The timestamp of when the vulnerability DB was updated last time.
 
-**scanner.result, required** list
+**scanner.result, required** object list
 
 > > The result contains a list of vulnerabilities. Note that an empty list means the **scanner** found no vulnerabilities.
 > > This is the most important part of this field because it'll store the scan result as a whole. So, people might want
 > > to use this field to take decisions based on them by making use of Policy Engines tooling whether allow or deny these images.
+> > Each object defines information about each one of the vulnerabilities found by the scanner.
 
-**scanner.result.[*].vulnerability, optional** object
 
-> > > The vulnerability object defines information about each one of the vulnerabilities found by the scanner.
+**scanner.result.[*].id, required** string
 
-**scanner.result.[*].vulnerability.id, required** string
+> > > This is the identifier of the vulnerability, e.g. [GHSA-fxph-q3j8-mv87](https://github.com/advisories/GHSA-fxph-q3j8-mv87) whose CVE id is [CVE-2017-5645](https://nvd.nist.gov/vuln/detail/CVE-2017-5645).
 
-> > > > This is the identifier of the vulnerability, e.g. [GHSA-fxph-q3j8-mv87](https://github.com/advisories/GHSA-fxph-q3j8-mv87) whose CVE id is [CVE-2017-5645](https://nvd.nist.gov/vuln/detail/CVE-2017-5645).
+**scanner.result.[*].severity, required** object list
 
-**scanner.result.[*].vulnerability.severity, required** object
+> > > The severity contains a list to describe the severity of a vulnerability using one or more quantitative scoring method.
 
-> > > > The severity contains a list to describe the severity of a vulnerability using one or more quantitative scoring method.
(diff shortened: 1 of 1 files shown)


hectorj2f:
@lumjjb Could we get this PR merged first https://github.com/in-toto/attestation/pull/345 ?

TomHennen:
> @lumjjb Could we get this PR merged first #345 ?

That makes sense to me.  Also, please let me know if you're waiting on me for #345.  I don't want to hold anyone up!

lumjjb:
@hectorj2f can you validate if any of the suggested changes here would reflect in #345 as well? 

fyi, i don't have maintainer bit here, so no control over that :)

spec/predicates/vuln.md
```diff
@@ -64,41 +64,39 @@ The `predicate` contains a JSON-encoded data with the following fields:
 
 > > > The version of the Vulnerability DB.
 
-**scanner.db.lastUpdate, required** string (timestamp)
+**scanner.db.lastUpdate, optional** string (timestamp)
```
hectorj2f, line 67: I won't change this setting. It is important to reflect the update of the database used for the scanner.

spec/predicates/vuln.md
```diff
@@ -64,41 +64,39 @@ The `predicate` contains a JSON-encoded data with the following fields:
...
-> > > > > This is a string representing the severity score based on the selected method.
+**scanner.result.[*].annotations, optional** object list
 
-**scanner.result.[*].vulnerability.annotations, optional** list
+> > > This is a list of key/value pairs where scanners can add additional custom information.
 
-> > > > > This is a list of key/value pairs where scanners can add additional custom information.
-
-**metadata.scanStartedOn, required** Timestamp
+**metadata.scanStartedOn, optional** Timestamp
```
hectorj2f, line 99: I have the same thoughts as above. This must be required to ensure all the details to reuse these attestations in the future.

spec/predicates/vuln.md
```diff
@@ -64,41 +64,39 @@ The `predicate` contains a JSON-encoded data with the following fields:
...
 
 > > The result contains a list of vulnerabilities. Note that an empty list means the **scanner** found no vulnerabilities.
 > > This is the most important part of this field because it'll store the scan result as a whole. So, people might want
 > > to use this field to take decisions based on them by making use of Policy Engines tooling whether allow or deny these images.
+> > Each object defines information about each one of the vulnerabilities found by the scanner.
 
-**scanner.result.[*].vulnerability, optional** object
 
-> > > The vulnerability object defines information about each one of the vulnerabilities found by the scanner.
+**scanner.result.[*].id, required** string
```
hectorj2f, line 79: These changes look good to me, we can remove the vulnerability, and assume they refer to vulnerabilities.

marcelamelara:
@lumjjb Have you rebased this onto the updated vulns predicate spec from #345 ?

alphachart (changes requested):
Check

spec/predicates/vuln.md
lumjjb: i find that this information is generally not very available. There usually is a last update on an entry of a vuln DB, but the lastUpdate of a vuln database i don't believe i've been able to populate (i'm thinking about OSV and MITRE db)

spec/predicates/vuln.md
lumjjb: i think i run into the same problem where scanners today don't provide that information as well? is there a particular use of the scan start time you're thinking of?

lumjjb:
i'm a little confused now, would this PR be negated with #345  then? since now we're pointing folks to v0.2, and v0.1 would just be deprecated?

spec/predicates/vuln.md
hectorj2f: This information is relevant for many scanners such as Grype, depending on the date of last update the report could be discarded.

spec/predicates/vuln.md
hectorj2f: This information is relevant for many scanners such as Grype, depending on the date of last update the report could be discarded.

PR #403, opened 2024-11-01
Adds custom marshal/unmarshal to Go Statement struct, fixes #363, see sigstore/sigstore-go#326

### LINKED ISSUE #363: Support for `predicateType`
We recently upgraded our dependency from `github.com/in-toto/in-toto-golang/in_toto` to `github.com/in-toto/attestation/go/v1` . While migrating, we encountered a breaking change related to the predicateType field in the Statement struct.

In the previous version (github.com/in-toto/in-toto-golang/in_toto), the StatementHeader struct was defined as:

```golang
type StatementHeader struct {
	Type          string    `json:"_type"`
	PredicateType string    `json:"predicateType"`
	Subject       []Subject `json:"subject"`
}
```

In the new version `github.com/in-toto/attestation/go/v1`, the Statement struct is defined as:

```golang
type Statement struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	Type          string                `protobuf:"bytes,1,opt,name=type,json=_type,proto3" json:"type,omitempty"`
	Subject       []*ResourceDescriptor `protobuf:"bytes,2,rep,name=subject,proto3" json:"subject,omitempty"`
	PredicateType string                `protobuf:"bytes,3,opt,name=predicate_type,json=predicateType,proto3" json:"predicate_type,omitempty"`
	Predicate     *structpb.Struct      `protobuf:"bytes,4,opt,name=predicate,proto3" json:"predicate,omitempty"`
}
```

With this when we try to run the cosign verification command it throws an error as 

```bash
osign verify-blob-attestation --insecure-ignore-tlog --key k8s://tekton-chains/signing-secrets --signature sig --type slsaprovenance --check-claims=false /dev/null   
WARNING: Skipping tlog verification is an insecure practice that lacks of transparency and auditability verification for the blob attestation.
Error: invalid predicate type, expected slsaprovenance got 
main.go:74: error during command execution: invalid predicate type, expected slsaprovenance got 
```

Because the payload has `predicate_type` instead of `predicateType`

Reference links for chains update - https://github.com/tektoncd/chains/pull/1111/files#diff-669f2bfe131a7e3f3437d73f96e40cb50d831906b7fadcc999f47beb5b855001R60-R81

gillisandrew:
This PR adds custom marshaler and unmarshaler to the statement struct to help users avoid incorrectly serializing in-toto statements. This fixes #363 and stems from a discussion in sigstore/sigstore-go#326s

### DIFF
--- a/go/v1/statement_test.go
+++ b/go/v1/statement_test.go
@@ -5,6 +5,7 @@ Tests for in-toto attestation ResourceDescriptor protos.
 package v1
 
 import (
+	"encoding/json"
 	"fmt"
 	"testing"
 
@@ -39,7 +40,7 @@ func createTestStatement(t *testing.T) *Statement {
 	}
 }
 
-func TestJsonUnmarshalStatement(t *testing.T) {
+func TestProtojsonUnmarshalStatement(t *testing.T) {
 	var wantSt = `{"_type":"https://in-toto.io/Statement/v1","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}`
 
 	got := &Statement{}
@@ -47,8 +48,36 @@ func TestJsonUnmarshalStatement(t *testing.T) {
 	assert.NoError(t, err, "error during JSON unmarshalling")
 
 	want := createTestStatement(t)
-	assert.NoError(t, err, "unexpected error during test Statement creation")
 	assert.True(t, proto.Equal(got, want), "protos do not match")
+}
+
+func TestJsonUnmarshalStatement(t *testing.T) {
+	var wantSt = `{"_type":"https://in-toto.io/Statement/v1","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}`
+
+	got := &Statement{}
+	err := json.Unmarshal([]byte(wantSt), got)
+	assert.NoError(t, err, "error during JSON unmarshalling")
+
+	want := createTestStatement(t)
+	assert.True(t, proto.Equal(got, want), "protos do not match")
+}
+
+func TestProtojsonMarshalStatement(t *testing.T) {
+	var wantSt = `{"_type":"https://in-toto.io/Statement/v1","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}`
+	want := createTestStatement(t)
+
+	gotSt, err := protojson.Marshal(want)
+	assert.NoError(t, err, "error during JSON marshalling")
+	assert.JSONEq(t, wantSt, string(gotSt), "JSON objects do not match")
+}
+
+func TestJsonMarshalStatement(t *testing.T) {
+	var wantSt = `{"_type":"https://in-toto.io/Statement/v1","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}`
+	want := createTestStatement(t)
+
+	gotSt, err := json.Marshal(want)
+	assert.NoError(t, err, "error during JSON marshalling")
+	assert.JSONEq(t, wantSt, string(gotSt), "JSON objects do not match")
 }
 
 func TestBadStatementType(t *testing.T) {
@@ -61,6 +90,8 @@ func TestBadStatementType(t *testing.T) {
 	err = got.Validate()
 	assert.ErrorIs(t, err, ErrInvalidStatementType, "created malformed Statement (bad type)")
 }
+
+
 
 func TestBadStatementSubject(t *testing.T) {
 	tests := map[string]struct {
@@ -122,3 +153,4 @@ func TestBadStatementPredicate(t *testing.T) {
 		assert.ErrorIs(t, err, test.err, fmt.Sprintf("%s in test '%s'", test.noErrMessage, name))
 	}
 }
+
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/in-toto/attestation
 
-go 1.20
+go 1.21
+
+toolchain go1.21.0
 
 require (
 	github.com/stretchr/testify v1.9.0
--- a/go.sum
+++ b/go.sum
@@ -1,11 +1,13 @@
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/google/go-cmp v0.5.5 h1:Khx7svrCpmxxtHBq5j2mp/xVjsi8hQMfNLvJFAlrGgU=
+github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/stretchr/testify v1.9.0 h1:HtqpIVDClZ4nwg75+f6Lvsy/wHu+3BoSGCbBAcpTsTg=
 github.com/stretchr/testify v1.9.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
+golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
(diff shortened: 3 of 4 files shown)


go/v1/statement_test.go
```diff
@@ -39,18 +40,46 @@ func createTestStatement(t *testing.T) *Statement {
...
-func TestJsonUnmarshalStatement(t *testing.T) {
+func TestProtojsonUnmarshalStatement(t *testing.T) {
 	var wantSt = `{"_type":"https://in-toto.io/Statement/v1","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}`
 
 	got := &Statement{}
 	err := protojson.Unmarshal([]byte(wantSt), got)
 	assert.NoError(t, err, "error during JSON unmarshalling")
 
 	want := createTestStatement(t)
-	assert.NoError(t, err, "unexpected error during test Statement creation")
```
codysoyland, line 50: I was confused about this line removal, but I see that this error is already checked above 👍🏻

codysoyland:
This gets a +1 from me!

go/v1/statement_test.go
gillisandrew: Yeah, removed the redundant checks 😅

adityasaky:
cc @pxp928 @TomHennen @marcelamelara

marcelamelara (commented):
Thanks for sending this @gillisandrew . It seems like the need for these custom functions stems from the use of the soon-to-be deprecated in-toto-golang data structures (hence why #363 was closed). 

Are these added functions needed for backward compatibility with pre-SLSA v1 Provenance attestations?

codysoyland:
I think #363 was closed erroneously, as the issue here is mostly unrelated to the in-toto-golang data structures. We encountered the same bug in sigstore-go after migrating to the data structures in this repo. The issue is specifically about how the new data structures serialize to JSON. When using the new data structures, when you pass them to `json.Marshal`, the output has field names that are not consistent with the in-toto specification (e.g. `predicate_type` instead of `predicateType`). When using protobuf types, the default serialization always assumes `snake_case` instead of `camelCase`. This PR works around that fact by explicitly using [`protojson`](https://protobuf.dev/programming-guides/json/) for marshal/unmarshal, which produces the correct JSON format that is consistent with all other tooling that works with in-toto.

gillisandrew:
@marcelamelara as cody noted, the structs in this repo are the issue, the field names emitted in the json struct tags are incorrect.

There's some background and alternative solutions in golang/protobuf#52

SantiagoTorres:
ugh, CI complains about golang 1.21, but I can confirm that wfm locally...

adityasaky:
we should update the run-go-tests.yml to use a newer version of the toolchain.

codysoyland:
> we should update the run-go-tests.yml to use a newer version of the toolchain.

That can be updated to always use what's in `go.mod` using the [`go-version-file` input](https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file).

marcelamelara:
@codysoyland and @gillisandrew thanks for the clarifications.

I'm fine with the tests you've added, thank you! But my preference would be to find an alternative way to emit structs with the right json tags from the start, rather than adding custom marshalling/unmarshalling functions because these will be needed in more places than statement.go.

marcelamelara:
Does this seem like a reasonable alternative? https://github.com/favadi/protoc-go-inject-tag

codysoyland:
This has unfortunately been an issue with protoc-gen-go for years (see https://github.com/golang/protobuf/issues/52).

Looking at other workarounds... maybe [this package](https://github.com/TheThingsIndustries/protoc-gen-go-json) could help?

marcelamelara:
> Looking at other workarounds... maybe [this package](https://github.com/TheThingsIndustries/protoc-gen-go-json) could help?

Thanks for the suggestion @codysoyland . My only concern with this package is their "we don't accept issues with bug reports or PRs with bug fixes" clause at the bottom of their README. I worry this package will be quite brittle compared to one that has a bigger community around it.

codysoyland:
> > Looking at other workarounds... maybe [this package](https://github.com/TheThingsIndustries/protoc-gen-go-json) could help?
> 
> Thanks for the suggestion @codysoyland . My only concern with this package is their "we don't accept issues with bug reports or PRs with bug fixes" clause at the bottom of their README. I worry this package will be quite brittle compared to one that has a bigger community around it.

Good callout, the one you posted looks promising!

marcelamelara:
I'd be a bit more comfortable with a plugin like [this one](https://github.com/searKing/golang/tree/master/tools/protoc-gen-go-tag), or the tool I shared earlier.

marcelamelara:
> Protojson should be used, be it by the end users or included in a custom marshal/unmarshal method. Given the limited number of structs in the library, I feel manually implementing the latter isn't especially onerous and avoids requiring additional tooling and dependencies.

Concerns about whether or not re-tagging the json field name in the Go bindings fully resolves the issues you are currently encountering aside, we (as the upstream) repo can certainly take on one additional tool/dependency if it means that "it will just work" for downstream. (I get that it's never quite so simple ;) )

At the same time, I don't see a problem with prescribing the use of protojson for our Go bindings, and we can offer custom marshalling functions for v1 structs as a quick workaround. 

So really, I'd like to ultimately see both changes (what you've sent here + the corresponding functions in resource_descriptor.go) _and_ the proto-gen-go changes (possibly in a different PR). 

And yes, if you have a fix for the CI, please feel free to include that in this PR.

@TomHennen @pxp928 please chime in.

gillisandrew:
@marcelamelara Yes, ideally downstream users won't have to worry about this too much. Having delved into [ProtoJSON](https://protobuf.dev/programming-guides/json/) a bit further it seems the unmarshal may be too permissive here given that both proto field names and lowerCamelCase names are accepted. Overly permissive given that by the time the statement is unmarshaled the original field name isn't accessible for validation meaning invalid statements (i.e. ones using `type` instead of `_type`) are passing.

marcelamelara:
> Having delved into [ProtoJSON](https://protobuf.dev/programming-guides/json/) a bit further it seems the unmarshal may be too permissive here given that both proto field names and lowerCamelCase (or json_name) names are accepted. Overly permissive given that by the time the statement is unmarshaled the original field name isn't accessible for validation meaning invalid statements (i.e. ones using type instead of _type) are passing.

@gillisandrew Interesting. So what you're saying is that protojson unmarshalling only behaves how we need it to if we have marshalled the struct with protojson as well. Or is there more to it? I'm thinking we may want to document a warning about this at the very least.

gillisandrew:
@marcelamelara Sorry I could have explained that a little more clearly. The issue with using protojson to unmarshal (json -> struct) before validation is that it will [accept either the JSON name (lowerCamelCase of field name unless json_name is specified) or field name (matching proto field name)](https://github.com/protocolbuffers/protobuf-go/blob/30f628eeb303f2c29be7a381bf78aa3e3aabd317/encoding/protojson/decode.go#L189-L193), but [not both
](https://github.com/protocolbuffers/protobuf-go/blob/master/encoding/protojson/decode.go#L214-L216)

That behavior is specified in the [standard](https://protobuf.dev/programming-guides/json/):
> Parsers accept both the lowerCamelCase name (or the one specified by the json_name option) and the original proto field name.

This is an issue because the in-toto spec doesn't accept both (yet the generated libs do). Suppose a user parses an validates a struct with the following:

```golang
func parseStatementJSON(jsonText []byte) (*v1.Statement, error) {
	s := &v1.Statement{}
	// Step 1: Unmarshal
	err := protojson.Unmarshal(jsonText, s)
	if err != nil {
		return nil, fmt.Errorf("failed to unmarshal statement: %w", err)
	}
	// Step 2: Validate
	err = s.Validate()
	if err != nil {
		return nil, fmt.Errorf("statement validation failed: %w", err)
	}
	// Step 3: Profit???
	return s, nil
}
```

They would (rightfully) expect that the statement satisfies the statement layer spec, but since protojson is used invalid statements that use proto field names pass validation e.g.

```json
{
  "type": "https://in-toto.io/Statement/v1",
  "subject": [
    {
      "name": "sub name",
      "digest": {
        "sha256": "2ce2e480e3c3f7ca0af83418d3ebaeedacee135dbac94bd946d7d84edabcdb64"
      }
    }
  ],
  "predicate_type": "https://example.com/unknownPred2",
  "predicate": {
    "foo": {
      "bar": "baz"
    }
  }
}
```
Because `s.Validate()` doesn't have access to the original json key name, there is no way to validate it at that point.

Further, since protojson doesn't allow both the JSON name and the field name, it means any proto field name is essentially "reserved".

Perhaps a good solution, killing two birds with one stone, is to provide a method to validate and unmarshal the json. Reflection can identify field name that don't match the JSON name and return an error when one is encountered. Or, going full circle, updating the json struct tags to match the spec and using encoding/json to handle the unmarshal (ensuring the cases it can't handle aren't included in protos).

marcelamelara:
@gillisandrew I appreciate the more detailed explanation.

> Perhaps a good solution, killing two birds with one stone, is to provide a method to validate and unmarshal the json. Reflection can identify field name that don't match the JSON name and return an error when one is encountered. Or, going full circle, updating the json struct tags to match the spec and using encoding/json to handle the unmarshal (ensuring the cases it can't handle aren't included in protos).

I will summarize what I think you're suggesting here: 
0) Provide unmarshal functions that use `encoding/json` (rather than the overly permissive protojson). 
1) Add a `ValidateJson()` method for Statements and RDs that handles the unmarshalling and checks that the schema is followed.
2) Generate the language bindings with the right json struct tags.

For 1), I think it's not unreasonable to add this functionality to the existing `Validate()` methods, but these suggestions make sense to me otherwise.

marcelamelara:
Hi @gillisandrew , friendly ping. Is there anything we can do to help move this PR forward?

gillisandrew:
@marcelamelara Sorry for the long delay, had my attention drawn to other subjects.

To me the fundamental issue is that the aim of protobufs (with protojson) isn't in alignment with the aim of defining the schema for JSON documents. It makes the assumption that it's messages will be consumed by protojson, which isn't an assumption we can make here and that has already led to bugs. 

The solution in my view is to re-evaluate protobufs as the authoritative schema and perhaps move to a better suited alternative such as JSON Schema that has broader support and better reflects the heterogeneous toolchain these documents will be consumed with. See https://github.com/ocsf/ocsf-schema for example.

marcelamelara:
@gillisandrew Thanks for your reply, no worries about the delay. Given that we won't find a resolution in this PR and the downstream issue is currently patched, I think we should close this PR. Let's move the discussion about protobufs vs. schema to an issue.

PR #408, opened 2024-11-16
fix discrepencies in v0.2 example

lumjjb:
As part of fixing #391 following up #345, removing invocation since it is not part of the spec in the example


marcelamelara (approved):
LGTM! Thanks for sending this @lumjjb !

marcelamelara:
I noticed separately that there are some slight discrepancies between how the spec names certain fields and how they are defined in the .proto. For example, the spec defines a `metadata` field, which in the proto is defined as `ScanMetadata`. Does this pose a problem?

PR #417, opened 2024-11-29
add slsa provenance predicate v0.2 protobuf

kpauljoseph:
This PR adds a proto definition for the slsa v0.2 predicate type as part of the ongoing deprecation efforts to move all the dependencies from in-toto/in-toto-golang to in-toto/attestation.

Followed the spec provided at https://slsa.dev/spec/v0.2/provenance


marcelamelara (changes requested):
Thanks for sending this @kpauljoseph !  The proto looks great, though I do have one requested change.

Our convention is typically to track major versions of predicates, i.e. to store them in `/v0`, `/v1` etc. directories. Because this is the only pre-v1 version of Provenance that we support, I suggest renaming this directory structure (as well as the language binding packages) to `v0`.

adityasaky:
@marcelamelara should there be an exception for v0? v1, v2 etc make sense due to semver as changes are backwards compatible but we may well have very different protos for v0.1 and v0.2? I suggest allowing minor versions for v0 so that we don't have to rename a bunch for consistency later?

marcelamelara:
> should there be an exception for v0? v1, v2 etc make sense due to semver as changes are backwards compatible but we may well have very different protos for v0.1 and v0.2? I suggest allowing minor versions for v0 so that we don't have to rename a bunch for consistency later?

@adityasaky good question. We've kind of been deciding this on a case by case basis for v0.x predicates. If we were to pull in the v0.1 provenance predicate (which I don't think is the intent) and they were vastly different, I'd say that we can distinguish between v0.1 and v0.2. We do this for the vulns predicate, for example. But for other predicates, we've bult the breaking changes into the same proto. I think for me the answer ultimately depends on how widely used the different v0.x predicates are.

kpauljoseph:
@marcelamelara I found repos like guacsec/guac importing "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1" and "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"

Are we supposed to rename the language bindings for v0.2 proto to the following?

```
option go_package = "github.com/in-toto/attestation/go/predicates/provenance/v0";
option java_package = "io.github.intoto.attestation.predicates.provenance.v0";
```

marcelamelara (approved):
LGTM ! Thanks again @kpauljoseph !

pxp928:
> Thanks for doing this search. If GUAC is distinguishing these, then there are probably other systems that do this as well.

Yes, GUAC does use both import"github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1" and "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2" as we want to be able to support ingestion of different versions of the spec.

> At some point, we probably also need to go to folks like GUAC and recommend switching over to the language bindings in this repo.

I created an issue to switch it over: https://github.com/guacsec/guac/issues/2450

pxp928 (approved):
LGTM

PR #429, opened 2025-01-16
vsa: Make verifiedLevels field repeated to match spec

### LINKED ISSUE #425: VSA's verified levels field should be an array rather than a string
https://github.com/in-toto/attestation/blob/main/protos/in_toto_attestation/predicates/vsa/v1/vsa.proto#L36 has `verifiedLevels` as a string rather than an array. The spec for the VSA (https://slsa.dev/spec/v1.0/verification_summary#verifiedLevels) requires this to be an array, to support indicating the verified levels for different tracks.

adityasaky:
The specification requires the verifiedLevels field to be a list of string values rather than a single string.

Fixes #425


marcelamelara (approved):
Thanks for this fix @adityasaky ! LGTM

pxp928 (approved):

PR #431, opened 2025-01-17
Ensure the Envelope spec meets ITE-5

### LINKED ISSUE #21: Clarify that any envelope can be used

marcelamelara:
Between some past changes and missing content, the Envelope spec isn't fully compliant with [ITE-5](https://github.com/in-toto/ITE/tree/master/ITE/5). This PR fixes that.

Fixes #21

### DIFF
--- a/spec/v1/envelope.md
+++ b/spec/v1/envelope.md
@@ -1,18 +1,40 @@
 # Envelope layer specification
 
-Version: [DSSE v1.0]
-
-The Envelope is the outermost layer of the attestation, handling
-authentication and serialization.
+The Envelope is the outermost layer of the attestation, handling  serialization
+and authentication (via digital signatures).
 
 ## Schema
 
-The format and protocol are defined per [DSSE v1.0].
+The RECOMMENDED format and protocol for Envelopes are defined per [DSSE v1.0].
+Producers MAY use other signature methods and formats that meet the [ITE-5]
+specification:
+
+-   MUST support the inclusion of multiple signatures in a single envelope
+-   SHOULD include an authenticated payload type
+-   SHOULD avoid depending on canonicalization for security
+-   SHOULD support a hint indicating what signing key was used, i.e., a KEYID
+-   SHOULD NOT require the verifier to parse the payload before verifying
+-   SHOULD NOT require the inclusion of signing key algorithms in the signature
+
+### Alternative Envelope schemas
+
+-   The [Sigstore Bundle], while [supporting DSSE], is not currently [ITE-5]
+    compliant because it requires a _single signature_ in the envelope.[^1]
+-   The [COSE\_Sign] structure is [ITE-5] compliant, whereas the `COSE\_Sign1`
+    format that supports only one signer is NOT compliant.
 
 ## Fields
 
-The in-toto Attestation Framework has the following requirements for the
-standard DSSE fields.
+The in-toto Attestation Framework has the following general field requirements
+for an Envelope:
+
+-   `signature` (or equivalent) is REQUIRED and MUST be defined as an array.
+-   A `keyid` (or equivalent) SHOULD be included for each signing key used.
+-   `payload` (or equivalent) SHOULD be included and contain the attestation
+    data that was signed.
+
+In addition, the Envelope spec has the following specific requirements for the
+standard [DSSE][DSSE v1.0] fields.
 
 -   `payloadType` MUST be set to `application/vnd.in-toto+json`, which
     indicates that the Envelope contains a JSON object with a `_type` field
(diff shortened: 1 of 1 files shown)


spec/v1/envelope.md
```diff
@@ -57,10 +80,16 @@ Example media types for single DSSE-signed attestation predicates include:
 -   SPDX: `application/vnd.in-toto.spdx+dsse`
 -   VSA: `application/vnd.in-toto.vsa+dsse`
 
+[^1]: There is an [ongoing discussion](https://github.com/sigstore/sig-clients/issues/9) about supporting [DSSE Signature Extensions](https://github.com/secure-systems-lab/dsse/blob/devel/envelope.md#signature-extensions-experimental) to extend the current features of Sigstore Bundles.
+
 [Bundle]: bundle.md
-[DSSE v1.0]: https://github.com/secure-systems-lab/dsse/blob/v1.0.0/envelope.md
+[COSE\_Sign]: https://datatracker.ietf.org/doc/html/rfc8152#section-4.1
```
trishankatdatadog, line 86: Is the `\` intended?

trishankatdatadog (approved):

spec/v1/envelope.md
marcelamelara: I added the `\` because I was seeing some markdown rendering weirdness without it. But there is another place where I need to remove it :)

marcelamelara: turns out the `\` aren't needed after all. I removed them.

spec/v1/envelope.md
```diff
@@ -1,18 +1,41 @@
...
+The in-toto Attestation Framework has the following general field requirements
+for an Envelope:
+
+-   `signature` (or equivalent) is REQUIRED and MUST be defined as an array.
+-   A `keyid` (or equivalent) SHOULD be included for each signing key used.
+-   `payload` (or equivalent) SHOULD be included and contain the attestation
+    data that was signed.
+-   `payloadType` (or equivalent) SHOULD be signed along with the `payload`.
+
+In addition, the Envelope spec has the following specific requirements for the
```
adityasaky, line 37: should this say more about how payloadType is structured for other envelopes? i.e., that it communicates this is an in-toto payload and the serialized format?

marcelamelara: Good point, we probably should. 

And maybe this is stricter than ITE-5, but I've also been wondering if that SHOULD should be a MUST? As in: "`payloadType` (or equivalent) MUST be signed along with the `payload`"

trishankatdatadog: Are we blocking only on this?

TomHennen (approved):
Sorry I missed this request for review  :(

spec/v1/envelope.md
marcelamelara: Yes, sorry for the delay on this.

trishankatdatadog: No worries! I asked only because I wanted to know whether either or both of you needed help with this :)

marcelamelara: alright, PTAL https://github.com/in-toto/attestation/pull/431/commits/4c0ae5eedd40a413ed886ecc5189174b57365cc2 . 
@adityasaky @trishankatdatadog

adityasaky: lgtm

adityasaky (approved):

spec/v1/envelope.md
```diff
@@ -1,23 +1,48 @@
...
 
 ## Fields
 
-The in-toto Attestation Framework has the following requirements for the
-standard DSSE fields.
+The in-toto Attestation Framework has the following general field requirements
+for an Envelope:
+
+-   `signature` (or equivalent) is REQUIRED and MUST be defined as an array.
+-   A `keyid` (or equivalent) SHOULD be included for each signing key used.
```
trishankatdatadog, line 32: It's also not obvious that a `keyid` is under a signature. Furthermore, it's missing the `sig` bit. Should we just point to the DSSE spec instead?

trishankatdatadog (changes requested):

spec/v1/envelope.md
marcelamelara: I see what you mean. I think we specifically wanted to avoid prescribing DSSE here, so let me see what I can do.

marcelamelara: I added some more text to clarify expectations a bit more. PTAL.

marcelamelara:
@trishankkarthik Can you please re-approve?

spec/v1/envelope.md
```diff
@@ -1,18 +1,46 @@
 # Envelope layer specification
 
-Version: [DSSE v1.0]
-
-The Envelope is the outermost layer of the attestation, handling
-authentication and serialization.
+The Envelope is the outermost layer of the attestation, handling  serialization
```
marcelamelara:
```suggestion
The Envelope is the outermost layer of the attestation, handling serialization
```

PR #434, opened 2025-01-21
VulnsV2: Add missing invocation to protos and spec

puerco:
This PR updates the vulns02 protos to rename the `scanner.database` field (to `scanner.db`) and the `scanner_metadata` field (to just `metadata`) to match the vulns02 spec.

Sample generated predicate (with the go library):

```json
{
  "scanner": {
    "uri": "pkg:github/aquasecurity/trivy@244fd47e07d1004f0aed9",
    "version": "0.19.2",
    "db": {
      "uri": "pkg:github/aquasecurity/trivy-db/commit/4c76bb580b2736d67751410fa4ab66d2b6b9b27d",
      "version": "v1-2021080612",
      "last_update": {
        "seconds": 1737495873,
        "nanos": 773202315
      }
    },
    "result": [
      {
        "id": "CVE-123",
        "severity": [
          {
            "method": "nvd",
            "score": "medium"
          },
          {
            "method": "cvss_score",
            "score": "5.2"
          }
        ]
      }
    ]
  },
  "metadata": {
    "scan_started_on": {
      "seconds": 1737495873,
      "nanos": 773202413
    },
    "scan_finished_on": {
      "seconds": 1737495873,
      "nanos": 773202481
    }
  }
}

```

Signed-off-by: Adolfo García Veytia (Puerco) <adolfo.garcia@uservers.net>

### DIFF
--- a/protos/in_toto_attestation/predicates/vulns/v0.2/vulns.proto
+++ b/protos/in_toto_attestation/predicates/vulns/v0.2/vulns.proto
@@ -11,13 +11,13 @@ option java_package = "io.github.intoto.attestation.predicates.vulns.v02";
 // Validation of all fields is left to the users of this proto.
 message Vulns {
   Scanner scanner = 1;
-  ScanMetadata scan_metadata = 2;
+  ScanMetadata metadata = 2;
 }
 
 message Scanner {
   string uri = 1;
   optional string version = 2;
-  VulnDatabase database = 3;
+  VulnDatabase db = 3;
   repeated Result result = 4;
 }
 
--- a/spec/predicates/vulns_02.md
+++ b/spec/predicates/vulns_02.md
@@ -135,7 +135,7 @@ The `predicate` contains a JSON-encoded data with the following fields:
          "id": "CVE-123",
          "severity": [
           { "method": "nvd", "score": "medium"},
-          { "method": "cvss_score", "score", "5.2" }
+          { "method": "cvss_score", "score": "5.2" }
          ]
         },
         {...}


puerco:
I've pushed another commit renaming `scanner.db` to `scanner.database` in the documentation to match the code. I wonder which one was wrong?

puerco:
Another problem:
The example metadata shows the metadata field name as `metadata`:

```json
    "metadata": {
      "scanStartedOn": "2021-08-06T17:45:50.52Z",
      "scanFinishedOn": "2021-08-06T17:50:50.52Z"
    }
```

Yet the protos have it as `scan_metadata`:

https://github.com/in-toto/attestation/blob/11ca4fce9b7a0708a460918c8f539701e7a5c456/protos/in_toto_attestation/predicates/vulns/v0.2/vulns.proto#L12-L15

Should we rename it? I would suggest changing the proto (and thus, the generated code) to avoid ugly hacks with annotations, etc. On the downside, it would be a breaking change but I think the generated libraries have not been released yet.

marcelamelara:
@puerco have you seen #408 ? I think that the `Invocation` field was actually [removed in the vulns v0.2 spec](https://github.com/in-toto/attestation/pull/345).

> I've also regenerated the go, python and java libraries.

No need for that :) We automatically regenerate these with the CI whenever the protos change.

marcelamelara:
> Should we rename it? I would suggest changing the proto (and thus, the generated code) to avoid ugly hacks with annotations, etc. On the downside, it would be a breaking change but I think the generated libraries have not been released yet.

@puerco Yes! The `metadata` field needs to be changed in the proto.

puerco:
> have you seen https://github.com/in-toto/attestation/pull/408

Grrr no! I missed that :( :(
OK, let me drop that change from this PR and I'll handle renaming the metadata field

marcelamelara:
> Grrr no! I missed that :( :(
OK, let me drop that change from this PR and I'll handle renaming the metadata field

@puerco Thanks so much, sorry for the extra work!

puerco:
No worries, I should have gone through the examples :D

I rescoped the PR to handle the `database` and `metadata` inconsistency with the vulnsv2 spec (also fixed a tiny typo in the spec example). I've edited the PR body with a generated example.

marcelamelara (approved):
LGTM @puerco ! Thanks so much for sending this!

pxp928 (approved):
Thanks @puerco good catch!

Issue #361, opened 2024-05-30
FR or ITE? CDDL definition for JSON/CBOR plus bindings in Go/Python


deeglaze:
The protobuf representation for in-toto attestations isn't as portable as standards bodies like the IETF would like to depend on. RFC8610 defines a concise data description language that specifically has the intention of unifying JSON (RFC8259) and CBOR (RFC8949). Given that JSON is not as compact and bandwidth-friendly, I think we should expand the in-toto information model to have CBOR encoding and COSE_Sign1 (RFC9052) signing envelopes. This should be in line with ITE-5, just subject to a new content media type `application/vnd.in-toto+cose` for the signed attestation and `application/vnd.in-toto+cbor` for the unsigned attestation for example.

By incorporating CBOR, in-toto attestations can be more easily be included in CoRIM-based attestation verifiers like the Veraison project.

The biggest task is deciding on key indices for maps where previously there were textual names, though my recommendation is to assign from 0 in alphabetical order for the current version of the schema. I'm not sure if that assignment counts as needing an ITE or if an FR suffices.

PR #452, opened 2025-03-25
ITE-12: translate data definitions to CDDL

deeglaze:
As a means to improve the conciseness of in-toto, we can dually express objects in both JSON and CBOR using a standard data definition language called CDDL. This is complementary to https://github.com/in-toto/ITE/pull/60


marcelamelara:
closing due to inactivity. This is being tracked in #361 so we can reopen at a later time.

PR #464, opened 2025-05-19
Support validation of Cosign OCI statements with type https://in-toto.io/Statement/v0.1

### LINKED ISSUE #461: Support Validation of Cosign OCI Statements with Type https://in-toto.io/Statement/v0.1
Currently, the `Validate` method in the Statement implementation enforces that the statement type must have a prefix of https://in-toto.io/Statement/ and specifically checks for version v1. However, Cosign OCI signatures use statements with the type https://in-toto.io/Statement/v0.1. This causes validation to fail for Cosign OCI statements, even though they are valid in-toto statements.

To address this, the `Validate` method should be updated to allow statements with type https://in-toto.io/Statement/v0.1 in addition to https://in-toto.io/Statement/v1. This change will ensure compatibility with Cosign OCI signatures while maintaining the integrity of the validation process.

alenon:
Support validation of Cosign OCI statements with type https://in-toto.io/Statement/v0.1

This change updates the  method to allow statements with type v0.1 in addition to v1. This ensures compatibility with Cosign OCI signatures.

Fixes #461

### DIFF
--- a/go/v1/statement.go
+++ b/go/v1/statement.go
@@ -4,9 +4,13 @@ Wrapper APIs for in-toto attestation Statement layer protos.
 
 package v1
 
-import "errors"
+import (
+	"errors"
+)
 
-const StatementTypeUri = "https://in-toto.io/Statement/v1"
+const statementTypeUriPrefix = "https://in-toto.io/Statement/"
+const statementTypeUriLegacy = statementTypeUriPrefix + "v0.1"
+const StatementTypeUri = statementTypeUriPrefix + "v1"
 
 var (
 	ErrInvalidStatementType  = errors.New("wrong statement type")
@@ -17,7 +21,7 @@ var (
 )
 
 func (s *Statement) Validate() error {
-	if s.GetType() != StatementTypeUri {
+	if !s.isValidType() {
 		return ErrInvalidStatementType
 	}
 
@@ -47,4 +51,8 @@ func (s *Statement) Validate() error {
 	}
 
 	return nil
+}
+
+func (s *Statement) isValidType() bool {
+	return s.GetType() == StatementTypeUri || s.GetType() == statementTypeUriLegacy
 }
--- a/go/v1/statement_test.go
+++ b/go/v1/statement_test.go
@@ -52,7 +52,7 @@ func TestJsonUnmarshalStatement(t *testing.T) {
 }
 
 func TestBadStatementType(t *testing.T) {
-	var badStType = `{"_type":"https://in-toto.io/Statement/v0","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}`
+	var badStType = `{"_type":"https://not-in-toto.io/Statement/v1","subject":[{"name":"theSub","digest":{"alg1":"abc123"}}],"predicateType":"thePredicate","predicate":{"keyObj":{"subKey":"subVal"}}}`
 
 	got := &Statement{}
 	err := protojson.Unmarshal([]byte(badStType), got)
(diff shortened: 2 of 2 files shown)


go/v1/statement.go
```diff
@@ -17,7 +21,7 @@ var (
 )
 
 func (s *Statement) Validate() error {
-	if s.GetType() != StatementTypeUri {
+	if !strings.HasPrefix(s.GetType(), StatementTypeUriPrefix) {
```
marcelamelara: I think this is actually a little too permissive. I would rather check the two specific supported versions.

marcelamelara (changes requested):
Thanks for sending this @alenon , apologies for the very long delay. I think these changes will be helpful for backwards compatibility, but don't want to lose the specificity of checking the supported versions.

marcelamelara:
@in-toto/attestation-maintainers can we please get a second review here?

go/v1/statement.go
alenon: Thank you @marcelamelara for you review. I've double checked it, and it seems like the only two options are indeed v1 and v0.1 (nothing in between), which means I can do the strict check. Will soon update the code.

go/v1/statement.go
```diff
@@ -48,3 +55,13 @@ func (s *Statement) Validate() error {
 
 	return nil
 }
+
+func isValidType(t string) bool {
```
trishankatdatadog: This function could be more concise, but it works.

go/v1/statement.go
```diff
@@ -17,7 +24,7 @@ var (
 )
 
 func (s *Statement) Validate() error {
-	if s.GetType() != StatementTypeUri {
+	if !isValidType(s.GetType()) {
```
trishankatdatadog: Could we make `isValidType()` a function of Statement itself?

trishankatdatadog (commented):
Thanks for your much-needed PR! Left some minor comments...

go/v1/statement.go
alenon: Sure. Done!

go/v1/statement.go
```diff
@@ -17,7 +24,7 @@ var (
 )
 
 func (s *Statement) Validate() error {
-	if s.GetType() != StatementTypeUri {
+	if !s.isValidType(s.GetType()) {
```
trishankatdatadog: Almost there: couldn't we avoid passing the type as a public parameter at all now that the function has access to this internal field anyway?

alenon: Thank you for paying an attention, I do completely agree with you! Done!

go/v1/statement.go
```diff
@@ -48,3 +55,8 @@ func (s *Statement) Validate() error {
 
 	return nil
 }
+
+func (s *Statement) isValidType() bool {
+	return strings.HasPrefix(s.GetType(), StatementTypeUriPrefix) &&
```
trishankatdatadog: On second thought, could we make this check tighter by making it a regex? Right now, it could actually allow arbitrary strings between the prefix and two suffixes...

trishankatdatadog: Well, not a regex, but just checking for one of either two strings exactly.

alenon: @trishankatdatadog 

Another idea I had was to make it optional to validate the legacy type or not.
`Validate(opts ...func(s *Statement))`

if there is a `WithCustomTypeValidator` or `WithLegacyTypeValidator`
-> use it instead of `isValidType()`

trishankatdatadog: Hmm, interesting. Up to you whether you want to implement it now or not, but this is a great start!

trishankatdatadog (approved):

PR #470, opened 2025-07-03
Create a simple verification attestation

arewm:
While a VSA exists within the SLSA specification, the current tracks have grown past the the initially proposed spec. We are proposing to create a new attestation predicate to replace the VSA which is also simpler to consume.

There is intentionally less information in this attestation. This was partially done to create smaller predicates to enable faster verification as well as removing informational fields. Even though policy information is not present in the attestation itself, we call out the ability to use a SLSA Provenance attestation to provide more information about how the policy was evaluated. This provenance MAY be used to qualify trust in the simple verification attestation if needed.

resolves: #277


spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,108 @@
...
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/simple-verification/v0.1",
+  "predicate": {
+    "verifier": {
+      "id": "<URI>"
+    },
+    "timeVerified": "<TIMESTAMP>",
+    "resourceUri": "<artifact-URI>",
+    "properties": ["<SCOPED_VERIFIED_PROPERTY>", ...]
```
jkjell:
```suggestion
    "properties": ["<VERIFIED_PROPERTY>", ...]
```

I didn't see anywhere else that used the term `SCOPED`. I don't think that dropping that loses any intent.

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Purpose
+
+A Simple Verification Attestation (SVA) communicates that an artifact has been verified against one or more policies, and records the properties that were verified. This enables consumers to trust a concise summary of verification results, without requiring access to all underlying attestations or policies.
+
+## Use Cases
+
+-   Allowing software consumers to delegate complex policy decisions to a trusted verifier, and rely on a simple summary of which properties were verified.
+-   Supporting evolving verification needs (e.g., SLSA, vulnerability scanning, organizational policies) without coupling the SVA format to any specific framework.
+-   Enabling verifiers to add additional information (e.g., policy references, expiration, property-level timing) as needed, while maintaining a simple core schema.
```
jkjell: Should we add optional fields for a policy and an array input_attestations as resource descriptors? This could ensure that verifiers who may add that information, do so in a consistent way.

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,108 @@
...
+
+Type URI: https://in-toto.io/attestation/simple-verification/v0.1
+
+Version: 0.1
+
+Authors: Tom Hennen (@TomHennen), Andrew McNamara (@arewm)
+
+## Purpose
+
+A Simple Verification Attestation (SVA) communicates that an artifact has been verified against one or more policies, and records the properties that were verified. This enables consumers to trust a concise summary of verification results, without requiring access to all underlying attestations or policies.
```
jkjell: > communicates that an artifact has been verified against one or more policies

Is there a specific use case for multiple policies in one SVA? 

As a current implementor of policy verification, we would only ever attest to a single policy verification at a time. This would allow a simple conveying of policy used, input attestations, and verified properties. If we considered multiple policies in a single SVA, I'd feel tempted to use some sort of hierarchy (complexity) to show a relationship between the policy verified and other metadata.

Perhaps the simplest question for multiple policies, what does `timeVerified` mean if multiple policies are allowed?

spec/predicates/vsa.md
```diff
@@ -2,6 +2,8 @@
 
 Type URI: https://slsa.dev/verification_summary/v1
 
+Deprecated Type URI: https://in-toto.io/attestation/simple-verification/v0.1
```
jkjell: I don't understand this addition here. 😅

jkjell (commented):
Thanks for (re)starting this conversation! I was tempted to suggest we call this a `sVSA`. 😂

spec/predicates/vsa.md
arewm: I was trying to indicate that the VSA predicate is deprecated in favor of the simple-verification one... but I agree, this change doesn't convey that.

spec/predicates/simple-verification.md
arewm: I wanted to call out that verified properties should be scoped to the entity that the property belongs to. For example, anything prefixed with SLSA would be defined by the SLSA spec. If a policy engine has pre-computed policy rules, then it can claim these as verified but it should scope those to its own namespace (i.e. `CONFORMA_*` or `AMPEL_*`).

To your point, however, it would likely be better to only put `VERIFIED_PROPERTY` here and define the need to scope those properties elsewhere.

spec/predicates/simple-verification.md
arewm: I think that the desired pattern for this would be to have that information included as a provenance attestation for this attestation.

spec/predicates/simple-verification.md
arewm: I don't have a use case off hand, but it seems reasonably possible that a policy engine could consume multiple policies when making a claim. Any hierarchy determination would be handled by the evaluating engine.

TomHennen: I think this all comes down to how people are defining 'policy'.

In _my_ usage, a policy is the set of rules that get evaluated for any given verification.  In that sense 'multiple policies' just become the union of all the rules.

Different policy engines might implement things differently.  One engine (the sort I'm the most used to) has a single file that defines all the rules a given package needs to meet.  Other engines, however, might define multiple different files to achieve the same effect.  Still, for an artifact to be 'verified' it needs to pass all the rules provided.

> Perhaps the simplest question for multiple policies, what does timeVerified mean if multiple policies are allowed?

The time when all the rules across all the policies were evaluated?

I don't think we need to (or should) conflate policy with a single evaluation.

In either case, does it matter much for the purposes of this predicate definition?  Could those details be left to the implementor?

spec/predicates/simple-verification.md
TomHennen: We've previously seen folks get confused about those fields and were incorrectly trying to validate them when verifying the VSA.

The thought we were discussing last week (and previously IIRC) is that for the primary purposes of a VSA that information just isn't needed and, as Andrew suggests, if it is needed it can be provided in a separate attestation.

_Additionally_, some users are space constrained when passing attestations.  Including those other fields could cause those users to unnecessarily exceed their space constraints.

jkjell: I like the idea of additional, related attestations.

spec/predicates/simple-verification.md
jkjell:
```suggestion
A Simple Verification Attestation (SVA) communicates that an artifact has been evaluated against one or more policies, and records the properties that were verified, at the point-in-time of the evaluation. This enables consumers to trust a concise summary of verification results, without requiring access to all underlying attestations or policies.
```
I'm convinced by all your explanations. I think the proposed changes above capture what you've said, as I understand them. You don't necessarily need to include the updates but, I like how you laid out the intent in your comments above.

jkjell (approved):
I can't seem to resolve the conversations around my comments. I like the simplicity of this!

TomHennen:
Just wanted to make sure [this comment](https://github.com/in-toto/attestation/issues/277#issuecomment-3046852408) was seen and at least considered.  CC @AdamZWu

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,108 @@
...
+  // Standard attestation fields:
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [{ ... }],
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/simple-verification/v0.1",
+  "predicate": {
+    "verifier": {
+      "id": "<URI>"
+    },
```
Hayden-IO: VSAs supported additional verifier metadata with verifier versions. Was this found to not be necessary, or if we wanted to include such metadata, would we do so as a property now?

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,108 @@
...
+
+> Timestamp indicating what time the verification occurred.
+
+**`resourceUri`, required** string (ResourceURI)
+
+> URI that identifies the resource associated with the artifact being verified.
+
+**`properties`, required** array of string
+
+> Indicates the properties verified for the artifact.
```
Hayden-IO: Do you have any prescriptive guidance on how to name properties, or is this entirely up to the SVA producer?

spec/predicates/simple-verification.md
arewm: @jkjell , I represented this pattern of additional attestations in the model section. Do you think that I need to add more or can I just resolve this thread?

spec/predicates/simple-verification.md
arewm: Those suggestions seem reasonable, thanks.

spec/predicates/simple-verification.md
arewm: When @TomHennen and I were discussing this, we were talking about how the additional information was often confusing. This can certainly still be captured in a provenance attestation for the SVA itself.

spec/predicates/simple-verification.md
arewm: I added some further guidance here

```
**`properties`, required** array of string

> Indicates the properties verified for the artifact. These SHOULD be scoped according
> to the framework being verified or the verifier's policy rules.
```

spec/predicates/vsa.md
arewm: @jkjell , is the change to the VSA predicate clearer now? I _want_ to deprecate the VSA partially because I don't wants to have its versioning bumped every time the SLSA spec's version changes.

@TomHennen , what are your thoughts here? Another option if we don't want to deprecate the VSA is that we can just move it wholesale into in-toto as-is to remove the arbitrary version bumps.

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,109 @@
...
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [{ ... }],
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/simple-verification/v0.1",
+  "predicate": {
+    "verifier": {
+      "id": "<URI>"
+    },
+    "timeVerified": "<TIMESTAMP>",
```
AdamZWu: Just a nit, but could "timeCreated" be a better name?

Since the checks that produce the properties are not pre-determined, different properties can be produced based on quite diverse processes. For example, some properties may be "inherited" from verification that happened a long time ago; some others may not be a result of a single verification; etc.

Hence "verified" could be mis-interpreted by consumers.

I think this attribute should only designate the time that a particular SVA is created.

spec/predicates/simple-verification.md
jkjell: Yes, please resolve.

spec/predicates/vsa.md
jkjell: The updated wording makes sense to me. 👍

spec/predicates/simple-verification.md
arewm: I like that clarity. Thanks for the suggestion.

spec/predicates/simple-verification.md
arewm: @haydentherapper , do you think that we should provide more information on this in the predicate specification?

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,109 @@
...
+  "subject": [{ ... }],
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/simple-verification/v0.1",
+  "predicate": {
+    "verifier": {
+      "id": "<URI>"
+    },
+    "timeCreated": "<TIMESTAMP>",
+    "resourceUri": "<artifact-URI>",
```
puerco: Perhaps I'm missing something, but isn't `resourceUri` essentially the attestation subject?  Also, should this field be a resource descriptor?

> An SVA is an attestation that a verifier (`verifier.id`) has checked an artifact (`resourceUri`)

spec/predicates/simple-verification.md
Hayden-IO: Thinking more, does a consumer of a SVA need to know that a verifier has changed its verification logic? These verifiers aren't expected to be auditable, so a new revision of a verifier isn't meaningful, unless we expect that verifiers will publish the list of properties that they may include in a SVA.

Another option is to ask that verifiers version their identifiers, as in if they update their verification logic, they use a new unique ID. This puts additional work on the SVA consumer however, to maintain an up to date list of verifier IDs.

AdamZWu: +1 on versioning the identifier.

spec/predicates/simple-verification.md
AdamZWu: > Perhaps I'm missing something, but isn't `resourceUri` essentially the attestation subject? Also, should this field be a resource descriptor?
> 
> > An SVA is an attestation that a verifier (`verifier.id`) has checked an artifact (`resourceUri`)

IMO they are orthogonal.

The subject records info relevant for the artifact being verified; the "resourceUri", OTOH, identifies the protection domain the artifact is verified against.

For example,
- The subject may have a uri "sftp://akamai.net/foo/bar.zip" -- that is where it is retrieved from;
- The SVA resourceUri may be "https://foo.com/app_bar" -- that is the service the artifact is going to be deployed onto (hence it needs to be verified against a policy for this URI).

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,109 @@
...
+
+Type URI: https://in-toto.io/attestation/simple-verification/v0.1
+
+Version: 0.1
+
+Authors: Tom Hennen (@TomHennen), Andrew McNamara (@arewm)
+
+## Purpose
+
+A Simple Verification Attestation (SVA) communicates that an artifact has been evaluated against one or more policies, and records the properties that were verified, at the point-in-time of the evaluation. This enables consumers to trust a concise summary of verification results, without requiring access to all underlying attestations or policies.
```
marcelamelara: We can probably be a bit more precise here
```suggestion
A Simple Verification Attestation (SVA) communicates that an artifact has been evaluated against one or more policies, and records the properties that were verified, at the point-in-time of the evaluation. This enables consumers to establish trust in a concise summary of verification results coming from a trusted verifier, without requiring access to all underlying attestations or policies.
```

marcelamelara:
@arewm Can you please line wrap at 80 chars? Just makes it a bit easier to comment on specific pieces.

spec/predicates/simple-verification.md
marcelamelara: I agree that this field may not necessarily be the Statement subject, but I also think that turning this into a `ResourceDescriptor` type makes sense. It facilitates describing the verified artifact in more detail, if need be, in a standard format.

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,109 @@
...
+{
+  // Standard attestation fields:
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [{ ... }],
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/simple-verification/v0.1",
+  "predicate": {
+    "verifier": {
+      "id": "<URI>"
```
marcelamelara: I'm assuming this is defined as an object with a single field so that verifier information can be extended. Is there any reason the `verifier` couldn't just be defined as a `ResourceDescriptor`?

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,109 @@
...
+
+> URI indicating the verifier's identity.
+
+**`timeCreated`, required** string (Timestamp)
+
+> Timestamp indicating what time the verification occurred.
+
+**`resourceUri`, required** string (ResourceURI)
+
+> URI that identifies the resource associated with the artifact being verified.
```
marcelamelara: Is there a reason this artifact doesn't need to be immutable? I think the spec should at least RECOMMEND describing an immutable artifact as that removes ambiguity about which artifact was verified. This may be another reason to switch to a `ResourceDescriptor` field type. 

I also suggest picking a less generic field name, like `verifiedArtifact`, so that it's unambiguous what this field represents.

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,109 @@
...
+
+**`properties`, required** array of string
+
+> Indicates the properties verified for the artifact. These SHOULD be scoped according
+> to the framework being verified or the verifier's policy rules.
```
marcelamelara: Is this a list of properties that were `verified` regardless of pass/fail status, or only the properties that passed validation?

marcelamelara (changes requested):
Thanks for sending this @arewm and @TomHennen. Apologies for the delayed review. I've got several comments about making the predicate spec a bit crisper and clearer, but in general really like where this is heading.

I did make a comment at a SLSA meeting a while back about whether there'd be potential confusion between SVA <> VSA, but it seems like this may not be a widespread concern. One potential way to address this is to call the predicate an SV predicate only or change the name to something like SVR (where R stands for Result) since I think it's implied that the predicate would be part of an attestation.

spec/predicates/simple-verification.md
```diff
@@ -0,0 +1,109 @@
...
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/attestation/simple-verification/v0.1",
+  "predicate": {
+    "verifier": {
+      "id": "<URI>"
+    },
+    "timeCreated": "<TIMESTAMP>",
+    "resourceUri": "<artifact-URI>",
+    "properties": ["<VERIFIED_PROPERTY>", ...]
```
trishankkarthik: While I understand the simplicity of stringly-typed properties, I think they're a bit too simple to be generalised.

Personally, I think it's more useful to point to a policy along with an attestation bundle as supporting evidence. A single policy can be used to succinctly summarise a bunch of properties at once. As usual, policies and attestations can be identified with ResourceDescriptors.

More importantly, I'm curious to know what the group's thinking here is about what an SVA should say with respect to policies. Should an SVA always be restricted to one policy? What about multiple policies? Furthermore, what about explicitly saying that an artifact _failed_ to meet a certain policy at a certain time? These are things we found useful in practice to answer at Datadog, and I'd like to know what others think.

trishankatdatadog:
> @arewm Friendly ping. Please let us know if our feedback doesn't make sense.

Yes, please, @TomHennen, too. My hot take is that SVAs MUST be [reproducible](https://docs.google.com/document/d/1Tr6jCtTcTB8KpDMp2I1PFuxP7A2nQa5BeuwOy05iuMw/edit?pli=1&tab=t.9h3hhecwdwym): that is, given exactly the same Artifact and Attestation Bundle as well as "semantically equivalent" Policies (defined in the Framework), any two Policy Verifiers MUST output essentially the same SVA.

spec/predicates/simple-verification.md
arewm: Only properties that passed validation.

spec/predicates/simple-verification.md
arewm: I wasn't intending for it to be mutable. I agree that it makes sense to be immutable.

spec/predicates/simple-verification.md
arewm: Policy rules would be an input to some evaluation. The output of these evaluations would be that some rules pass while others fail. The implementation of how those evaluations are run are ultimately irrelevant -- i.e. whether one or multiple policies are consumed. If you have two different policy engines, each capable of defining their own policy rules (i.e. AMPEL and Conforma), then the output of those _passed_ rules can be collected together with policy-engine scoped properties. 

The reproducibility of these results can be achieved by leveraging the provenance of the SVA.

> Furthermore, what about explicitly saying that an artifact failed to meet a certain policy at a certain time?

I posit that if you have a policy rule which can pass one day and fail another then it doesn't belong as a verified property. It is not a permanent property and therefore shouldn't be considered as part of a permanent attestation.

> what about explicitly saying that an artifact failed to meet a certain policy at a certain time?

I wanted to keep it simple by only having the positive case. If a policy rule not being met needs to be used by some verifier then that might need to be its own positive policy rule.

AdamZWu: > I posit that if you have a policy rule which can pass one day and fail another then it doesn't belong as a verified property. It is not a permanent property and therefore shouldn't be considered as part of a permanent attestation.

I slightly disagree. If we consider time as an input to the verification, I think there are several layers / classes of determinism, from weakest to strongest:

1. Non-deterministic: given the exact same input (incl. time), the verifier can produce different verdicts.
    - I don't think such verifiers are considered acceptable. :P
2. Deterministic but not monotonic: the exact same input (incl. time) will produce the same verdict, however as the time (as the sole variable input) progresses, the verdict can flip-flop.
    - I think this layer / class is not good to be stored in SVA or VSA.
3. Deterministic and monotonic: the exact same input (incl. time) will produce the same verdict, and the verdict can make a one-way flip along the time axis.
    - I feel / wish the SVA / VSA is better support this layer / class. It probably reflects the most common usage for time-sensitive verification in the SSC (artifact freshness, vulnerability check, etc.)
4. Deterministic and permanent: the verdict is insensitive to time input at all.
    - That is what build provenance verification belongs to. There may be a new other verification in this class, but I suspect it may not be the majority of all SSC checks.

spec/predicates/simple-verification.md
marcelamelara: Ok, thanks. This should be clarified in the description.

spec/predicates/simple-verification.md
arewm: Your arguments are starting to wear on me, @AdamZWu 😉 

Do you have a specific example of this type 3? Do you have a proposed update to the PR?

I will admit that I don't have as much background in verification aspects, so I would appreciate others' input/feedback here too.

AdamZWu: > Your arguments are starting to wear on me, @AdamZWu 😉
> 
> Do you have a specific example of this type 3? Do you have a proposed update to the PR?
> 
> I will admit that I don't have as much background in verification aspects, so I would appreciate others' input/feedback here too.

Actually I don't have much of a problem with the current PR. :)

I am only concerned with the assertion that "a policy rule which can pass one day and fail another then it doesn't belong as a verified property".

I think there are several useful supply chain checks falling into this category.

For example, vulnerability checks:
- When the artifact is created, the supply chain also produces a fresh vulnerability report attestation, which (presumably) certifies there are no known vulnerabilities in it.
    - So at time T, the verification result is a PASS
- However, new vulnerabilities are being discovered and disclosed as the time goes by, hence it is unsafe to take the same vulnerability report attestation at the artifact's birth time and use it forever.
- Therefore, at some point, the vulnerability check will stop accepting the old vulnerability report attestation, because it is considered stale
    - Hence at time T+X, the verification result will turn to FAIL (and it will stay FAIL forever)

Of course the software supply chain is "live", so before the expiration of the old vulnerability report attestation, some maintenance program will kick in and generate a new vulnerability report attestation with fresh knowledge. When that happens, it is a different input to the verification, so it is out-of-scope of this discussion.

((
But interestingly, even if we extend the scope, we will eventually have a very similar behavior:
- At some time T+Y, a vulnerability that is not allowed by the policy is disclosed on the source or dependencies of the software. So future vulnerability report attestation will start to indicate the artifact does contain a certain vulnerability;
    - When that happens, all future verification results will become FAIL, regardless of the freshness of the vulnerability report attestation
))

Anyway, if we decline vulnerability verification, and other verification results of similar nature to be presented in SVA, I feel it is a considerable loss of utility.

spec/predicates/simple-verification.md
arewm: I talked with @marcelamelara about this today. I am planning to drop this, instead state that the attestation is a statement about the subject. If we want to make multiple claims about different aspects, then multiple attestations can be made.

spec/predicates/simple-verification.md
arewm: If a verifier uses a policy which can change the logic, an extra field can be used to indicate what the policy is. This would be optional and could be used by consumers if they knew about this implementation detail. I think it makes sense for the verifier to be versioned though.

arewm: > These verifiers aren't expected to be auditable, so a new revision of a verifier isn't meaningful, unless we expect that verifiers will publish the list of properties that they may include in a SVA.

If there are verified properties associated with specific verifiers, then there should be appropriate mapping provided by the verifier. 

If there is some generic property (i.e. SLSA_BUILD_LEVEL_3), non-representation is the same as having an invalid property. An attestation would need to be generated with that property present in order for it to be accepted.

spec/predicates/simple-verification.md
arewm: I changed the content, but am pushing this update manually.

spec/predicates/simple-verification.md
arewm: I added an example for how to establish this with a policy rule. The policy rule can be that a check passes within a certain time frame. Then the consumer would validate the policy rule (i.e. the verified property) against the creation date.

spec/predicates/simple-verification.md
arewm: I pushed this change.

spec/predicates/simple-verification.md
arewm: Do we need to clarify this for the `subject` now that we removed the `resourceUri`?

spec/predicates/simple-verification.md
arewm: I'm going to resolve this thread as the field is gone. Please provide separate specific feedback if you think that we should add it back.

spec/predicates/svr.md
```diff
@@ -0,0 +1,217 @@
...
+      "SLSA_BUILD_LEVEL_3",
+      "EXAMPLE_DEPLOYMENT_APPROVED"
+    ]
+  }
+}
+```
+
+The `policy` field is an extension on the verifier object that consumers
+familiar with this verifier can use to identify exactly which policy was
+applied, enabling reproducibility or auditing of the verification decision.
```
puerco, line 212: Does this mean that the policy field can be any struct? Or is it a formal resource descriptor? (I would advocate to have it as an optional resource descriptor)

arewm: The thought would be for it to be any struct as it would be dependent upon whether a policy engine would leverage a separate policy and if that policy specification would need to be taken into account for any versioning indicator.

puerco: OK, sound good to me :+1:

marcelamelara: Maybe we can add a clarification here that a resource descriptor is recommended when the policy is a file/document defined elsewhere.

spec/predicates/svr.md
```diff
@@ -0,0 +1,217 @@
+# Predicate type: Simple Verification
```
marcelamelara: just for consistency
```suggestion
# Predicate type: Simple Verification Result
```

spec/predicates/svr.md
```diff
@@ -0,0 +1,217 @@
...
+    framework.
+-   Enabling verifiers to add additional information (e.g., policy references,
+    expiration, property-level timing) as needed, while maintaining a simple
+    core schema.
+
+## Prerequisites
+
+This predicate depends on the [in-toto Attestation Framework]. Familiarity with
+the concept of attestation statements and the use of URIs for identifying
+verifiers and resources is assumed.
```
marcelamelara: I would also add here that this predicate assumes policy engine logic / trusted verifiers will be provided elsewhere.

spec/predicates/svr.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Model
+
+An SVR is an attestation that a verifier (`verifier.id`) has checked the
+subject artifact(s) at a specific time (`timeCreated`) and asserts a set of
+verified properties (`properties`). Multiple SVRs can be created for the same
+subject, and a single SVR can reference multiple subjects in the attestation
+statement. The SVR does not include information for reproducing the
+verification result. Instead, verifiers MAY generate additional provenance
+attestations for the simple verification attestation as needed.
```
marcelamelara: A small phrasing fix, and: Is the provenance attestation referenced here SLSA build provenance or any other attestation? I would expect the latter, but it would be good to clarify.

```suggestion
verification result. Instead, verifiers MAY generate additional provenance
attestations for the simple verification result as needed.
```

spec/predicates/vsa.md
```diff
@@ -9,3 +9,5 @@ Version: 1.0
 Records a SLSA verification decision about an artifact.
 
 For more details, please see https://slsa.dev/spec/v1.0/verification_summary
+
+For a simpler verification attestation predicate, see https://in-toto.io/attestation/svr/v0.1
```
marcelamelara: for consistency
```suggestion
For a simpler verification results predicate, see https://in-toto.io/attestation/svr/v0.1
```

spec/predicates/simple-verification.md
marcelamelara: We shouldn't need to since that's clarified in the Statement spec already.

spec/predicates/simple-verification.md
marcelamelara: Resolved when @arewm and I chatted on Dec 11.

spec/predicates/vsa.md
arewm: This has been pushed, resolving.

spec/predicates/svr.md
arewm: This has been pushed, resolving.

spec/predicates/svr.md
arewm: This has been pushed, resolving.

spec/predicates/svr.md
arewm: Can you be more precise about what you are looking for? Is it just something like

```
## Prerequisites

This predicate depends on the [in-toto Attestation Framework]. Familiarity with
the concept of attestation statements and the use of URIs for identifying
verifiers and resources is assumed. Any verifier producing a SVR attestation will
need to clearly provide its policy engine logic and verifier identities to consumers.
```

?

marcelamelara: yes, this text works. I was basically just looking for an added sentence that gave the disclaimer that the SVR spec isn't responsible for defining/providing verifier logic.

arewm: Thanks. I just pushed that content.

spec/predicates/svr.md
arewm: I just pushed a change for this in 0c1ddbaa4f352ce49edfb1e240a2523cf32d7d7b.

Issue #405, opened 2024-11-09
Define a "processing hint" attribute to improve supply chain adoption experience


AdamZWu:
The attestation "bundle" provides a good mechanism to generalize, allowing supply chain components to handle all kinds of attestations in a homogeneous way, which is great when the components don't want to care about attestation types.

However, there are also cases where supply chain components do want to care. For example, certain supply chain hops may want to prune attestations from their bundle according to certain criteria, such as:
- Stripping certain attestations containing enterprise proprietary data before release, or
- Remove attestations for transitive dependencies that are too "remote" and considered irrelevant for subsequent supply chain processing, etc.

Today, this is a rather painstaking process, because there is nothing directly usable at the envelope layer -- pruners will have to: 1) parse DSSE layer, 2) base64 decide the payload, 3) parse the in-toto statement, before they can have even a basic understanding of the kind of data in the attestation. And they have to repeat this for every attestation in a bundle.

I received some consult earlier from collaborators adopting in-toto, asking whether they could use the MIME type (DSSE `_type` field) to distinguish attestations, for example, `application/vnd.in-toto+json.provenance`, `application/vnd.in-toto+json.vuln`. At first I thought it was a great idea... until I found in-toto currently [mandates the MIME type](https://github.com/in-toto/attestation/blob/main/spec/v1/envelope.md#fields) to be a single uniform value.

Opening up the MIME type value space would definitely help, but it is not a general solution.

I wonder, since DSSE specs do allow custom attributes, if in-toto could define an unauthenticated field in the DSSE layer, which can be used to convey "hints" for the supply chain components to optionally use to guide their attestation processing?

The unauthenticated field does not expose security holes, thanks to the monotonic principle conformance required by in-toto -- malicious attackers can leverage this mechanism is to prune out legit attestations, but that should not lead to security holes for verification; they cannot use it to inject malicious data, because all attestations are signature checked during verification.

PR #473, opened 2025-08-05
Add predicate type to the allowed media types for the envelope's payload type

mdeicas:
This PR adds the predicate type to the allowed values for the envelope's payload type. A typo is also fixed in `bundle.md`.

Predicate-specific media types are already recommended for individual attestations in storage systems -- see
https://github.com/in-toto/attestation/blob/df7135a090fae2ed449bd95341652a9a2905b8e8/spec/v1/envelope.md#storage-convention. This PR makes the equivalent change for the envelope's `payloadType`.

The motivation is to make it easier to consume attestations (e.g. when manually opening up a bundle, https://github.com/in-toto/attestation/issues/405), and it also makes the two uses of the media types more consistent. There may be some risk that a consumer uses the media type as evidence for a policy decision, but this is warned against in the documentation.

The current `payloadType` (`application/vnd.in-toto+json`) is left as an allowed value for backwards compatibility. 

This doesn't specify how to handle predicate-specific payload types for predicates that are not registered in this repo. These cases can continue to use `application/vnd.in-toto+json`, and this can be followed up on if needed.

### DIFF
--- a/spec/v1/envelope.md
+++ b/spec/v1/envelope.md
@@ -14,9 +14,19 @@ The format and protocol are defined per [DSSE v1.0].
 The in-toto Attestation Framework has the following requirements for the
 standard DSSE fields.
 
--   `payloadType` MUST be set to `application/vnd.in-toto+json`, which
-    indicates that the Envelope contains a JSON object with a `_type` field
-    specifying its schema.
+-   `payloadType` MUST be set to `application/vnd.in-toto.predicate+json` or to
+    `application/vnd.in-toto+json`. This indicates that the Envelope contains
+    a JSON object with a `_type` field specifying its schema. If the
+    predicate-specific media type is used, the following requirements apply:
+    -   `<predicate>` MUST match the [predicate specification filename] without
+        the file extension.
+    -   Consumers SHOULD NOT rely upon the media type for individual attestations
+        as faithful indicators of predicate type. Consumer SHOULD only rely on the
+        `predicateType` field in the [Statement] layer.
+    -   To obtain predicate information that is authenticated, consumers MUST
+        parse the Envelope's `payload`, and verify it against its `signatures`.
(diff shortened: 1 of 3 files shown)


marcelamelara (approved):
Thanks for this PR @mdeicas ! Apologies for the long turnaround time on this one.

marcelamelara:
Since this is a spec change, can we please get a second @in-toto/attestation-maintainers review? I believe we'll also want to bump the minor version of the spec with this PR.

spec/v1/envelope.md
```diff
@@ -14,9 +14,19 @@ The format and protocol are defined per [DSSE v1.0].
 The in-toto Attestation Framework has the following requirements for the
 standard DSSE fields.
 
--   `payloadType` MUST be set to `application/vnd.in-toto+json`, which
-    indicates that the Envelope contains a JSON object with a `_type` field
-    specifying its schema.
+-   `payloadType` MUST be set to `application/vnd.in-toto.predicate+json` or to
```
TomHennen: should this be `<predicate file name>` as above or `<predicate>` as in https://github.com/in-toto/attestation/blob/df7135a090fae2ed449bd95341652a9a2905b8e8/spec/v1/envelope.md#storage-convention?

marcelamelara: yeah, these should be consistent. I would stick with `<predicate>` everywhere with a link to the explanation as below.

marcelamelara:
@mdeicas Can you please rebase your branch onto the current main to fix the linter errors?

spec/v1/envelope.md
mdeicas: Agreed, done.

mdeicas:
Thanks for reviewing! The linter is now passing and I bumped the minor version to v1.2. Let me know if I didn't do that correctly.

TomHennen (approved):

PR #488, opened 2025-09-08
Upgrade Go in make-protos.yml@setup-go to be consistent w/ go.mod

trishankatdatadog:
Fix unintended consequence IIUC from https://github.com/in-toto/attestation/pull/485


marcelamelara (approved):
Thanks for catching this!

PR #495, opened 2025-10-07
Release predicate v0.2

### LINKED ISSUE #492: Consider renaming releaseId in release attestation
The release attestation has `releaseId` as a field https://github.com/in-toto/attestation/blob/main/spec/predicates/release.md#fields.

The semantics of this value is hard to get from the name, it's a common misconception that the `releaseId` is the id for _this_ release version, not the stable value (between releases) it represents to be able to detect resurrection attacks. 

If we look at the described [Purpose](https://github.com/in-toto/attestation/blob/main/spec/predicates/release.md#purpose) we can read:

> To authoritatively link from a specific release name and version string in a package registry, to the artifact names and hashes that make up that release.

This makes me think that `packageId` is a better name for the stable identifier for _all_ releases for this package. 

If accepted, this would be a breaking change and so 0.2 version would need to be published. 

Thoughts?

cc @steiza

bdehamer:
Closes #492 

Bumps the release predicate from v0.1 to v0.2 in order to rename the `releaseId` predicate field to `packageId`.

[Rendered version](https://github.com/bdehamer/attestation/blob/12ccefd69681644b82a1f8f9d066e0bb4a5ce594/spec/predicates/release.md)

### DIFF
--- a/spec/predicates/release.md
+++ b/spec/predicates/release.md
@@ -2,7 +2,7 @@
 
 Type URI: https://in-toto.io/attestation/release
 
-Version 0.1
+Version 0.2
 
 ## Purpose
 
@@ -74,7 +74,7 @@ B.
   "predicateType": "https://in-toto.io/attestation/release/v0.1",
   "predicate": {
     "purl": <ResourceURI>,
-    "releaseId": "..."
+    "packageId": "..."
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


marcelamelara (changes requested):
Thank you @bdehamer ! The spec changes LGTM, but can you please also update the corresponding .proto file under `protos/` ? No need to regenerate the language bindings, since that's handled by the CI.

PR #548, opened 2026-04-15
docs(spec): restore spdx.md as a redirect to spdx2.md / spdx3.md

### LINKED ISSUE #547: Broken link to spx predicate
PR #508 moved /spec/predicates/spdx.md to /spec/predicates/spdx2.md breaking all external links to the original file.
We have [one such instance in the SLSA spec](https://github.com/slsa-framework/slsa/issues/1577) and although we could update the spec I feel like this should really be handled on this side.

A redirect would be in order but I realize this isn't possible in a github repo so I'm not sure what the right solution is but there should still be a spdx.md, maybe pointing to the two difference versions?

SAY-5:
Per #547, #508 moved `spec/predicates/spdx.md` to `spdx2.md` (and later added `spdx3.md`), which broke every external link to the original file — including the one in the [SLSA spec](https://github.com/slsa-framework/slsa/issues/1577).

Restore `spdx.md` as a stable redirect page that points at both version-specific specs and explains the move. This keeps long-lived external references working without requiring downstream specs to chase the rename.

Closes #547


marcelamelara (commented):
Thanks @SAY-5 ! LGTM. Can you please fix your DCO and the linter issues?

PR #555, opened 2026-05-21
spec: bump SVR predicate to v0.2 and require policies

### LINKED ISSUE #551: Update the SVR to have a required field verifier.policies
As discussed on the community call today, the policy used by the verifier in producing the SVR is often a critical part of an result. We only have this present in the last example on the specification: https://github.com/in-toto/attestation/blob/main/spec/predicates/svr.md#optional-policy-information-example:

```json
    "verifier": {
      "id": "https://example.com/policy_engine/v1",
      // Optional extension field specific to this verifier
      "policy": {
        "uri": "https://example.com/policies/production-deployment",
        "digest": {"sha256": "789abc012def..."}
      }
```

We should update the parsing rules for the SVR to require policies to be parsed. If no policy is present, the parser should know how to interpret that non-representation.

arewm:
Make policy references explicit in SVR so consumers can consistently reason about which policy inputs were applied, including explicit no-policy representation via an empty policies array.

Fixes #551

Assisted-by: Codex 5.3

### DIFF
--- a/spec/predicates/svr.md
+++ b/spec/predicates/svr.md
@@ -1,8 +1,8 @@
 # Predicate type: Simple Verification Result
 
-Type URI: https://in-toto.io/attestation/svr/v0.1
+Type URI: https://in-toto.io/attestation/svr/v0.2
 
-Version: 0.1
+Version: 0.2
 
 Authors: Tom Hennen (@TomHennen), Andrew McNamara (@arewm)
 
@@ -32,8 +32,8 @@ policies.
 This predicate depends on the [in-toto Attestation Framework]. Familiarity with
 the concept of attestation statements and the use of URIs for identifying
 verifiers and resources is assumed. Any verifier producing a SVR attestation
-will need to clearly provide its policy engine logic and verifier identities
-to consumers.
+will need to clearly provide its policy engine logic, verifier identities, and
+the policy references used during verification to consumers.
 
 ## Model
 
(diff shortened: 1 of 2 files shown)


spec/predicates/svr.md
```diff
@@ -70,6 +71,10 @@ attestations for the simple verification result as needed.
 This predicate follows the
 [in-toto Attestation Framework's parsing rules](../v1/README.md#parsing-rules).
 
+The `verifier.policies` field MUST be present. If no explicit policies were
+used, or the verifier cannot reference the policies, producers MUST encode this
+as an empty array (`"policies": []`).
```
puerco, line 76: I love this. In the end we want to ensure that a verification occurred based on a policy. There may be corner cases where policies are not required, but I think this would mostly be a bug. If there was no policy involved in the verification I want to know.

arewm: Are you suggesting that I need to change the language in this?

puerco (approved):
This looks good to me if we want to scope 0.2 to this change only

arewm:
I am okay to have a 0.2 change including only this even though there were some other modifications that we had talked about previously. There is of course a balance between multiple smaller changes and updating a predicate too frequently. I suspect that many folks haven't adopted this yet so multiple versions are not going to be that detrimental.

arewm:
@puerco , what else is needed to be able to merge this?

spec/predicates/svr.md
puerco: No, i think it's right as is!

puerco:
We just need another approval from one of the other maintainers.

/cc @in-toto/attestation-maintainers please take a look whenever possible 🙏

spec/predicates/svr.md
adityasaky, line 76: I like this explicit treatment for a scenario with no policies. I'm curious, though, if this specific approach will cause problems in tooling. For instance, we use protobuf to define schemas for predicates, and have libraries generated in various languages. I wonder if we'll have any trouble with one library silently dropping the field because it's empty while another keeps it. Maybe protobuf has a way to specify this behavior!

This is not meant to be a blocker--I imagine we'll add the svr v0.2 definition in a follow up PR? We can test then and revisit how to handle the no policy case if needed then.

adityasaky (approved):
Thanks for working on this! I think making policies a MUST is a good idea on the whole.

Issue #557, opened 2026-06-09
Proposal: runtime-trace v0.2 (backward-compatible additions and clarifications)


rung:
## Overview

I would like to discuss a v0.2 of the [runtime-trace predicate][runtime-trace]. Two kinds of change:

1. Backward-compatible field additions
2. Clarification of the intent and representation of existing fields

Both are designed with verification in mind (e.g. Sigstore policy-controller).

## Background and motivation

I am building [cicd-sensor][cicd-sensor], an eBPF-based CI/CD runtime security tool, and want to emit its runtime behavior as a build attestation.

While designing its predicate on top of [runtime-trace v0.1][runtime-trace] I found I needed some extensions, and after posting on the [CNCF Slack][slack] it was suggested to open a GitHub issue, hence this one.

I have aimed for a proposal that fits my use case while aligning with the existing runtime-trace design: it intentionally avoids being cicd-sensor specific, and aims to work naturally for other monitors such as Tetragon and to verify cleanly with tools like policy-controller.

### Full example (relevant parts)

Annotations mark only what this proposal touches; unannotated fields are existing v0.1, unchanged. `[NEW]` proposed addition / `[clarify]` existing field, entry shape clarified / `[discuss]` open question / `[ext]` producer extension.

```json5
{
  "_type": "https://in-toto.io/Statement/v1",
  "subject": [{ "name": "pkg:...", "digest": { "sha256": "..." } }],
  "predicateType": "https://in-toto.io/attestation/runtime-trace/v0.2", // v0.2 type URI
  "predicate": {
    "monitor": { "type": "https://example.org/runtime-monitor" },
    "monitoredProcess": {                                               // [discuss] (section 3)
      "hostID": "https://github.com",
      "type": "<TypeURI: TBD>",
      "event": "https://github.com/<org>/<repo>/actions/runs/<run_id>/attempts/<n>"
    },
    "monitorResult": "FAILED",                                          // [NEW] (section 1)
    "monitorLog": {
      "network": [                                                      // [clarify] (section 2)
        { "ip": "140.82.112.3", "domain": "github.com", "port": 443, "protocol": "tcp" },
        { "ip": "8.8.8.8", "port": 53, "protocol": "udp" },
        { "domain": "example.test" }
      ]
    },
    "https://<producer>/runtime_trace/detections/v1alpha1": [ ... ],    // [ext] (section 4)
    "metadata": {
      "buildStartedOn": "...",
      "buildFinishedOn": "..."
    }
  }
[...]

PR #568, opened 2026-07-12
runtime-trace v0.2: define monitorLog.network entry shape, clarify monitoredProcess

rung:
Part of #557. As suggested there, this PR contains the parts of the proposal we converged on; everything else stays in the issue or breakaway issues/PRs.

## Changes

-   **Defined a recommended entry shape for `monitorLog.network`**: `ip`, `hostname`, `port`, `protocol`, and an optional monitor-dependent `process`. All fields are optional, with SHOULD-level constraints (at least one of `ip`/`hostname`; `protocol` when `port` is present). Producers MAY add monitor-specific fields to entries.
-   **Clarified `monitoredProcess` for CI/CD job identity**: `hostID` records the platform, `type` SHOULD reuse the platform's published SLSA `buildType` URI, and `event` SHOULD uniquely identify the job or run so the trace can be cross-referenced with co-located provenance. Added a GitHub Actions example to each field description to make this concrete.

## Design notes

-   `hostname` (rather than `domain`) and the optional `process` field (e.g. `{ "pid": 1234, "comm": "curl" }`) align with the shape used by witness's network tracing work ([discussion](https://github.com/in-toto/attestation/issues/557#issuecomment-4663042060)).
-   Per the discussion, a monitor verdict (`monitorResult`) is **not** added.
-   The existing Tekton example's `monitoredProcess` is left as-is; the new CI/CD guidance is illustrated in the field descriptions.

## Open question

-   I added myself to the `Authors` line as the author of the v0.2 changes, but I could not find an established convention for updating this line on a version bump (the predicate template has no `Authors` field). Please let me know if you'd prefer this handled differently, e.g. dropping the addition, or also crediting other participants in the #557 discussion.

### DIFF
--- a/spec/predicates/runtime-trace.md
+++ b/spec/predicates/runtime-trace.md
@@ -187,7 +187,7 @@ Port number of the remote endpoint, between 1 and 65535.
 Transport protocol, lowercase. Ex: `tcp`, `udp`. If `port` is present,
 `protocol` SHOULD also be present.
 
-`monitorLog.network[*].process` _object_, _optional_
+`monitorLog.network[*].invokingProcess` _object_, _optional_
 
 The process that generated the network activity. The exact format of this
 field is dependent on the monitor, and can be determined using `monitor.type`.
--- a/spec/predicates/runtime-trace.md
+++ b/spec/predicates/runtime-trace.md
@@ -71,7 +71,7 @@ runtime trace information.
                     "hostname": "<STRING>",
                     "port": /* integer */,
                     "protocol": "<STRING>",
-                    "process": { /* object */ }
+                    "invokingProcess": { /* object */ }
                 }
             ],
             "fileAccess": ["<ResourceDescriptor>", ...]
--- a/spec/predicates/runtime-trace.md
+++ b/spec/predicates/runtime-trace.md
@@ -1,11 +1,11 @@
 # Predicate type: Runtime Trace
 
-Type URI: https://in-toto.io/attestation/runtime-trace/v0.1
+Type URI: https://in-toto.io/attestation/runtime-trace/v0.2
 
-Version: 0.1.0
+Version: 0.2.0
 
 Authors: Parth Patel (@pxp928), Shripad Nadgowda (@nadgowdas),
-  Aditya Sirish A Yelgundhalli (@adityasaky)
+  Aditya Sirish A Yelgundhalli (@adityasaky), Hiroki Suezawa (@rung)
 
 ## Purpose
 
@@ -49,7 +49,7 @@ runtime trace information.
 {
     "_type": "https://in-toto.io/Statement/v1",
     "subject": [{ ... }],
-    "predicateType": "https://in-toto.io/attestation/runtime-trace/v0.1",
+    "predicateType": "https://in-toto.io/attestation/runtime-trace/v0.2",
     "predicate": {
         "monitor": {
             "type": "<TypeURI>",
@@ -66,7 +66,13 @@ runtime trace information.
                 { /* object */ }
             ],
             "network": [
-                { /* object */ }
+                {
+                    "ip": "<STRING>",
+                    "hostname": "<STRING>",
+                    "port": /* integer */,
+                    "protocol": "<STRING>",
+                    "process": { /* object */ }
+                }
             ],
             "fileAccess": ["<ResourceDescriptor>", ...]
         },
@@ -121,16 +127,26 @@ Identifies the process being monitored.
 
 `monitoredProcess.hostID` _string (URI)_, _required_
 
-URI indicating the process host’s identity.
+URI indicating the process host’s identity. Ex: when monitoring a job on a
+CI/CD platform, this field records the platform’s identity, such as
+`https://github.com`.
 
 `monitoredProcess.type` _string (TypeURI)_, _required_
 
 URI indicating the type of process performed. Ex: when monitoring a build, this
-field records the build type.
+field records the build type. For builds on CI/CD platforms, producers SHOULD
+reuse the platform’s published SLSA `buildType` URI, such as
+`https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1` for
+GitHub Actions.
 
 `monitoredProcess.event` _string_, _required_
 
 String identifying the specific job or task associated with the attestation.
+The value SHOULD uniquely identify the monitored job or task, so that
+consumers can tie the trace back to the exact run and cross-reference it with
+other attestations for the same run, such as build provenance. Ex:
+`https://github.com/<org>/<repo>/actions/runs/<run_id>/attempts/<n>` for a
+GitHub Actions job.
 
 `monitorLog` _object_, _required_
 
@@ -146,10 +162,36 @@ schema.
 
 `monitorLog.network` _list of objects_, _optional_
 
-Record of network activity observed by monitor. The exact format of this field
-is currently dependent on the monitor, and can be determined using
-`monitor.type`. In future, after consulting different monitors, this field may
-have a consistent schema.
+Record of network activity observed by monitor. Entries SHOULD use the
+following fields so that consumers can process network activity consistently
+across monitors. Producers MAY include additional, monitor-specific fields in
+each entry.
+
+`monitorLog.network[*].ip` _string_, _optional_
+
+IP address of the remote endpoint. Both IPv4 and IPv6 addresses use this
+field. The canonical textual form is recommended (for IPv6, [RFC 5952]).
+
+`monitorLog.network[*].hostname` _string_, _optional_
+
+Observed or resolved hostname of the remote endpoint. An entry SHOULD carry at
+least one of `ip` or `hostname`, and MAY carry just one of them when the
+association between the two is unknown.
+
+`monitorLog.network[*].port` _integer_, _optional_
+
+Port number of the remote endpoint, between 1 and 65535.
+
+`monitorLog.network[*].protocol` _string_, _optional_
+
+Transport protocol, lowercase. Ex: `tcp`, `udp`. If `port` is present,
+`protocol` SHOULD also be present.
+
+`monitorLog.network[*].process` _object_, _optional_
+
+The process that generated the network activity. The exact format of this
+field is dependent on the monitor, and can be determined using `monitor.type`.
+Ex: `{ "pid": 1234, "comm": "curl" }`.
 
 `monitorLog.fileAccess` _list of ResourceDescriptor objects_, _optional_
 
@@ -186,7 +228,7 @@ Other properties of the monitoring event.
 ```json
 {
   "_type": "https://in-toto.io/Statement/v1",
-  "predicateType": "https://in-toto.io/attestation/runtime-trace/v0.1",
+  "predicateType": "https://in-toto.io/attestation/runtime-trace/v0.2",
   "subject": [
     {
       "name": "ttl.sh/testin123",
@@ -262,6 +304,19 @@ Other properties of the monitoring event.
           ],
           "privileged": null
         }
+      ],
+      "network": [
+        {
+          "ip": "140.82.112.3",
+          "hostname": "github.com",
+          "port": 443,
+          "protocol": "tcp"
+        },
+        {
+          "ip": "8.8.8.8",
+          "port": 53,
+          "protocol": "udp"
+        }
       ]
     },
     "metadata": {
@@ -271,3 +326,18 @@ Other properties of the monitoring event.
   }
 }
 ```
+
+## Changelog and Migrations
+
+### v0.2
+
+-   Defined a recommended entry shape for `monitorLog.network` entries;
+    the format was previously monitor-dependent, so producers SHOULD
+    migrate to the recommended field names.
+-   Clarified how `monitoredProcess` identifies CI/CD jobs.
+
+### v0.1
+
+Initial version.
+
+[RFC 5952]: https://www.rfc-editor.org/rfc/rfc5952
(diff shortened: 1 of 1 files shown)


stupendoussuperpowers:
The network/monitoredProcess changes look good from a witness/SBOMit aspect as well! Before we finalize all this, I would also like to consider:

- A well-defined structure for `monitorLog.process[]` entries as well. The stated example does assume a well-typed struct for the entries and formalizing this might be useful. 

- How should `eventType`s be covered in such a standardization. For instance, `process` and `exit` do not seem compatible (we could have either `spawn/fork/exec` and `exit`, or just track processes and not lifecycle events)

- How the `process` in a network log map to the `monitorLog.process[]` entry, does it restate the entire entry or do we use identifiers like PID etc.

rung:
Thanks @stupendoussuperpowers! These are all questions worth settling, and I'd like to be part of that discussion. To keep this PR focused on what we already converged on in #557 (the `network` entry shape and the `monitoredProcess` clarifications, per @adityasaky's suggestion), how about a breakaway issue for the `monitorLog.process[]` shape? As I mentioned in #557, I think it's best designed building on the hands-on experience from the witness/SBOMit work, and I'd be glad to contribute from the cicd-sensor side. In any case, this PR doesn't block any of it.

On each point:

- `monitorLog.process[]` shape: agreed it's worth standardizing. From monitoring real CI jobs with cicd-sensor, a single job can spawn hundreds of processes or more, so I'd first lay out the use cases (per-event entries, a summary layer, or both) before fixing a shape. Unlike `network`, where the same entry shape works for both per-connection entries and a deduplicated summary (https://github.com/in-toto/attestation/issues/557#issuecomment-4664450678), per-event process entries and a summary would likely need different shapes.
- `eventType`: I see this as part of the same discussion, since lifecycle modeling (`spawn/fork/exec` + `exit` vs. tracking processes without lifecycle events) changes the entry shape fundamentally. The current example is inherited from v0.1 (Tetragon-based) and isn't normative.
- `network[*].process` mapping: this PR intentionally leaves it monitor-dependent so we don't pre-decide this. Whether it should be an identifier-style reference (PID) into `process[]` or a small self-contained object can be settled together with the `process[]` shape.

astrogilda (commented):
This matches what #557 converged on, and the shape works from the agent-runtime side too. We emit this kind of egress record and could map to these field names directly. A few points from writing policy against records like these.

First, direction. The fields describe "the remote endpoint," which reads naturally for outbound activity, but monitors that see both sides (eBPF tracers included) will also observe inbound accepts. A consumer writing egress policy needs to tell an outbound connect to `140.82.112.3:443` apart from an inbound connection whose peer happens to be that address. Without either an optional `direction` field (`outbound`/`inbound`) or an explicit statement that entries record the remote peer regardless of direction, two monitors will encode this differently, and that is the kind of drift this PR is trying to prevent.

Second, where `hostname` came from. "Observed or resolved" covers three different trust levels in practice: a hostname the workload itself asked for (DNS query, TLS SNI), one the monitor forward-resolved, and one reverse-resolved from the IP. For policy purposes these aren't interchangeable, since an attacker has far more influence over reverse DNS than over SNI. I'm not necessarily asking for a new field here. Even a sentence noting that producers SHOULD say, in their monitor-specific fields, how the hostname was obtained would stop consumers from treating reverse DNS as authoritative.

Last one: entries carry no time field, so events from one run cannot be ordered or correlated with build steps except by monitor-specific extras. The spec's only timestamps are the run-level `metadata.buildStartedOn` / `buildFinishedOn`, which bound the window but can't order anything inside it. Since the goal is consistent processing across monitors, would it be worth adding an optional RFC 3339 `timestamp` per entry now? Retrofitting one after producers have shipped their own variants gets messy. Fine to defer if you want to keep this PR minimal.

Leaving the `process` object monitor-dependent pending the breakaway @stupendoussuperpowers raised makes sense; worth cross-linking that issue from the field description once it exists. The changelog and migration section is a nice touch.

spec/predicates/runtime-trace.md
```diff
@@ -66,7 +66,13 @@ runtime trace information.
                 { /* object */ }
             ],
             "network": [
-                { /* object */ }
+                {
+                    "ip": "<STRING>",
+                    "hostname": "<STRING>",
+                    "port": /* integer */,
+                    "protocol": "<STRING>",
+                    "process": { /* object */ }
```
adityasaky: I think it's a good idea to have a shape here to prevent a complete free for all. The only concern I have with a specific shape for this is whether we accidentally prevent including some information that doesn't cleanly fit into one of these fields. cc @pxp928 for thoughts on this.

adityasaky: > Producers MAY include additional, monitor-specific fields in each entry.

My concern is somewhat abated further down. :D

spec/predicates/runtime-trace.md
```diff
@@ -146,10 +162,36 @@ schema.
...
+`monitorLog.network[*].port` _integer_, _optional_
+
+Port number of the remote endpoint, between 1 and 65535.
+
+`monitorLog.network[*].protocol` _string_, _optional_
+
+Transport protocol, lowercase. Ex: `tcp`, `udp`. If `port` is present,
+`protocol` SHOULD also be present.
+
+`monitorLog.network[*].process` _object_, _optional_
```
adityasaky:
```suggestion
`monitorLog.network[*].invokingProcess` _object_, _optional_
```

For clarity perhaps?

spec/predicates/runtime-trace.md
```diff
@@ -121,16 +127,26 @@ Identifies the process being monitored.
 
 `monitoredProcess.hostID` _string (URI)_, _required_
 
-URI indicating the process host’s identity.
+URI indicating the process host’s identity. Ex: when monitoring a job on a
+CI/CD platform, this field records the platform’s identity, such as
+`https://github.com`.
```
adityasaky, line 132: Would this be too broad for a large platform like github? FWIW, I think this isn't really an issue introduced by this PR but rather ambiguity that's getting surfaced now.

adityasaky (commented):
Thanks for filing this @rung!

spec/predicates/runtime-trace.md
rung: Thanks for the suggestion! There was a discussion about aligning this field shape with witness's network tracing work where possible (https://github.com/in-toto/attestation/issues/557#issuecomment-4663042060). 
@stupendoussuperpowers, could you confirm whether renaming this to invokingProcess works from that side?

stupendoussuperpowers: Don't have an issue with this name change, either works.

spec/predicates/runtime-trace.md
rung: @adityasaky 
Thanks for raising this. As a producer (I'm planning to emit runtime-trace attestations for GitHub Actions and GitLab CI/CD), I'd like to keep a concrete example here to avoid guessing. At this stage, I'd align the granularity with the existing Tekton example, whose `hostID` is the SLSA `builder.id` value Tekton Chains emits. Concretely, I'd update the description as follows; would that address your concern?

```diff
 URI indicating the process host’s identity. Ex: when monitoring a job on a
-CI/CD platform, this field records the platform’s identity, such as
-`https://github.com`.
+CI/CD platform, this field records the identity of the environment hosting
+the job, such as `https://github.com/actions/runner/github-hosted` for a
+job on a GitHub-hosted runner.
```

This value distinguishes GitHub-hosted from self-hosted runners, and keeps `hostID` at the same granularity as SLSA provenance's `builder.id`:

| runtime-trace | SLSA provenance v1 | Existing Tekton example | GitHub Actions example (proposed) |
| --- | --- | --- | --- |
| `hostID` | `runDetails.builder.id` | `https://tekton.dev/chains/v2` | `https://github.com/actions/runner/github-hosted` |

If `hostID` itself turns out to need rethinking (its semantics or naming), would a separate issue be the right place, given the ambiguity predates this PR?

spec/predicates/runtime-trace.md
rung: Thanks @stupendoussuperpowers for confirming, and @adityasaky for the suggestion. Applied: renamed the field to invokingProcess in both the JSON skeleton and the field description.

rung:
Thanks for the detailed review, @adityasaky! I've replied to each comment.
On https://github.com/in-toto/attestation/pull/568#discussion_r3606111643, I'll leave the thread open and wait for @pxp928's thoughts.

astrogilda:
Now that the invokingProcess rename has landed and the entry shape looks close to settled, I want to circle back on the three points from my earlier review, since none of them has surfaced in the revisions so far. We're building a producer (and a policy consumer) against exactly the v0.2 shape, which is why I keep pressing. Each point is a one-or-two-line addition today, and considerably more expensive to retrofit once producers ship v0.2 with their own monitor-specific variants. To make them as cheap as possible to act on, here is ready-to-paste wording for each. Deferring any of them is a legitimate outcome; I'd just ask for an explicit yes/no/defer on each, so nothing falls through review a second time.

1. Direction / remote peer. The entry fields describe "the remote endpoint," which reads outbound-first. But monitors that see both sides (eBPF tracers included) will also record inbound accepts, and a consumer writing egress policy has to tell an outbound connect to `140.82.112.3:443` apart from an inbound connection whose peer happens to be that address. There are two ways to close it, and either one works for us.

The zero-field version, appended to the `monitorLog.network` intro paragraph:

> Each entry describes the remote peer of the observed activity, regardless of which side initiated the connection.

Or, if you'd rather let monitors state it when they know it, an optional field in the existing style:

```
`monitorLog.network[*].direction` _string_, _optional_

Direction of the activity from the monitored process's perspective, lowercase.
One of `outbound` (initiated by the monitored process) or `inbound` (initiated
by the remote peer). If absent, consumers cannot assume a direction.
```

The one-sentence version costs nothing and prevents the drift; the field version additionally lets policy distinguish the two cases. I'd take either over leaving it unstated, since that's how two monitors end up encoding it differently.

2. Hostname provenance. "Observed or resolved" spans three trust levels: a hostname the workload asked for (DNS query, TLS SNI), one the monitor forward-resolved, and one reverse-resolved from the IP. An attacker has far more influence over reverse DNS than over SNI, so I'm not asking for a field here, just one sentence appended to the hostname description:

> Producers SHOULD indicate, in monitor-specific fields or in the monitor's documentation, how the hostname was obtained (for example: observed from the workload's DNS query or TLS SNI, forward-resolved by the monitor, or reverse-resolved from the IP), since these carry different trust levels for consumers.

That single SHOULD is enough to stop consumers from treating reverse DNS as authoritative, without constraining any producer.

3. Per-entry timestamp. I flagged before that retrofitting gets messy, and working through our producer's field mapping has sharpened that. Entries carry no time field, and the only timestamps in the predicate are the run-level `buildStartedOn`/`buildFinishedOn`. Nothing inside the window can be ordered or correlated with build steps. Every real monitor already has this value; it's in the event before the producer strips it. Producers who need it will ship it under monitor-specific names the moment v0.2 lands. If a canonical field then arrives in v0.3, consumers face three regimes at once: entries with no time, entries with a vendor field, and entries with the canonical field. Reserving the name now is one optional row and obligates nobody:

```
`monitorLog.network[*].timestamp` _string_, _optional_

Time at which the activity was observed, as an [RFC 3339] timestamp. Ex:
`2026-01-15T04:31:28Z`. When a monitor aggregates repeated activity to the
same endpoint into one entry, this SHOULD be the time of the first
observation.
```

plus `[RFC 3339]: https://www.rfc-editor.org/rfc/rfc3339` alongside the existing RFC 5952 link.

If you want to keep this PR strictly to what #557 converged on, deferring 3 (and the direction field form of 1) to a follow-up is fine by me. The sentence forms of 1 and 2 are pure clarification of text this PR already adds, so I'd argue they belong here. Either way, an explicit call on each would help us lock our producer's field mapping against v0.2 rather than guessing.

spec/predicates/runtime-trace.md
```diff
@@ -121,16 +127,26 @@ Identifies the process being monitored.
...
 
 `monitoredProcess.type` _string (TypeURI)_, _required_
 
 URI indicating the type of process performed. Ex: when monitoring a build, this
-field records the build type.
+field records the build type. For builds on CI/CD platforms, producers SHOULD
+reuse the platform’s published SLSA `buildType` URI, such as
+`https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1` for
+GitHub Actions.
```
adityasaky, line 140: more of a nit: Can we point the example to what github uses officially when producing an attestation using `actions/attest`?

spec/predicates/runtime-trace.md
adityasaky: I think so. I'd also like input from @puerco here 😄

adityasaky (commented):
I think this is pretty close to being approved from my end! @puerco, @pxp928 could you take a look when you get a chance?

spec/predicates/runtime-trace.md
rung: @adityasaky 
Thanks for the review! updated. `actions/attest` actually emits `https://actions.github.io/buildtypes/workflow/v1`, which resolves to the GitHub-maintained spec at [actions/buildtypes](https://github.com/actions/buildtypes/blob/main/workflow/v1/README.md), so I've pointed the example there.

astrogilda:
@rung, my three 19 July points are still open at head 4f7166ef. No `direction`, no `timestamp`, and hostname still reads "Observed or resolved", so the branch has moved twice without touching any of them. Two only clarify text this PR adds, and an eBPF tracer records inbound accepts a consumer cannot tell from an outbound connect. Dropping monitorResult was right; a yes, no or defer on each locks my mapping, and I will contribute a `vectors-runtime-trace` corpus to `agent-evidence-vectors` once this merges.

astrogilda:
Thanks, @rung, that works: none of the three conflicts with v0.2, so each can wait. Once #568 merges I'll open the follow-up with one commit per point, so each can be taken or declined on its own.

I'd also propose runtime-trace cases for the conformance layer in #597, an accept and a reject vector per point. A verifier could then show it enforces each.

Issue #184, opened 2023-04-04
How to handle gitCommit with sha256 commits


TomHennen:
@laurentsimon filed [this bug](https://github.com/slsa-framework/slsa/issues/797) in slsa-framework, but we probably need to consider it here too.

Do we need to provide any guidance for dealing with sha256 `gitCommit`s.

adityasaky:
I meant to post this here but left it on the SLSA thread by mistake:

I don't know of GitHub's support for this but last I checked, there's a [4 stage transition plan in Git](https://git-scm.com/docs/hash-function-transition/2.40.0#_object_names_on_the_command_line) where phase 2 and 3 support both SHA-1 and SHA-256 hashes. I don't know if this thinking has changed elsewhere, perhaps on the ML, because this doc is quite old. I _believe_ on the in-toto side we can update our support for `git{Commit,Blob,Tree,Tag}` to indicate it could be SHA-1 or SHA-256. Differentiating by length should also suffice for now because a transition plan isn't active, if someone's using a SHA-256 Git repo, it has no interop with SHA-1 repos. When there is interop, we'll know how to identify object equivalence when presented with its SHA-1 and SHA-256 hashes.

MarkLodato:
Same as git, different length. See https://git-scm.com/docs/hash-function-transition/2.40.0#_object_names:

> Objects can be named by their 40 hexadecimal digit SHA-1 name or 64 hexadecimal digit SHA-256 name

We should document this.

TomHennen:
Would there be a need for an attestation to reference both the sha1 and sha256 commits simultaneously?  If so there might be a problem.

Apparently repos [will allow users to use either sha1 or sha256](https://git-scm.com/docs/hash-function-transition/2.40.0#_object_names:~:text=Users%20can%20use%20SHA%2D1%20and%20SHA%2D256%20identifiers%20for%20objects%20interchangeably%20(see%20%22Object%20names%20on%20the%20command%20line%22%2C%20below)).  Since [DigestSet](https://github.com/in-toto/attestation/blob/main/spec/v1.0/digest_set.md) uses a map there can only be one value for `gitCommit`, so generators will have to pick which one to use??

If someone trying to verify a commit matches they'll either need to have both `sha1` or `sha256` available or be able to read the commit from the attestation and then look it up that way?

Alternatively we could recommend people always use X.

Does that make sense or am I missing something?

adityasaky:
[...]

PR #578, opened 2026-07-27
go/v1: accept SHA-256 git object digests

arpitjain099:
digest_set.md says gitCommit, gitTree, gitBlob and gitTag hold the lowercase hex of the object's SHA-1 (40 char) or SHA-256 (64 char) hash. But HexLength() returns 20 bytes for all four git algorithms, so ResourceDescriptor.Validate rejects a valid SHA-256 git digest with ErrIncorrectDigestLength.

This tracks the accepted byte sizes per algorithm instead of a single length, and lets the git object algorithms accept 20 or 32 bytes. Existing SHA-1 digests keep working, and a wrong-length git digest is still rejected. Added a test case covering a 64-char gitCommit digest.

Refs #184.

### DIFF
--- a/go/v1/resource_descriptor.go
+++ b/go/v1/resource_descriptor.go
@@ -8,6 +8,9 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"slices"
+	"strconv"
+	"strings"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


puerco:
Ah thanks for handling this @arpitjain099  I realized we were missing this right after we merged the original parch and had been in my mental todo since then. I'll review it in a little bit

puerco (approved):
Thanks for fixing this, lgtm

PR #570, opened 2026-07-15
Add Adversarial Execution Evidence predicate (v0.7)

astrogilda:
## Adversarial Execution Evidence predicate

This adds a predicate for the evidence produced by deliberately executing an untrusted artifact against a known adversarial corpus inside a containment substrate: agent tools, MCP servers, plugins, build steps. The producer runs the artifact, injects the corpus, and signs what the substrate intercepted.

The predicate has been through fifteen rounds of review here, and @Rul1an has written a checker from the text alone and runs its conformance corpus, so the field shape is settled enough to read as written. They are the same reader in every reference below. I still want review of the fields before the vetting meeting. To answer the [new-predicate guideline](../docs/new_predicate_guidelines.md) questions up front:

The use case is an admission controller gating a third-party MCP or agent image on evidence it was detonated against a named corpus under an enforcing catch policy, and an auditor re-verifying a specific interception offline without trusting the producer's infra. Existing predicates do not cover it: [runtime-trace] carries raw monitor activity with no corpus binding, no coverage denominator, and no per-event signature; [SCAI] carries attribute assertions, not an adversarial corpus with recomputable outcomes; [VSA] and [SVR] carry policy verdicts computed downstream of evidence like this. This is the evidence layer those consume, so verdict semantics stay out of scope here. The policy question it answers is which attacks this exact image faced, what the substrate did about each, and under what network posture. A consumer recomputes all of that from the attestation plus the producer's published taxonomy, with no call back to the producer.

The shape carries an explicit `doesNotAssert` negative scope, and five further design properties motivate it. I'm happy to defend or change any of them in review:

- A recomputable result: a deterministic function of the carried evidence, which a consumer re-derives for itself instead of taking the producer's word for it.
- An attack-granular coverage binding, where the coverage denominator is a digest-committed manifest mapping each class to its attack ids and carried in the attestation, so an edit to the assessed set changes `corpus.digest`.
- Independently signed intercept records (DSSE-shaped, verify-then-read) under a single predicate-level `batchRoot`.
- An I-JSON safe-integer profile on every rail, so canonicalization is byte-identical across languages.
[...]

PR #581, opened 2026-07-28
Add source-review-coverage predicate

DrVelvetFog:
Proposes a new predicate, `source-review-coverage`, per the [new predicate guidelines](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md). Answering the four questions asked there:

### What's your use case?

SLSA v1.2 Source Track Level 4 requires that changes to protected branches "be agreed to by two or more trusted persons prior to submission," and requires this of *the final revision submitted* — with re-review if changes occur during review. That requirement currently has no verifier. SLSA states that it "leaves source provenance attestations undefined and up to the SCSs to determine what works best," and does not address how squash merges or rebases interact with approval, nor how post-approval changes are detected.

So the property is asserted by the source control system that performed the merge, to a consumer who cannot check it. This predicate carries what a third party needs to recompute it: the tree each approval covered, the tree that shipped, and how one became the other.

### Why don't existing predicates cover this?

-   **SLSA Provenance** — describes how an artifact was *built*, not how a change was *admitted*.
-   **SLSA Verification Summary** — communicates *that* a revision meets a source level. A conclusion, not the evidence to reach it independently.
-   **Source Provenance Attestations** — the natural home, deliberately left undefined by SLSA.
-   **Link / SCAI / SBOM** — describe steps, attributes and contents. None binds an approval to a tree.

### What does it look like?

See the Schema and Example sections. The Example is a complete Statement taken from a real repository, not composed by hand. Subjects use the existing `gitCommit` and `gitTree` digest types, so the central claim needs no new encoding.

### What policy questions does it answer?

-   Did an approval cover the exact tree that shipped?
-   If not, is the shipped tree the reviewed change replayed onto its base, with nothing added?
-   Which bytes shipped that no approval covers?
-   Did the checks run over a tree an approval covers, or over some other tree?
-   Was the evidence signed by a party other than the one being audited?

Each is answerable offline from the repository, without calling the system that produced the claim.

---

### Measured before proposing

The replay model was checked against production history rather than reasoned about, on Git 2.50.1:

| repository | workflow | sampled | identity | replay | residual | unexplained |
|---|---|---|---|---|---|---|
| huggingface/huggingface.js | squash | 160 squash commits | 87 | 73 | 0 | **0** |
| x402-foundation/x402 | squash | 250 squash commits | 88 | 162 | 0 | **0** |
| maziyarpanahi/openmed | merge | 400 merge commits | 336 | — | 64 | **0** |

810 merges, every one accounted for as `identity`, `replay`, or a named residual — no third category and no special cases. All 64 residuals are conflict resolutions: unreviewed by construction, and ordinary. Zero merge commits carried content absent from their parents.

### Notes for reviewers

**Type URI.** The specification is currently self-hosted at `https://drvelvetfog.github.io/source-review-coverage/v0.1`, because the predicate is already being issued and verified. If you would prefer it under `in-toto.io/attestation/source-review-coverage/v0.1`, I am happy to switch it and file the redirect PR — the namespace is yours to decide, and I did not want to assume it.

**Protobuf.** A definition exists and compiles under protoc 35.1, with the emitted JSON validated against it. I have left it out of this PR because generated bindings for four languages are committed in this repository, and it seemed better to have the design reviewed before adding that surface. Happy to open it as a follow-up, or add it here if you would rather review both together.

**Reference implementation.** Issues and verifies these attestations, including Sigstore signing by a workflow identity: <https://github.com/DrVelvetFog/source-review-coverage>.

One implementation note that may interest maintainers: `sigstore-python` rejects Statements whose subject uses `gitCommit` or `gitTree`, since its `DigestSet` is restricted to SHA-2 and SHA-3. That is a deliberate choice on their side (sigstore/sigstore-python#1018) rather than a problem with this predicate, but it does mean such statements cannot currently travel in a DSSE envelope signed by that client.

### DIFF
--- a/spec/predicates/README.md
+++ b/spec/predicates/README.md
@@ -28,6 +28,8 @@ our [vetting process], and may be of general interest:
 -   [SPDX2] and [SPDX3]: SPDX-formatted BOM for software artifacts.
 -   [Simple Verification Result]: Evidence that an artifact has been evaluated
     against one or more policies.
+-   [Source Review Coverage]: Evidence that a source revision was covered by
+    the review it claims, recomputable from the repository.
 -   [Test Result]: A generic schema to express results of any type of tests.
 -   [VULNS]: Defines the metadata to share the results of vulnerability scanning
     on software artifacts.
@@ -48,3 +50,4 @@ our [vetting process], and may be of general interest:
 [vetting process]: ../../docs/new_predicate_guidelines.md#vetting-process
 [Reference]: reference.md
 [Simple Verification Result]: svr.md
+[Source Review Coverage]: source-review-coverage.md
--- a/spec/predicates/source-review-coverage.md
+++ b/spec/predicates/source-review-coverage.md
@@ -1,0 +1,217 @@
+# Predicate type: Source Review Coverage
+
+Type URI: `https://drvelvetfog.github.io/source-review-coverage/v0.1`
+
+Version: v0.1
+
+Predicate name: `source-review-coverage`
+
+## Purpose
+
+To carry the evidence needed to decide, independently of the system that produced it, whether a source revision was covered by the review it claims.
+
+SLSA v1.2 Source Track Level 4 requires that changes to protected branches "be agreed to by two or more trusted persons prior to submission," and requires this of the *final revision submitted* — if changes occur during review, they need re-review. The requirement is precise. The verification is absent: SLSA "leaves source provenance attestations undefined and up to the SCSs to determine what works best," and does not address how squash merges or rebases interact with approval, nor how post-approval changes are detected.
+
+In practice the Source Control System asserts the property. A consumer receives "two people approved this" from the same platform that performed the merge, and either believes it or does not. This predicate replaces that assertion with recomputable evidence: which tree each approval covered, which tree shipped, and — where those differ because the base moved beneath the change — whether the shipped tree is the reviewed change replayed onto its base, exactly.
+
+## Use Cases
+
+**A reviewed change is not the change that shipped.** A reviewer approves a branch; the base advances; the forge merges. The tree that ships is no longer the tree anyone read. Under this predicate a verifier replays the reviewed change onto the base it landed on and compares: an exact match means the review covers the shipped state, and any difference is emitted as the specific bytes no approval covers. Conflict resolution is the common cause — a hunk is resolved into a third form at merge time, appears in no pull request, and reaches the default branch unreviewed.
+
+**Content enters inside a merge commit.** A merge commit can carry changes present in no parent. Forges do not display these in the pull-request diff, so they are unreviewed as a matter of course. Replaying the parents and comparing against the merge commit's tree surfaces them.
+
+**An approval is claimed but not bound.** Where an approval does not name the tree it covers, this predicate has nothing to check, and a verifier reports that rather than assuming coverage — which is the distinction between evidence and a claim.
+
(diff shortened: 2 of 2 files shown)


DrVelvetFog:
Checking in seven weeks on, with what's changed rather than just a ping.

Since I opened this, the predicate has gone from a spec with a reference verifier to something
being issued in production: a GitHub Action (Marketplace, `DrVelvetFog/source-review-coverage@v1`)
that emits one of these statements for every merge, signed by the workflow identity through
Sigstore, and a verifier that recomputes the result against a clone with git alone. It runs on
four repositories, all mine, so every verdict is honestly `UNVERIFIED` — nobody can approve their
own pull request, and I would rather ship that than a green badge I minted for myself. The first
outside contributor found a real divergence in the multi-parent replay; the fix landed on 11
September with a rename/rename octopus fixture that has to replay to the identical tree. None of
that changes the spec text in this PR.

The thing I said I had in view, I have now done, on someone else's predicate rather than my own.
I served this week as the second implementer for Adversarial Execution Evidence (#570): an
independent validator written from the specification text alone, Python and standard library
only, and written with an AI coding agent, which I name beside the language because on the
evidence of that thread it belongs there. It ran against the conformance corpus at a pinned tag
and is at https://github.com/DrVelvetFog/aee-e2.

The part I would point at is the order, not the score: spec text pinned by digest and corpus
verified before scoring, then the first run committed raw before any fix, then the resolution
log. One limit on that, which a reviewer there found before I did — I built locally and pushed
once, so every date in that history is mine to assert. The order is true; nothing outside the
repository can check that it is. I will timestamp anything I pre-register from here, which that
corpus already shows how to do.

It found fifteen defects, every one of them mine. That is narrower than I would have said a week
ago: a corpus caught my errors, which is what a corpus is for. It is not evidence that the text
is determinate, and I withdrew that inference on #570 when it was put to me.

One thing from it bears on ITE#63 directly. I have since filed a crosswalk into the reference
author's vocabulary registry, which promotes a term once an independent system emits it in a
signed artifact — and both implementations of that predicate are verifiers, so neither of us can
promote anything. Independent verification and independent issuance are different properties, and
a second-implementer seat buys only the first. Worth knowing before a tier assumes otherwise.

I mention all of it here only because it is the evidence for a claim I previously made in the
future tense. If in-toto/ITE#63 lands as drafted, I would like this predicate to be a `contrib`
listing as it stands and to work toward `vetted` properly, and I now have a worked example of
what I think that costs.

So my only question is still process: keep this open through the ITE, or close it and resubmit
once the tiers are in? Either is fine. I am happy to bring it to a community meeting if that is
the better venue.

The proto stays a follow-up as before, and the namespace offer stands: if you would rather it
lived under `in-toto.io`, I will move it and file the redirect.

A pass establishes review coverage and nothing else: not correctness, not that a human read
anything, not truthful authorship, not reviewer independence, not existence at a time, not
compliance.

astrogilda:
The replay table with zero unexplained merges is what carries #581. A `gitTree` subject lets a consumer recompute each policy answer from a clone; keep it open through the ITE as a `contrib` candidate. Reciprocity, measured: aee-e2 at its head agreed with [agent-evidence-vectors](https://github.com/astrogilda/agent-evidence-vectors) on all 272 verdicts and 61 results, and my verifier agreed with the thirteen aee-e2 fixtures. Your verifier-cannot-promote point holds, and I answer it on [your PR #2](https://github.com/astrogilda/agent-evidence-vocabulary/pull/2).

PR #592, opened 2026-09-06
Add predicate: Computation Receipt (verifiable numeric computation)

rybruscoe:
This adds a community predicate for **Computation Receipts (CR)**: a claim that a numeric computation (an HPC reduction, a training step, a model inference) produced a specific output, checkable by re-execution on any conforming implementation and any hardware, with no trusted machine.

**Use case.** A client verifies on its own (different) hardware that a provider's inference or reduction result is exactly correct; an auditor re-derives a recorded model output months later from the pinned weights, inputs and arithmetic profile; a mixed-hardware fleet proves two nodes computed the same answer to the bit.

**Why existing predicates do not cover it.** Existing predicates attest provenance (SLSA), review (link), or scanning results; none attests *numerical correctness of a computed result*. CR closes that gap for computations run under order-independent exact arithmetic, where every honest execution is bit-identical, so a digest-bound manifest plus re-execution is a complete verification protocol. For ordinary floating-point profiles the predicate still binds what ran, with verdicts that refuse to certify (`UNVERIFIABLE`) rather than overclaim.

**Policy questions it answers.** Was this output produced from these exact weights, inputs and arithmetic profile? Does an independent re-execution reproduce it bit for bit? Which implementation and profile does the claim depend on?

**Concrete example.** The format is an open spec (CC-BY-4.0) with conformance vectors, reference implementations in Python, C and Go, and a runnable in-toto Statement demo: https://github.com/anomly-labs/computation-receipts (`INTEROP.md`, `examples/python/in_toto_statement_demo.py`). A production emitter/verifier (llama.cpp and Ollama front-ends, SCITT statements) is https://github.com/anomly-labs/invar.

**Naming note.** "Receipt" here predates our awareness of the SCITT / RFC 9942 usage (transparency-log inclusion proof). The predicate name and document use "Computation Receipt" as a proper noun for the claim object and never bare "receipt" for an inclusion proof; the spec's interop guidance addresses the distinction explicitly.

On acceptance we will repoint our published `predicateType` from the current GitHub-URL placeholder to `https://in-toto.io/attestation/computation-receipt/v0.1` and submit the in-toto.io redirect line. A protobuf definition can follow if the maintainers want language bindings.


Rul1an:
Ran the conformance suite at 9450cfda before writing this and it passes, refusal vectors included, which is a better starting point than most format submissions get.

The predicate document looks like it stopped partway: Schema is an empty heading, and the template asks for a Parsing Rules subsection, a Fields subsection, and an Example carrying a whole Statement. svr.md and test-result.md are good short models for the first of those, scai.md if you want the fuller treatment.

The part I would most like to see written down is the verdict discipline, because it is the best thing about CR-v0.1 and none of it currently travels with the attestation. §6 keeping UNVERIFIABLE distinct from REJECT "including when the outputs happen to match", and §6.1 treating the registry as authoritative rather than the receipt, are exactly what someone needs to know before they write a policy against this predicate type. A line in Parsing Rules saying that the presence of a receipt is not an ACCEPT, and that the policy shape is "deny unless the profile is registered and order-independent", would carry INTEROP.md §3's point about verification being re-execution into the place a policy author actually reads.

rybruscoe:
@Rul1an Thank you for running the suite first and for reading CR-v0.1 closely enough to name §6 and §6.1 — those are the two rules the format exists for, and you are right that none of it travelled with the attestation.

Pushed a completed predicate document ([e29dc6c](https://github.com/in-toto/attestation/pull/592/commits/e29dc6c34d4b5e49d5080939c7e158d2303e1471), [rendered](https://github.com/rybruscoe/attestation/blob/e29dc6c34d4b5e49d5080939c7e158d2303e1471/spec/predicates/computation-receipt.md)): Schema is filled in, and it now has Parsing Rules, Fields and an Example carrying a whole Statement (produced by the reference implementation under the exact profile, so the digests are real). I followed svr.md and test-result.md for shape.

Parsing Rules now carries the verdict discipline in the place a policy author reads it:

- the presence of a receipt is not an ACCEPT — a valid signature, an inclusion proof or a certificate match establishes only that someone claimed the output; the only path to ACCEPT is the verifier's own re-execution with a matching digest;
- UNVERIFIABLE is distinct from REJECT, including when the outputs happen to match, and MUST NOT be mapped to either by a policy;
- the registry is authoritative, not the receipt: declared accumulation / order_independent / params must match the registry entry or the receipt is MALFORMED, and an unregistered profile is UNVERIFIABLE;
- the recommended policy shape is stated as "deny unless the profile is registered and order-independent, and re-execution returned ACCEPT";
- the integrity check (recompute digest(canonical(manifest)) against certificate) is called out as arithmetic-free so any in-toto consumer can run it, and `meta` is stated as excluded from the certificate and never evidence.

Happy to trim or restructure if the maintainers prefer the shorter svr.md length over this.

Rul1an:
Two things I could not check before are checkable now the document is complete, and both came back
clean: the subject join holds in your own Example, with `3bbac7c8…c71b` on both sides, and all six
section references resolve in `CR-v0.1.md`, including §11 correctly marked as draft. The closing
paragraph, the one where the same Statement with `float64` is equally well-formed and equally
signable and a verifier still answers `UNVERIFIABLE`, does more work than the verdict bullet it
restates. If the maintainers do ask for the shorter `svr.md` length, that is the paragraph I would
keep.

One more thing while the document is fresh. Parsing Rules says the only path to `ACCEPT` is
re-execution "with a matching output digest", and the Example says the verifier compares its output
digest to `manifest.output.digest`. §6 sets the bar higher: `ACCEPT` is when the re-execution
reproduced the certificate, and `REJECT` covers the case where the model, input, computation,
profile or output digest differs. Those two readings coincide until a verifier meets a manifest
field it does not model, which the Schema already admits with the optional `graph_digest`. A
verifier that reproduces the whole manifest sees a computation section it did not produce and
refuses, and that is what §13.1 leans on when it answers that a bogus `graph_digest` is caught
because the verifier's own computation section differs. A verifier reading only the output-digest
sentence here ignores the field, re-executes by name, and can accept. Matching those two sentences
to §6 costs a few words and keeps §13.1's answer travelling with the attestation.

The same thing one level up may be worth a sentence too. The framework says consumers MUST ignore
unrecognized fields, and that producers MAY add extension fields to any JSON object. Both need a
caveat inside `manifest`, because `certificate` covers the whole object. A consumer that drops an
unknown field there and then recomputes `digest(canonical(manifest))` over what it recognized gets
`MALFORMED` on a receipt that is well-formed. And an extension field added inside `manifest` changes
`certificate`, which the framework's own rule that an extension MUST NOT influence the meaning of
any other field already rules out. Your carried-verbatim bullet is aiming at this, and the
unrecognized-fields rule carries the exception "unless otherwise noted in the predicate
specification". Noting that `manifest` is closed, and that extensions belong in `meta` where you
already keep them out of the certificate, turns that bullet into a rule the framework explicitly
makes room for.

Smaller, and only if you are touching Parsing Rules anyway: the subject join keys on `subject[0]`
and says nothing about the length of the array, so a Statement with a matching first subject and a
second unrelated one satisfies it as written. Requiring exactly one entry closes that.

rybruscoe:
@Rul1an All three taken, pushed as one commit on the branch ([5f7f972](https://github.com/in-toto/attestation/pull/592/commits/5f7f972a14ba9ffb301cc554f19823a238a6949b)).

The ACCEPT wording was the real one. It now says the verifier's own re-execution has to reproduce the certificate, meaning the manifest it builds from what it actually ran canonicalizes to the same digest, and it calls out that comparing the output digest alone would let an unmodeled field like `graph_digest` slide through. The Example's verifier sentence says the same thing now instead of "compares its output digest".

On the closed manifest, I used the framework's own "unless otherwise noted in the predicate specification" clause as you suggested: consumers must not drop unrecognized fields inside `manifest` before recomputing, producers must not put extension fields there, and extensions go in `meta` where they were already outside the certificate.

Subject is now exactly one entry.

Thanks for checking the join and the section references against the spec. Agreed on the float64 paragraph; if the maintainers want it shorter, that one stays and the verdict bullet goes.

rybruscoe:
@marcelamelara @adityasaky This one is ready for a maintainer look whenever you have time. The predicate doc is complete after two rounds with @Rul1an (Schema, Parsing Rules carrying the CR verdict discipline, Fields, a full-Statement example with real digests), the conformance suite was run independently and passes, and CI is green. Happy to trim to svr.md length if you prefer.

goun7:
From an adjacent implementation of the same class — Tamga Protocol's charge receipts bind
input digest, pinned engine profile and output digest into a JCS-canonicalized, hash-chained
record, dispute resolution is re-run-and-compare (AT-002 determinism controls), and dead
endpoints yield an explicit indeterminate verdict that no retry loop may resolve into green.
Reading v0.1 end-to-end, two notes from having lived this shape:

1. The UNVERIFIABLE firewall ("agreement under order-dependent accumulation is coincidence,
   not evidence") is the right call and matches what we found independently: a verdict
   vocabulary that cannot express "cannot tell" quietly becomes a green-generating machine.
   Four verdicts vs our three is layering, not disagreement — MALFORMED/REJECT split mirrors
   our structure-vs-content REDs across layers (version gates, anchor-bind breaks).

2. Where we deliberately part ways: `meta` (operator, host, timestamp) is evidence-free by
   design in CR — right for numerical correctness, which is why we carry "who/when" on a
   different leg entirely (signed agent identity + optional node countersignature + epoch
   anchoring on our receipts). A reexecute class that wants to settle payments downstream
   needs both legs; the split between your predicate and a settlement extension is, in our
   reading, the interesting seam.

Conformance vectors with required refusals are exactly the culture this class needs; we run
47 controls + 51 slow including the RED families, and would be glad to check a CR receipt
against our verifier path as a cross-proof once the predicate lands, if useful.

*Disclosure: drafted with AI assistance; reviewed by the project founder. We maintain an
independent receipt-ledger protocol (Apache-2.0): https://github.com/goun7/tamga-protocol*

rybruscoe:
Thanks — and the line I'd most like on the record is yours: *"a verdict vocabulary that cannot express 'cannot tell' quietly becomes a green-generating machine."* That matches what we found the hard way.

Auditing our own certifier this week turned up seven defects. Three changed a verdict; four were a correct verdict wrapped in a sentence that would send an operator the wrong way — a coverage gap reported after the reassuring clause, a broken receipt log reading as "no receipts here", a mid-run redeploy misattributed to a batching bug. Those four matter more than they look: the machine-readable answer was right in every one, and the human reading it would have acted wrongly.

The worst of the three was exactly your dead-endpoint case. Two endpoints that both returned errors (500 and 503) compared as **identical across machines**, because an errored observation carries no tokens and no text, so the comparison found nothing to differ — a cross-machine bit-identity pass awarded for zero tokens produced. It fails closed now and names the reason. Independent convergence on that specific trap is worth more to me than agreement on the schema.

On the seam you've identified — agreed, and you've named it more precisely than we did. `meta` is evidence-free in CR because none of it is re-executable: operator, host and timestamp are attestations about the world, not about the computation, and mixing them into a numerical verdict is how "the right host said so" starts substituting for "the arithmetic checks out". Carrying who/when on a separate leg with its own signing and anchoring is the right shape, and a settlement extension joining the two legs at the subject digest sounds correct to me. No objection to CR being the numerical leg underneath that.

One distinction worth stating explicitly, since it may matter for cross-proofing: your receipts bind a **pinned engine profile**, and CR's exact profile is order-*independent* rather than pinned — the accumulation discards nothing, so the same request reproduces bit-identically on different hardware rather than only on the pinned stack. Both are valid answers to "can this be re-run", but they fail differently: a pinned receipt goes UNVERIFIABLE when the stack moves under it, which is the honest outcome, while an order-independent one should still reproduce. Worth knowing which kind a given receipt is making, because re-run-and-compare means something different in each case.

**Yes to the cross-proof.** Checking a CR receipt against your verifier path is exactly the kind of confirmation that counts — someone other than the author running it. The conformance vectors including the required refusals are in the repo now, so you needn't wait for the predicate to land if you want to start there; the refusal vectors are the interesting half.

goun7:
@rybruscoe Three points, then a commitment.

On the class distinction: you've named something our spec left implicit, and it deserves to be explicit. Tamga receipts are pinned-class — a charge receipt binds the engine-profile digest, and when the stack moves underneath it the correct verdict is UNVERIFIABLE, not a silent REJECT and not a lucky re-match. We'll state the class next to the determinism control in our docs so "re-run-and-compare" carries no ambiguity across the two schemes. Agreed that an order-independent profile is the stronger claim when it holds; it is simply not the claim we are making.

On your 500-vs-503 finding: that reads like our own dead-endpoint lesson wearing a different hat. An errored observation producing "nothing to differ" is exactly the shape our INDETERMINE rule exists to refuse — a verdict no retry loop may resolve into green. Independent convergence on that specific trap is good news for both formats: it suggests the refusal case is the load-bearing wall, not the acceptance path.

On the seam: welcome. If CR is the numerical leg underneath a who/when leg with its own signing and anchoring, the join at the subject digest is where a settlement story lives.

Commitment, since you took the cross-proof offer up: we'll run your conformance vectors — including the required refusals, which we agree are the interesting half — through our verifier path as a versioned acceptance control in our suite, and post the per-vector results here. "Someone other than the author running it" was your framing; same courtesy, reciprocated.

goun7:
Cross-proof, as promised — same day, so the word doesn't age into a plan.

We ran your published canonicalisation-layer vectors through our own path: the exact `jcs()` function our epoch-anchor digests are computed by (`tamga_verify_mini.jcs`, stdlib-only, no numpy — the tensor layer is `struct.pack` little-endian per your §3.1, and §3.2 named-collection is a chained hasher over sorted-name canonical parts). Grading is by **your** referee, vendored byte-identical (Apache-2.0, sha256-pinned, the pins re-verified inside our control so upstream drift turns us RED rather than silently sliding).

```
  PASS  canonical/empty-object                 [canonicalisation]  (canonical, digest)
  PASS  canonical/key-order                    [canonicalisation]  (canonical, digest)
  PASS  canonical/utf8                         [canonicalisation]  (canonical, digest)
  PASS  canonical/escaping                     [canonicalisation]  (canonical, digest)
  PASS  tensor/float64-2x3                     [canonicalisation]  (digest)
  PASS  tensor/int32-4                         [canonicalisation]  (digest)
  PASS  tensor/int8-6                          [canonicalisation]  (digest)
  PASS  tensors/named-order-independent        [canonicalisation]  (digest)
```

`canonical/escaping` is the one we cared about most: the `|`-strip on `i1`, lowercase `\u00xx`, unescaped solidus — exactly the class of "two honest emitters, different certificates" your §2 rationale describes, and our byte-identity held on first contact.

Scope honesty, since your runner lets one lie: our control asserts the candidate carries ONLY canonicalisation-layer names. The receipt/chain/refuse half is CR arithmetic semantics — Tamga is not a CR certifier, so we neither claim it nor fake it. A negative vector pins the referee's teeth: flip one hex digit of the escaping digest and the grading REDs (rc1) with the acceptance path intact.

Now a versioned acceptance control (AT-029) in our suite: `https://github.com/goun7/tamga-protocol/commit/ea54de7` — 8/8, and it fails CI if your vectors and our bytes ever diverge. Which is, we think, the whole point of "someone other than the author" running it.

goun7:
Correction to our cross-proof above — one of the eight vectors did not actually hold, and the divergence was only invisible because two different wrong answers agreed.

**What we claimed:** "8/8 byte-identity held on first contact" across the canonicalisation vectors, graded by your vendored referee.

**What was true:** 7 of 8. The exception is `canonical/utf8` — the `|\`-strip, lowercase `\\u00xx` vector. Our `jcs` at the time was `json.dumps(obj, sort_keys=True, separators=(\",\",\"\"), ensure_ascii=False)`, which is *not* RFC 8785: it escaped non-ASCII as `\\u00e9` where RFC 8785 §3.2.2.2 emits raw UTF-8. Our escaped output and the referee's expected-escaped output matched each other, so the test passed — but both were the same deviation from the RFC, not agreement on it. A canonicalizer that agreed with the *spec* would have failed our own test.

**How we found it (not how we should have found it):** a separate audit of the sister project (stillmarcus24, Dümen issue #4) flagged the same function as not-RFC-8785. Reproducing it against a Node/ECMAScript oracle showed the divergence immediately. Our own review had signed this function off as RFC 8785-compliant for weeks, which is the more damning part: the test was self-consistent, not self-correcting.

**What changed (Tamga v0.2.11, live today):**

- Single-center `tamga_canon.py` — ECMAScript `Number::toString` for the number layer, UTF-16 code-unit member ordering, raw-UTF-8 non-ASCII. All call sites delegate.
- AT-036: 13 vectors through our canonicalizer and a Node reference, hex-compared, byte-identity required. Node is ECMAScript-native, so it's an oracle rather than a second implementation with the same blind spot. 13/13. No Node → İNDETERMİNE (rc2), never a silent pass.
- AT-029 now grades 17 of your published vectors (was 8), including the ones the old code would have passed wrongly.

**What this means for your PR:** nothing about your spec or your referee was wrong — your vectors were correct and the referee was correct; we were the deviating party. The 6 `refuse/*` certificates are unaffected (digests byte-identical before and after). Only `canonical/utf8` moved, and it moved *toward* the RFC.

Apologies for the delay in catching it. The lesson we're keeping: a conformance test where the candidate and the expectation can be wrong together is not a conformance test.

goun7:
Following this closely — our project ships the same primitive from the agent side (a hash-chained work receipt whose verifier is a deterministic re-execution, no trusted machine), and the predicate shape here maps onto ours almost field-for-field. Two things measured on our side that may be useful to the predicate design:

**1. Pin the artifact, not the claim.** Our receipts bind the pinned module (`agent.wasm` sha256) plus declared resource limits, so the digest is recomputable by a stranger holding the same artifact. The subject of a Computation Receipt has to be the pinned artifact plus the input commitment; anything looser ("we ran the standard inference image") is not re-executable, only re-assertable.

**2. What is checkable vs what is true — the boundary belongs in the predicate.** A re-execution receipt is falsifiable only for what the record declares it covers. Ours carries declared egress (default-deny: the run-box records the only network window granted), so undetected extra behavior changes the digest even though no receipt can forbid it. For this predicate, the equivalent sentence is what a verifier may NOT conclude — a CR pinned over inputs and weights does not prove the weights were honestly derived, and the predicate saying so explicitly is worth more than any optional field.

We keep our duplicated canonicalization implementation in sync with a machine-checked triple-parity selftest because two copies of a rule will drift; the same discipline applies to a predicate that multiple tools will emit.

Not requesting changes — reading this as input to our own receipt schema, and happy to contribute the negative-vector conformance pattern (synthetic RED cases asserting the specific failure reason, not merely failure) if the predicate grows a conformance suite.

rybruscoe:
@goun7 Pushed a revision that does both:

- Recomputable binding. input is now required in every receipt, and model is required whenever the computation has weights; a verifier that can't rebuild the manifest from pinned artifacts returns UNVERIFIABLE, not ACCEPT. "The standard inference image" isn't a binding.
- manifest.coverage. An optional scope object (absent = the full named output). An ACCEPT attests exactly what it declares; anything outside — other outputs, rows, timing, side effects — is out-of-scope, neither attested nor denied. It lives inside manifest, so certificate binds it and the scope can't be widened after signing.

Both were your calls to make explicit — thanks for the push on each.

goun7:
The re-execution model is the right one. One failure mode we hit that's worth thinking about early in the spec.

We verify settlement claims where the claim carries its own digest. Our verifier does a three-way reconciliation: recompute the digest from the claim's own fields (not the provided one), check the claim actually follows from its inputs, and compare the presented verdict to the expected one.

The reason that first step matters: before we recomputed, a claim could be flipped after signing and the signature would still verify — the binding was cosmetic. The digest was present and well-formed, it just wasn't derived from anything. A re-execution predicate has the same shape: if the CR's manifest digest is taken as given rather than recomputed from the pinned manifest, the claim can be edited after the fact while every byte of the signature stays valid — and re-executing the edited manifest would agree with the edited claim.

Worth having the spec say a verifier MUST recompute the digest from the manifest rather than trust the field, and that a CR whose manifest doesn't recompute is UNVERIFIABLE rather than malformed — same verdict discipline already proposed for hardware mismatch.

Also seconding the naming note: our receipts are payment proofs, and the collision with SCITT inclusion proofs has cost us real confusion in review.

Rul1an:
At `c1c97bb`, the recomputation check raised above is already in Parsing Rules: a mismatch between `digest(canonical(manifest))` and `certificate` is `MALFORMED`, before anything else is considered. That is distinct from `UNVERIFIABLE` when the verifier cannot rebuild the manifest from pinned artifacts.

One sentence still carries the older, weaker criterion: Model says the claim is established by re-execution and comparing the output digest, whereas Parsing Rules and the Example require reproducing the certificate from the verifier's own manifest. Could Model use the same certificate-reproduction wording? That keeps an output-only match from reading as sufficient.

This is a document consistency point; I have not rerun the suite for this revision.

AI assistance disclosure: Codex performed this source comparison and drafted this comment.

